PR #11, opened 2016-08-18
rpc-test: create block files for import synchronous

bas-vk:
Currently the `makechain.js` create block files for the rpc-tests in an asynchronous matter. This has several problems:
1. the blocks directory might not been created, and writing block files into that directory can fail
2. the nodejs script can exit while not all files have been written since no callback is given for the `writeFile` method

This PR will do everything synchronously. This ensures all block files are written before the nodejs script exits.


karalabe:
Awesome, thx :)

PR #16, opened 2016-08-29
Add built-ins to parity chain spec

arkpar:
Fixes https://github.com/ethcore/parity/issues/2007


arkpar:
Just noticed that @tomusdrw has submitted the same fix. My version seems cleaner though and does not break tests.

karalabe:
LGTM :+1: Thanks!

PR #19, opened 2016-10-24
Fix parity client

holiman:
There were some strange switches there. See https://github.com/karalabe/hive/issues/18


karalabe:
SGTM

PR #20, opened 2016-10-25
Refactored the Dockerfiles

holiman:
This PR organises the Dockerfiles so that the docker cache is better reused between clients. Basically: 
- Fetch all general utilities and update package repositories 
- Check out the relevant github repo
- For parity: `cargo fetch` to download most of the dependencies

The steps above are cached by docker after the first client has been built. Afterwards, the remaining non-cached steps are
- Check out correct branch
- Build

This shaves a lot of times off the build times for the docker images
See https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#/build-cache

### DIFF
--- a/clients/go-ethereum:develop/Dockerfile
+++ b/clients/go-ethereum:develop/Dockerfile
@@ -5,17 +5,20 @@
 # needing to constantly rebuild.
 FROM alpine:3.4
 
-# Build go-ethereum on the fly and delete all build tools afterwards
-RUN \
-  apk add --update go git make gcc musl-dev                                    && \
-  git clone --depth 1 --branch develop https://github.com/ethereum/go-ethereum && \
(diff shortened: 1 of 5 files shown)


karalabe:
This will blow up the final size of the images by a huge amount. The idea behind the image organization was that they result in tiny sizes, only storing the essentials. By adding a lot of intermediate steps, it's not possible any more to clean up all the build junk that's only temporarily needed.

Also by removing the `--depth 1` flags from the various git clone commands, you now download the entire repository history instead of just pulling the head,

holiman:
Ah, that's a valid concern. Personally, or maybe for my use-case while testing locally, I'd rather have larger images and shorter build-times. And I'm aware about the `--depth 1`, I removed it to enable a shared base image. Primarily for parity, to be able to download dependencies.

PR #22, opened 2016-11-18
Added pythereum:develop client.

### LINKED ISSUE #7: Add pyethereum to the list of clients

Smithgift:
Some time ago, I saw that there was no pythereum harness. And I said to myself. "Hey, I can make that! Even if I'm not associated with pyethapp or pythereum at all, it'll just take a bit of work. How hard can it be?"

A month and many, many hours worth of labors later, I've finished a hacked-together almost-complete harness for pyethapp:develop.

Some notes:

* I'm a beginner at both pyethapp and docker, so apologies for any rough edges (of which there are many.)
* Despite a config file to the contrary, pyethapp will somehow bootstrap itself anyway? This is possibly a security hazard, or at least a testing clean room hazard, but it beats me how to stop it from doing it.
* pyethapp will reject a genesis.json where `mixHash` is used instead of `mixhash`. Parity will reject a genesis.json where `mixhash` is used instead of `mixHash`. I therefore manually alter the file with sed before starting pyethapp. Similarly, it's picky about account hexes, wherein I have stripped some `0x`s to please it. pyethapp's [#174](https://github.com/ethereum/pyethapp/issues/174)
* pyethapp will throw an exception if it receives an RPC call with an id of a bare zero. I am baffled why this one specific circumstance is a problem, but the simple solution is to modify all smoke tests to not do it. pyethapp's [#181](https://github.com/ethereum/pyethapp/issues/181)  
* I've altered a bunch of `test "$hash" == ...` statements to remove quotes. This magically causes it to work. I believe this is because pyethapp returns its responses with slightly different spacing. The bash sorcery to get this done properly is beyond me--and, technically, it works.
* The only smoke test it _doesn't_ pass is smoke/testnet-nonces. This is because pyethapp _interprets the command differently!_ While go-ethereum/parity use it as a simple nonce getting function, pyethapp does a little more work and returns the actual number of transaction.  After looking at the RPC spec itself, I can see it's ambiguous. The current test should pass if pyethapp adapts the other two implementation's interpretation.

(In case it isn't obvious, fixes #7)


Smithgift:
Rebased.

Smithgift:
Replaced by [this](https://github.com/karalabe/hive/pull/47).

PR #26, opened 2017-01-31
Hive blocktests

holiman:
This PR adds the simulator to perform blocktests using Hive. 

The simulator checks out the ethereum tests, and for each test and each client, loads the defined prestate and checks the posttate using the standard web3 API.


README.md
```diff
@@ -159,7 +159,11 @@ The behavioral configuration variables:
     * If `full`, assumes fast sync and consecutive pruning of historical state
     * If `light`, assumes header only sync and no state maintenance at all
   * `HIVE_FORK_HOMESTEAD` the block number of the Ethereum Homestead transition
-  * `HIVE_FORK_DAO_BLOCK` the block number of the DAO hard-fork transition
+  * `HIVE_FORK_TANGERINE` the block number of the Ethereum TangerineWhistle transition
+    * The HF for repricing certain opcodes, EIP 150
+  * `HIVE_FORK_SPURIOUS` the block number of the Ethereum Homestead transition
+    * The HF for replay protection, state cleaning etc. EIPs 155,160,161.
```
karalabe: Please move the tangerine and spurious fork details after the DAO in the list. I'm trying to maintain chronological order.

blocktests.sh
```diff
@@ -0,0 +1,29 @@
...
+#Create jsonp aswell
+echo "onData(" > $REPORTDIR/data.jsonp
+cat data.json >> $REPORTDIR/data.jsonp
+echo ");" >> $REPORTDIR/data.jsonp
+
+cp  data.json $REPORTDIR/
+
+tar -cvzf $REPORTDIR.tar.gz $REPORTDIR/ && rm -rf $REPORTDIR
+ 
+echo "Report generated into $NOW.tar.gz"
```
karalabe: I really don' think this file should be part of hive. Its place is in the project that uses hive to run some blocktests in particular. We could try to figure out a graphical reporting functionality for hive, but that should be generic and fully part of hive, not depending on an external resource to work.

clients/go-ethereum:local/geth.sh
```diff
@@ -64,6 +64,9 @@ fi
 echo "Initializing database with genesis state..."
 /geth $FLAGS init /genesis.json
 
+# Don't immediately abort, some imports are meant to fail
+set +e
+
```
karalabe: Could you detail where we have such scenarios. Notably, why would we add a block as a pre-state if we know it will fail. Shouldn't that block just be omitted since it's a noop anyway?

clients/go-ethereum:master/geth.sh
```diff
@@ -18,6 +18,8 @@
 #  - HIVE_FORK_DAO_VOTE  whether the node support (or opposes) the DAO fork
 #  - HIVE_MINER          address to credit with mining rewards (single thread)
 #  - HIVE_MINER_EXTRA    extra-data field to set for newly minted blocks
+#  - HIVE_FORK_TANGERINE block number of TangerineWhistle
+#  - HIVE_FORK_SPURIOUS  block number of SpurioisDragon
```
karalabe: Please move these two lines above the miner, right below the DAO fork details.

clients/go-ethereum:master/geth.sh
```diff
@@ -56,6 +58,17 @@ fi
...
 	chainconfig=`echo $chainconfig | jq ". + {\"daoForkSupport\": true}"`
 fi
+
+if [ "$HIVE_FORK_TANGERINE" != "" ]; then
+	chainconfig=`echo $chainconfig | jq ". + {\"eip150Block\": $HIVE_FORK_TANGERINE}"`
+fi
+if [ "$HIVE_FORK_SPURIOUS" != "" ]; then
+	chainconfig=`echo $chainconfig | jq ". + {\"eip158Block\": $HIVE_FORK_SPURIOUS}"`
+	chainconfig=`echo $chainconfig | jq ". + {\"eip155Block\": $HIVE_FORK_SPURIOUS}"`
+
```
karalabe: Please delete empty line.

clients/go-ethereum:local/geth.sh
```diff
@@ -78,6 +81,8 @@ if [ -d /blocks ]; then
 	done
 fi
 
+set -e
+
 # Load any keys explicitly added to the node
 if [ -d /keys ]; then
 	FLAGS="$FLAGS --keystore /keys"
```
karalabe: The fork additions into the genesis seems to be missing from this script.

clients/go-ethereum:stable/Dockerfile
```diff
@@ -5,6 +5,9 @@
 # needing to constantly rebuild.
 FROM alpine:latest
 
+# Make sure bash and jq is available for easier wrapper implementation
+RUN apk add --update bash jq
```
karalabe: Please make this same change for the other two geth dockerfiles too.

clients/go-ethereum:stable/geth.sh
```diff
@@ -18,6 +18,8 @@
 #  - HIVE_FORK_DAO_VOTE  whether the node support (or opposes) the DAO fork
 #  - HIVE_MINER          address to credit with mining rewards (single thread)
 #  - HIVE_MINER_EXTRA    extra-data field to set for newly minted blocks
+#  - HIVE_FORK_TANGERINE block number of TangerineWhistle
+#  - HIVE_FORK_SPURIOUS  block number of SpurioisDragon
```
karalabe: Please see review comments from clients/go-ethereum:master/geth.sh

simulators/blocktests/testrunner/Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+RUN apk add --update curl git && \
+  git clone --depth 1 https://github.com/ethereum/tests.git
+
+
+# Add the simulation controller
+ADD simulator.sh /simulator.sh
+RUN chmod +x /simulator.sh
+COPY *.py /
+
+ENTRYPOINT ["python","/simulator.py"]
```
karalabe: From the hive readme:

> The simulator definition(s) should reside in the simulators folder, each nested as <group>/<sim>, where <group> is a higher level collection of similar tests (e.g. dao-hard-fork), and <sim> is an individual simulator.

I'm not really sure what `blocktests/testrunner` hierarchy really means. Wouldn't it make more sense to have `consensus/block(tests)`? I.e. it would be nice to have all EF consensus tests groupped under one high level group (i.e. `consensus`), and I don't see a reason to have a further netsted `testrunner` folder in there.

holiman: The hierarchy is arbitrary, I wasn't sure how it should be structured. But to be specific, would you 
want me to replace

* simulators
    * blocktests
        * testrunner
            * Docker files etc

With 

* simulators
    * consensus
        * blocktests
            * Docker files etc

Or do you want me to skip a level (this is what I think you want): 
 
* simulators
    * consensus
        * Docker files etc

blocktests.sh
holiman: I agree, the reason it's there was mainly to serve as an example.. I can remove it though

holiman:
Think I've addressed all comments now..

PR #27, opened 2017-02-08
Fixed some Parity issues, and CPP-ethereum integration

holiman:
and the simulator to crash

PR #28, opened 2017-02-09
fix parity homesteadTransition config

cdetrio:
For `parity:master/parity.sh` the jq command `setpath` fails if the hex string `HEX_HIVE_FORK_HOMESTEAD` is not in quotes ("118C30" should be in quotes):

```
[39a55b90] jq: error: syntax error, unexpected IDENT, expecting ';' or ')' (Unix shell quoting issues?) at <top-level>, line 1:
[39a55b90] setpath(["engine", "Ethash", "params", "homesteadTransition"]; 118C30)                                                                  
[39a55b90] jq: 1 compile error
```

And for `parity:beta/parity.sh` the `homesteadTransition` config should be in hex according to the [wiki example](https://github.com/ethcore/parity/wiki/Proof-of-Work-Chains).


cdetrio:
Nevermind, duplicates #27.

PR #31, opened 2017-02-16
Add RPC test suite

bas-vk:
Commit 019f0fb includes a bugfix for geth clients. With the use of the gousb package building geth requires the linux headers to be installed. Currently this package isn't installed causing the build of geth master and stable clients to fail.


simulators/ethereum/rpc/eth/simulator.sh
```diff
@@ -0,0 +1,103 @@
...
+
+	for i in `seq 1 20`; do
+		# Fetch the current peer count and stop if it's correct
+		peers=`netPeerCount $ip`
+		if [ "$peers" == "$2" ]; then
+		  echo "Node $1 ($ip) has $peers/$2 peers"
+			break
+		fi
+		# Seems peer count is wrong, unless too many trials, sleep a bit and retry
+		if [ "$i" == $N_NODES ]; then
```
karalabe: `$N_NODES` should be 20 here. The goal of waitPeers is to allow a few retries for reaching the desired peer count before giving up. This condition checks whether the retry count was reached or not. That count is 20 in your case per line 35.

PR #34, opened 2017-02-25
Fix concurrent write

holiman:
Fix for https://github.com/karalabe/hive/issues/33
Also implements concurrent consensus testing


simulator.go
```diff
@@ -203,6 +204,8 @@ type simulatorAPIHandler struct {
 	nodes  map[string]*docker.Container
 
 	result *simulationResult
+	nodesMutex sync.RWMutex
```
karalabe: Please only call it lock, it will be needed when updating subresults too concurrently.

simulator.go
```diff
@@ -296,8 +299,10 @@ func (h *simulatorAPIHandler) ServeHTTP(w http.ResponseWriter, r *http.Request)
 				time.Sleep(100 * time.Millisecond)
 			}
 			// Container online and responsive, return it's ID for later reference
-			h.nodes[container.ID[:8]] = container
 			fmt.Fprintf(w, "%s", container.ID[:8])
+			h.nodesMutex.Lock()
+			defer h.nodesMutex.Unlock()
+			h.nodes[container.ID[:8]] = container
```
karalabe: I'd really prefer without using defer here. That thing is useful if you can have multiple exit paths, or if you have a simpler method. Using it inside a huge function nested deep somewhere is a bit brittle to refactors. Simply modifying the switch statement so it doens't return immediately could deadlock it.

```go
h.lock.Lock()
h.nodes[container.ID[:8]] = container
h.lock.Unlock()
```

holiman:
Done!

PR #35, opened 2017-03-01
Fix failing RPC eth tests + light client support

bas-vk:
This PR includes:

- fixes 3 tests failing due to timeouts on slower environments (latency, difficulty too high)
- updates `internal/ethash` so that it uses the new log package
- adds 2 light clients to the cluster and run the same set of RPC tests against them


holiman:
Closing this, stale now

PR #39, opened 2017-03-18
Faster geth imports, measure times, Ethash log fix

holiman:
* added retries on failing instance-startup,  
* added speedier block import in geth, 
* added testcase time measurements

### DIFF
--- a/clients/go-ethereum:master/geth.sh
+++ b/clients/go-ethereum:master/geth.sh
@@ -88,12 +88,7 @@ fi
 # Load the remainder of the test chain
 echo "Loading remaining individual blocks..."
 if [ -d /blocks ]; then
-	cd blocks
-	../geth $FLAGS --nocompaction import `ls | sort -n`
-	cd ..
-#	for block in `ls /blocks | sort -n`; do
-#		/geth $FLAGS import /blocks/$block
-#	done
+	(cd blocks && ../geth $FLAGS --nocompaction import `ls | sort -n`)
 fi
 
 set -e
--- a/clients/go-ethereum:master/geth.sh
+++ b/clients/go-ethereum:master/geth.sh
@@ -88,9 +88,12 @@ fi
 # Load the remainder of the test chain
 echo "Loading remaining individual blocks..."
 if [ -d /blocks ]; then
-	for block in `ls /blocks | sort -n`; do
-		/geth $FLAGS import /blocks/$block
-	done
+	cd blocks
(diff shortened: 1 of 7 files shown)


clients/go-ethereum:master/geth.sh
```diff
@@ -88,9 +88,12 @@ fi
...
 if [ -d /blocks ]; then
-	for block in `ls /blocks | sort -n`; do
-		/geth $FLAGS import /blocks/$block
-	done
+	cd blocks
+	../geth $FLAGS --nocompaction import `ls | sort -n`
+	cd ..
+#	for block in `ls /blocks | sort -n`; do
+#		/geth $FLAGS import /blocks/$block
+#	done
```
karalabe: Any reason for leaving this comment here? Any reason for updating only master?

clients/go-ethereum:master/geth.sh
```diff
@@ -88,9 +88,12 @@ fi
 # Load the remainder of the test chain
 echo "Loading remaining individual blocks..."
 if [ -d /blocks ]; then
-	for block in `ls /blocks | sort -n`; do
-		/geth $FLAGS import /blocks/$block
-	done
+	cd blocks
+	../geth $FLAGS --nocompaction import `ls | sort -n`
+	cd ..
```
karalabe: This could be simply

```
(cd blocks && ../geth $FLAGS --nocompaction import `ls | sort -n`)
```

karalabe:
https://github.com/karalabe/hive/blob/master/clients/go-ethereum:stable/geth.sh#L89

https://github.com/karalabe/hive/blob/master/clients/go-ethereum:local/geth.sh#L89

https://github.com/karalabe/hive/pull/39#discussion_r112429761

>  Any reason for updating only master?

holiman:
Kind of, I'm not sure which versions of geth got `--nocompaction` in it.

karalabe:
Well, given that we just released 1.6.0, it's safe to say every one of them has it by now.

karalabe:
Although I see your PR was open way longer. I apologize for not seeing/forgetting about it.

holiman:
Not in 1.5, which is what `stable`  points to : 
https://github.com/ethereum/go-ethereum/blob/release/1.5/cmd/utils/flags.go

karalabe:
Erm, bummer, we need to update that then.

holiman:
Ok, updated

PR #41, opened 2017-04-21
Fix Parity, Geth-json, ignore invalid hex/rlp

holiman:
Fixes in this PR

* Parity changed repo to 'paritytech', thus dockerfiles stopped working
* Geth's new (and grumpy) json-parsing, requires us to rewrite various fields in genesis
* Skips tests which contain invalid hex; specifically the "invalid rlp"-tests which cannot be converted into binary, thus impossible to execute. 
* Added config for Metropolis, both Geth and Parity (parity tests use Metro-rules unless this is included, so tests fail)
* Defined `Metropolis` network, which the ethereum tests use.

### DIFF
--- a/clients/go-ethereum:local/geth.sh
+++ b/clients/go-ethereum:local/geth.sh
@@ -17,7 +17,8 @@
 #  - HIVE_FORK_DAO_BLOCK block number of the DAO hard-fork transition
 #  - HIVE_FORK_DAO_VOTE  whether the node support (or opposes) the DAO fork
 #  - HIVE_FORK_TANGERINE block number of TangerineWhistle
-#  - HIVE_FORK_SPURIOUS  block number of SpurioisDragon
+#  - HIVE_FORK_SPURIOUS  block number of SpuriousDragon
+#  - HIVE_FORK_METROPOLIS block number for Metropolis transition
 #  - HIVE_MINER          address to credit with mining rewards (single thread)
 #  - HIVE_MINER_EXTRA    extra-data field to set for newly minted blocks
 
@@ -65,6 +66,9 @@ fi
 if [ "$HIVE_FORK_SPURIOUS" != "" ]; then
 	chainconfig=`echo $chainconfig | jq ". + {\"eip158Block\": $HIVE_FORK_SPURIOUS}"`
 	chainconfig=`echo $chainconfig | jq ". + {\"eip155Block\": $HIVE_FORK_SPURIOUS}"`
+fi
+if [ "$HIVE_FORK_METROPOLIS" != ""]; then
+	chainconfig=`echo $chainconfig | jq ". + {\"MetropolisBlock\": $HIVE_FORK_METROPOLIS}"`
 fi
 
 if [ "$chainconfig" != "{}" ]; then
--- a/simulators/ethereum/consensus/hivemodel.py
+++ b/simulators/ethereum/consensus/hivemodel.py
@@ -29,7 +29,7 @@ class HiveNode(object):
         return self._getNodeData("web3_clientVersion",[])
 
     def getBlockByNumber(self,blnum):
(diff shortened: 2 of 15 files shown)


holiman:
Oh, and incidentally closes https://github.com/karalabe/hive/issues/40

simulators/ethereum/consensus/hivemodel.py
```diff
@@ -177,13 +177,17 @@ def perform_work(testcase):
                     testcase.details()
                 )
 
+        if True:
+            pool = ThreadPool(7) 
+            #Turns out a raw iterator isn't supported, so comprehending a list instead :(
+            pool.map(perform_work, [x for x in iterator()])
+            pool.close()
+            pool.join()
+        else:
```
gsalgado: The else block will never be executed here

holiman: True, the point was just that I wanted to run it sequentially as a one-off hack, and be able to do it with minimal modification on the the server.

clients/go-ethereum:local/geth.sh
```diff
@@ -66,6 +67,9 @@ if [ "$HIVE_FORK_SPURIOUS" != "" ]; then
 	chainconfig=`echo $chainconfig | jq ". + {\"eip158Block\": $HIVE_FORK_SPURIOUS}"`
 	chainconfig=`echo $chainconfig | jq ". + {\"eip155Block\": $HIVE_FORK_SPURIOUS}"`
 fi
+if [ "$HIVE_FORK_METROPOLIS" != ""]; then
+	chainconfig=`echo $chainconfig | jq ". + {\"MetropolisBlock\": $HIVE_FORK_METROPOLIS}"`
```
karalabe: This will probably be `metropolisBlock`, i.e. lowercase M. Perhaps it would be better to name it like that from the get go.

holiman:
Corrected in https://github.com/karalabe/hive/pull/43

simulators/ethereum/consensus/testmodel.py
```diff
@@ -122,31 +143,41 @@ def get(self, key):
...
+                if len(v) > 2 and v[:2] != '0x':
+                    raw_genesis[key] = '0x'+raw_genesis[key]
+
+            # And fix the alloc-section
+            alloc = {}
+
+            for addr in self.data['pre']:
+                v = self.data['pre'][addr]
+                if not addr[:2] == '0x':
+                    addr = '0x'+addr
```
gsalgado: Because of this change we can no longer run the consensus tests against pyethereum as now the addresses end up being 42 bytes long and that makes pyethereum [think they don't need to be hex decoded](https://github.com/ethereum/pyethereum/blob/develop/ethereum/blocks.py#L1465-L1467). I suppose this needs to be fixed in pyethereum, but in the meantime it'd be nice if we could workaround it here. Would that be possible?

holiman: Ah, messy. Well, geth has become a lot more strict in this regard, so geth requires `0x`-prefixes in alloc. But actually.. `alloc` seems specified as `UnprefixedAddress` , https://github.com/ethereum/go-ethereum/blob/9fc90b6747c4eada258071cf65bebb6afd991a1c/core/genesis.go#L74 , which uses [`UnmarshalFixedUnprefixedText`](https://github.com/ethereum/go-ethereum/blob/8771c3061f340451d0966adcc547338a25f2231f/common/hexutil/json.go#L96) :

> decodes the input as a string with optional 0x prefix

I just tested a bit, and it appears to work fine doing the opposite:

            for addr in self.data['pre']:
                v = self.data['pre'][addr]
                if addr[:2] == '0x':
                    addr = addr[2:]

gsalgado: Indeed, that works for pyeth as well. I've made that change on my rpc-compat branch

PR #42, opened 2017-04-24
Fix a couple goroutine leaks

gsalgado:
We were leaking up to two goroutines for every node started, and when
using loglevel>5, one of them would leave open FDs behind as well,
which would eventually cause the hive node to not be able to start
any more nodes due to having too many open files.

Also remove some trailing whitespace and simplify a couple things.

### DIFF
--- a/containers.go
+++ b/containers.go
@@ -312,6 +312,7 @@ func runContainer(daemon *docker.Client, id string, logger log15.Logger, logfile
 	// If we're the outer shell, log straight to stderr, nothing fancy
 	stdout := io.Writer(os.Stdout)
 	stream := io.Writer(os.Stderr)
+	var fdsToClose []io.Closer
 	if !shell {
 		// For non shell containers, create and open the log file for the output
 		if err := os.MkdirAll(filepath.Dir(logfile), os.ModePerm); err != nil {
@@ -322,12 +323,17 @@ func runContainer(daemon *docker.Client, id string, logger log15.Logger, logfile
 			return nil, err
 		}
 		stream = io.Writer(log)
+		fdsToClose = append(fdsToClose, log)
 
 		// If console logging was requested, tee the output and tag it with the container id
 		if *loglevelFlag > 5 {
 			// Hook into the containers output stream and tee it out
-			hookedR, hookedW := io.Pipe()
+			hookedR, hookedW, err := os.Pipe()
+			if err != nil {
+				return nil, err
+			}
 			stream = io.MultiWriter(log, hookedW)
+			fdsToClose = append(fdsToClose, hookedW)
 
 			// Tag all log messages with the container ID if not the outer shell
 			copy := func(dst io.Writer, src io.Reader) (int64, error) {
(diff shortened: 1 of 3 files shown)


simulator.go
```diff
@@ -266,14 +266,26 @@ func (h *simulatorAPIHandler) ServeHTTP(w http.ResponseWriter, r *http.Request)
 				http.Error(w, err.Error(), http.StatusInternalServerError)
 				return
 			}
-			logger = logger.New("id", container.ID[:8])
+			cLogger := logger.New("id", container.ID[:8])
```
karalabe: Why did you rename this? It will cause the loggers below not to have the `id` embedded any more. This will especially be bad for the new termination logs you added, since they don't have any context, so you'll just get a batch of `client container finished cleanly` messages one after the other. Am I understanding the code incorrectly?

containers.go
```diff
@@ -361,5 +369,28 @@ func runContainer(daemon *docker.Client, id string, logger log15.Logger, logfile
 		logger.Error("failed to start container", "error", err)
 		return nil, err
 	}
+	waiter = fdClosingWaiter{
+		CloseWaiter: waiter,
+		closers:     fdsToClose,
+		logger:      logger,
+	}
```
karalabe: Please refactor into 

```go
return fdClosingWaiter{
	CloseWaiter: waiter,
	closers:     fdsToClose,
	logger:      logger,
}, nil
```

I'd really like to avoid reassigning waiter with a completely different type that happens to implement the same interface.

containers.go
```diff
@@ -312,22 +312,30 @@ func runContainer(daemon *docker.Client, id string, logger log15.Logger, logfile
 	// If we're the outer shell, log straight to stderr, nothing fancy
 	stdout := io.Writer(os.Stdout)
 	stream := io.Writer(os.Stderr)
+	var log, hookedR, hookedW *os.File
+	var fdsToClose []io.Closer
+	var err error
```
karalabe: Declaring `err` here just makes the code convoluted. It "hints" that errors are propagated through for some reason. However the code below just checks for error conditions and either returns or continues. There's no need to declare it up front.

simulator.go
```diff
@@ -266,14 +266,26 @@ func (h *simulatorAPIHandler) ServeHTTP(w http.ResponseWriter, r *http.Request)
...
 
 			logfile := fmt.Sprintf("client-%s.log", container.ID[:8])
-			if _, err = runContainer(h.daemon, container.ID, logger, filepath.Join(h.logdir, logfile), false); err != nil {
+			waiter, err := runContainer(h.daemon, container.ID, cLogger, filepath.Join(h.logdir, logfile), false)
+			if err != nil {
 				logger.Error("failed to start client", "error", err)
 				http.Error(w, err.Error(), http.StatusInternalServerError)
 				return
 			}
+			go func() {
```
karalabe: Please don't start a new goroutine here. Instead, add a field to `simulatorAPIHandler` https://github.com/karalabe/hive/blob/master/simulator.go#L192 below `nodes` called `waiters`, that you can iterate over after closing the containers https://github.com/karalabe/hive/blob/master/simulator.go#L368. 

You'd also need to add the wait logic to the node `DELETE` handler https://github.com/karalabe/hive/blob/master/simulator.go#L340.

Having the extra field is cleaner that one more concurrent goroutine which may run at an arbitrary point in time.

containers.go
karalabe: Actually, if I read the code correctly, non of these pre-declares (apart from the closer slice) you added are needed here, and they just convolute the code. Please revert to the original code that declared them when used.

simulator.go
gsalgado: I thought we'd only want the client prefix in the log messages generated within runContainer(), but as you pointed out they're helpful for the termination logs I added, and probably also for other logs emitted further down, so I'll revert

containers.go
gsalgado: Ok, done

containers.go
gsalgado: The hookedW/hookedR/err declarations are leftovers from a previous version of this fix -- they're indeed not necessary, so I reverted them.

simulator.go
gsalgado: That means we'll end up with a ton of idle goroutines (up to two for every node we start, depending on the loglevel used) lingering around until the simulator node is terminated, and since we'd only close them at the end, the log files would remain open until then and we'd probably still hit the open file limit and crash, which is what I'm trying to fix.

> You'd also need to add the wait logic to the node DELETE handler https://github.com/karalabe/hive/blob/master/simulator.go#L340.

I don't understand why this would be necessary.

gsalgado:
@karalabe I think I've addressed all your comments. Can you have another look?

PR #46, opened 2017-05-02
Only use colors when logging to a tty

gsalgado:


karalabe:
This breaks when hive is ran in its own shell, since the outer shell will have colors, but the inner one will switch over to logfmt.

gsalgado:
Maybe I should add a new CLI flag to enable/disable colors? In that case, should the default be colors or no-colors?

gsalgado:
I'll abandon this as I found out I can use ansi2txt to get rid of the non-printable chars from the log when I want to parse/grep it

PR #50, opened 2017-05-02
New simulator for RPC compatibility tests

gsalgado:


simulators/ethereum/rpc-compat/Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+  cd interfaces                                               && \
+  cp -a rpc-specs-tests/. /                                   && \
+  mv /configs/bcRPC_API_Test.json /bcRPC_API_Test.json
+
+# Add the simulation controller
+ADD genesis.json /genesis.json
+# put keys in folder for parity
+ADD keys.tar.gz /keys/Hive
+# put keys in folder for geth
+ADD keys.tar.gz /keys
```
karalabe: Simulators should not care whether they run against Geth or Parity or whatever. Please don't have custom logic for any client.

simulators/ethereum/rpc-compat/parity-config.toml
```diff
@@ -0,0 +1,3 @@
+[account]
+unlock = ["0xbe93f9bacbcffc8ee6663f2647917ed7a20a57bb"]
+password = ["/keys/password.file"]
```
karalabe: Why do you need to unlock the account for Parity? Why do you need custom logic for Parity?

simulators/ethereum/rpc-compat/Dockerfile
gsalgado: I don't see any code using a keys.tar.gz file anywhere, so I think I can probably remove the four lines above. Will check with @cdetrio as he's the original author

cdetrio: The docker `ADD` command unzips a tar.gz file. Parity needs the keystore file in that folder https://github.com/paritytech/parity/issues/4653.

karalabe: Hive defines the keys to be injected to be located in `/keys`. It is up to client startup scripts to move them to their final destination. Geth does this by explicitly loading keys from that folder https://github.com/karalabe/hive/blob/master/clients/go-ethereum:master/geth.sh#L104. If Parity needs the keys in an explicit path, the startup script needs to copy it there.

karalabe: However AFAIK it already does https://github.com/karalabe/hive/blob/master/clients/parity:master/parity.sh#L125

simulators/ethereum/rpc-compat/parity-config.toml
cdetrio: parity-config.toml might be cruft, will double check

gsalgado: I've removed it and it still runs fine

gsalgado:
Pushed a new commit without the parity-specific bits, which were not needed anyway

gsalgado:
@karalabe I've addressed your comments and rebased. It'd be great if you could have another look

PR #51, opened 2017-05-04
Do not ignore docker errors when stopping/removing containers

gsalgado:


containers.go
```diff
@@ -148,7 +148,12 @@ func createClientContainer(daemon *docker.Client, client string, tester string,
 	if err != nil {
 		return nil, err
 	}
-	defer daemon.RemoveContainer(docker.RemoveContainerOptions{ID: t.ID, Force: true})
+	defer func() {
+		err := daemon.RemoveContainer(docker.RemoveContainerOptions{ID: t.ID, Force: true})
+		if err != nil {
+			log15.Error("failed to cleanup tester container", "error", err)
```
karalabe: Please add the ID here too.

internal.go
```diff
@@ -28,7 +28,10 @@ func makeGenesisDAG(daemon *docker.Client) error {
 	log15.Debug("created ethash container")
 	defer func() {
 		log15.Debug("deleting ethash container")
-		daemon.RemoveContainer(docker.RemoveContainerOptions{ID: ethash.ID, Force: true})
+		err := daemon.RemoveContainer(docker.RemoveContainerOptions{ID: ethash.ID, Force: true})
+		if err != nil {
+			log15.Error("failed to delete ethash container ", "error", err)
```
karalabe: Please add the ID here too.

internal.go
```diff
@@ -46,7 +49,10 @@ func makeGenesisDAG(daemon *docker.Client) error {
 	go func() {
 		<-interrupt
 		errc <- errors.New("interrupted")
-		daemon.StopContainer(ethash.ID, 0)
+		err := daemon.StopContainer(ethash.ID, 0)
+		if err != nil {
+			log15.Error("failed to stop ethash", "error", err)
```
karalabe: Please add the ID here too.

shell.go
```diff
@@ -31,7 +31,10 @@ func mainInShell(daemon *docker.Client, overrides []string) error {
 	log15.Debug("created shell container")
 	defer func() {
 		log15.Debug("deleting shell container")
-		daemon.RemoveContainer(docker.RemoveContainerOptions{ID: shell.ID, Force: true})
+		err := daemon.RemoveContainer(docker.RemoveContainerOptions{ID: shell.ID, Force: true})
+		if err != nil {
+			log15.Error("failed to delete shell container", "error", err)
```
karalabe: Please add the ID here too.

shell.go
```diff
@@ -48,7 +51,10 @@ func mainInShell(daemon *docker.Client, overrides []string) error {
 	go func() {
 		<-interrupt
 		log15.Error("shell interrupted, stopping")
-		daemon.StopContainer(shell.ID, 0)
+		err := daemon.StopContainer(shell.ID, 0)
+		if err != nil {
+			log15.Error("failed to stop hive shell", "error", err)
```
karalabe: Please add the ID here too.

internal.go
gsalgado: I didn't add the IDs here or in shell.go because AIUI there's only one container for ethash/shell, and none of the logs emitted here/there include the ID. Should I do so anyway?

karalabe: I guess that's fine too. I figured it might be useful to see the container id but it's not that relevant. Please add it to the tester though since there might be multiple testers per hive run.

gsalgado: Ok, done

PR #55, opened 2017-05-29
Add whisper RPC simulation

bas-vk:
This PR adds the `simulators/ethereum/rpc/shh` simulator.

### DIFF
--- a/clients/go-ethereum:local/geth.sh
+++ b/clients/go-ethereum:local/geth.sh
@@ -26,6 +26,8 @@
 # Immediately abort the script on any error encountered
 set -e
 
+FLAGS="-shh"
+
 # It doesn't make sense to dial out, use only a pre-set bootnode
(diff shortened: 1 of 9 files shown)


bas-vk:
After review comments the whisper client is moved to its own package. This causes this simulator to fail. After an upgrade of docker none of the hive tests/simulations work due to network issue in the docker containers. This prevents me from testing it. Next week I have access to my normal dev machine and will update the pr to the latest codebase.

intizarabe:
This is devide and conquer math method, annoyng frankly. Whisper imho is not healthy communication style. I am really wonder, this is combat operations instead of peacekeeping . I am bad guy , agreed, but i cant accept such society. Imho communications better when clear than opaque. Make humans departed and unhappy , it is no logics. Just my imho

gballet:
Whisper is no longer in active development, closing

PR #56, opened 2017-06-09
Fix pyethereum:develop client

gsalgado:
It broke with the massive changes that were merged recently onto
pyethapp/pyethereum.


gsalgado:
@karalabe can you have a look at this when you have a chance?

PR #57, opened 2017-06-09
simulators/consensus: do not crash when newNode() fails

gsalgado:
If newNode() raises an exception we were recording a test failure and
moving on, eventually trying to use its return value (to kill the node)
and crashing with an UnboundLocalError because the variable was
referenced before assignment. Now we record the test failure and exit
early.


gsalgado:
@karalabe can you have a look at this when you have a chance?

PR #62, opened 2017-06-30
Initial commit of transaction tests (not working)

holiman:
This is a WIP to implement transaction tests. Since testeth cannot create blocks with invalid transactions, it would be really good to have the ability to fire off various invalid transactions to the clients, and verify that the clients correctly rejects invalid ones. 

This is the start a simulator which attempts to do this. Not ready to be merged (although I guess it wouldn't actually break anything)


holiman:
cc @cdetrio @mattdf

PR #69, opened 2017-09-01
Configure cpp and parity for metro

holiman:
This is running on hive now, I'll merge if it it seems to work. Stats, pre-patch:

`6973 : 13013  / 19986`


holiman:
First fix had a bash-error in parity config, which is fixed now. Parity does not perform any better than before, but cpp have a couple of thousand less failures.

PR #70, opened 2017-09-05
skip constantinople tests in consensus runner

cdetrio:
## Warning: PR still untested

Skip Constantinople cases since they are guaranteed to show up as failures on [hivetests](http://hivetests.ethstats.net/) for clients that are only targeting Byzantium.


holiman:
I'd prefer fixing it in a different way... 
Rulesets are defined [here](https://github.com/karalabe/hive/blob/master/simulators/ethereum/consensus/testmodel.py#L5), and right now, it falls back to a default if the `network` doesn't match one of the defined. 

A better approach would be to call e.g. `testcase.validNetwork()` which checks against ruleset,instead of hardcoding `Constantinople`.

PR #72, opened 2017-09-09
Use parity metro branch

holiman:
Trying this out now, will merge if it works


holiman:
Merging this, it's a PITA to keep a separate branch. It doesn't break anything more than it already is br0ken

PR #74, opened 2017-09-12
actual fix for cpp Byzantium

cdetrio:
What https://github.com/karalabe/hive/pull/73 was supposed to be.  requires https://github.com/ethereum/cpp-ethereum/pull/4488


cdetrio:
The `cpp-ethereum:develop/eth.sh` script could be better, but at least its (hopefully) not broken.

PR #84, opened 2017-11-28
Adding EthereumJ client to Hive

zilm13:
Let's add EthereumJ, Ethereum Java implementation. It passes smoke, all validations, but fails on most simulations. As I understand, minimal requirements are to pass smoke so we could jump in. 
Also I've updated go-ethereum master and develop Dockerfiles to include required libs during the build as it fails for me (and I see you've included them in metro setup too).
We will continue to work with failed simulations tests later, need to switch to other tasks currently. If it requires another updates in Hive (hope, not) I'll submit another PR later.

### DIFF
--- a/clients/ethereumj:develop/Dockerfile
+++ b/clients/ethereumj:develop/Dockerfile
@@ -1,0 +1,106 @@
+# Docker container spec for building the master branch of EthereumJ.
+#
+# The build process it potentially longer running but every effort was made to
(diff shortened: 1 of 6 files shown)


zilm13:
Yeah, looks messy.
Updated it to 2 logical commits. Is it ok?

holiman:
Beautiful, thanks!

PR #85, opened 2017-12-10
EthereumJ local configuration

zilm13:
One more thing from us. Added local configuration for EthereumJ (Java Ethereum implementation) to use Hive in local environment and on CI. 
Also, could you let me know how to add EthereumJ tests results on CIrcleCI to this statistics: http://hivetests.ethstats.net/
Thank you.


holiman:
The data on http://hivetests.ethstats.net/ isn't from circleci. The server there runs hivetests back-to-back all the time, and exports the resulting json-files in a little bash script. We're right now in the process of moving it to another server, and I'll likely tear it down any day. I'll check if EthereumJ seems to work now, and enable that client too!

zilm13:
Great, thank you!

PR #108, opened 2018-08-09
Aleth update

chfast:
I'm not sure how to test it locally, hopefully that's not needed.
The aleth docker image should include version.json file and improved aleth.py wrapper, but you have to wait until nightly build today.

@holiman

PR #109, opened 2018-08-10
Alethnew

holiman:
Contains some fixes for aleth. I wasn't able to push to @chfast's fork, so opened a separate PR instead. Obsoletes #108


clients/aleth:nightly/Dockerfile
```diff
@@ -16,6 +16,16 @@ ADD config.json /config.json
 ADD aleth.sh /aleth.sh
 RUN chmod +x /aleth.sh
 
+RUN apk add --no-cache jq bc bash
+
+# The /version.json is a softlink, which hive won't resolve when trying to read it.
+# We copy it into a regular file, and also remove the 'bool' field is_prerelease
+
+
+RUN cat /version.json | jq "del(.is_prerelease)" > /x.json && \
```
chfast: Can't hive just ignore unknown fields?

holiman: Hive? It's aleth that compains about unknown fields, so hive has to massage the config/genesis to match what aleth expects.

holiman: Oh sorry, yeah, you're right. In this case, it's because @karalabe cheated a bit when reading the json file: https://github.com/karalabe/hive/blob/master/images.go#L86 
He reads it into a string to string map, so booleans doesn't fit :)

chfast: I will try to fix the issue with the config file.
For now I removed the `/version.json` symbolic link. Use `/usr/share/aleth/buildinfo.json` in case this fails tomorrow.

PR #115, opened 2018-08-23
WIP: Add client for py-evm (trinity)

cpurta:
Adding the py-evm (Trinity) client to hive so that tests can be ran against the client. This will resolve https://github.com/ethereum/py-evm/issues/1130 when merged. Currently the Dockerfile will run trinity without any of the flags. Will be working on adding the needed functionality to use a custom genesis config, add blocks and then run the client.


holiman:
As this is labelled WIP, I haven't really bothered with it. Do you want it merged?

cpurta:
Yes eventually. Currently waiting on some finalized changes to trinity to support a custom genesis file. Once that is done I'll update the startup script with finalized trinity options. Apologies that this has sat around for a while.

PR #120, opened 2018-08-31
Fixing Issue 111: Add json output of build error

SomniaStellarum:
Correcting issue 111 (https://github.com/karalabe/hive/issues/111). Issue corrected by creating buildError type to pass client information back to calling function. This type is then used to construct the json output as requested.

### DIFF
--- a/hive.go
+++ b/hive.go
@@ -89,15 +89,14 @@ func mainInHost(daemon *docker.Client, overrides []string, cacher *buildCacher) 
 		b, ok := err.(*buildError)
 		if ok {
 			results.Clients = make(map[string]map[string]string)
-			results.Clients[b.Client()] = make(map[string]string)
-			results.Clients[b.Client()] = map[string]string{"Failure": "Error building docker container for client"}
+			results.Clients[b.Client()] = map[string]string{"error": b.Error()}
+			out, errMarshal := json.MarshalIndent(results, "", "  ")
+			if errMarshal != nil {
+				log15.Crit("failed to report results. Docker Failed build.", "error", err)
+				return err
+			}
+			fmt.Println(string(out))
 		}
-		out, errMarshal := json.MarshalIndent(results, "", "  ")
-		if errMarshal != nil {
-			log15.Crit("failed to report results. Docker Failed build.", "error", err)
-			return err
-		}
-		fmt.Println(string(out))
 		return err
 	}
 	// Smoke tests are exclusive with all other flags
--- a/hive.go
+++ b/hive.go
@@ -86,6 +86,18 @@ func mainInHost(daemon *docker.Client, overrides []string, cacher *buildCacher) 
 	// Retrieve the versions of all clients being tested
 	if results.Clients, err = fetchClientVersions(daemon, *clientPattern, cacher); err != nil {
 		log15.Crit("failed to retrieve client versions", "error", err)
+		b, ok := err.(*buildError)
+		if ok {
+			results.Clients = make(map[string]map[string]string)
+			results.Clients[b.Client()] = make(map[string]string)
+			results.Clients[b.Client()] = map[string]string{"Failure": "Error building docker container for client"}
+		}
+		out, errMarshal := json.MarshalIndent(results, "", "  ")
+		if errMarshal != nil {
+			log15.Crit("failed to report results. Docker Failed build.", "error", err)
+			return err
+		}
+		fmt.Println(string(out))
 		return err
 	}
 	// Smoke tests are exclusive with all other flags
--- a/images.go
+++ b/images.go
@@ -81,11 +81,13 @@ func fetchClientVersions(daemon *docker.Client, pattern string, cacher *buildCac
 
 		blob, err := downloadFromImage(daemon, image, "/version.json", logger)
 		if err != nil {
-			return nil, err
+			berr := &buildError{err: err, client: image}
+			return nil, berr
 		}
 		var version map[string]string
 		if err := json.Unmarshal(blob, &version); err != nil {
-			return nil, err
+			berr := &buildError{err: err, client: image}
+			return nil, berr
 		}
 		versions[client] = version
 	}
@@ -154,7 +156,7 @@ func buildNestedImages(daemon *docker.Client, root string, pattern string, kind 
 }
 
 type buildError struct {
-	err error
+	err    error
 	client string
 }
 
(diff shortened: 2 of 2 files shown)


holiman:
Nice. Could you paste the output of hitting the error here ?

hive.go
```diff
@@ -86,6 +86,18 @@ func mainInHost(daemon *docker.Client, overrides []string, cacher *buildCacher)
 	// Retrieve the versions of all clients being tested
 	if results.Clients, err = fetchClientVersions(daemon, *clientPattern, cacher); err != nil {
 		log15.Crit("failed to retrieve client versions", "error", err)
+		b, ok := err.(*buildError)
+		if ok {
+			results.Clients = make(map[string]map[string]string)
+			results.Clients[b.Client()] = make(map[string]string)
+			results.Clients[b.Client()] = map[string]string{"Failure": "Error building docker container for client"}
```
holiman: This looks odd. `results.Clients[b.Client()] = make(map[string]string)` initializes and empty map at `results.Clients[b.Client()]`, which is then overwritten on the line after. 
IIUC youcna just remove L92

hive.go
```diff
@@ -86,6 +86,18 @@ func mainInHost(daemon *docker.Client, overrides []string, cacher *buildCacher)
...
+		}
+		out, errMarshal := json.MarshalIndent(results, "", "  ")
+		if errMarshal != nil {
+			log15.Crit("failed to report results. Docker Failed build.", "error", err)
+			return err
+		}
+		fmt.Println(string(out))
```
holiman: Shouldn't the entire output be within the same `if ok{` - clause? As it is, you'll almost certainly be marshalling a `nil` object if some other type of error occurred.

holiman:
I did some tests on it. When running `hive -sim ethereum/consensus -client baz -test none -loglevel 8`

## Bad dockefile 

Bad dockerfile, example 1
```
CRIT[09-01|13:31:41] failed to retrieve client versions       error="clients/bazonk: The command '/bin/sh -c rm /version.json' returned a non-zero code: 1"
{
  "clients": {
    "bazonk": {
      "Failure": "Error building docker container for client"
    }
  }
}
```

Bad dockerfile, example 2:

```
CRIT[09-01|13:32:31] failed to retrieve client versions       error="clients/bazonk: API error (400): {\"message\":\"Dockerfile parse error line 22: unknown instruction: BLARG\"}\n"
{
  "clients": {
    "bazonk": {
      "Failure": "Error building docker container for client"
    }
  }
}

```
I would prefer it if the actual error message was propagated into the json: 
```
{
  "clients": {
    "bazonk": {
      "error": "API error (400): {\"message\":\"Dockerfile parse error line 22: unknown instruction: BLARG\""
    }
  }
}
```
Also, please use lowercase for the `error` instead of `Failure`, it's consistent with the other fields in the case of success. 

## Other errors

When the version.json is missing, we get no output: 

``` 
CRIT[09-01|13:36:23] failed to retrieve client versions       error="API error (404): {\"message\":\"Could not find the file /version.json in container 8683cf3f00ee732bb1a69aec7307f332cda7279afc3424e234bf8358dabee916\"}\n"
{}
```

Should also propagate error: 
```
{
  "clients": {
    "bazonk": {
      "error": "API error (404): {\"message\":\"Could not find the file /version.json in container 8683cf3f00ee732bb1a69aec7307f332cda7279afc3424e234bf8358dabee916\"
    }
  }
}
```

hive.go
SomniaStellarum: I originally wrote it without line 91 and 92 but it gave an uninitialized map error. I can try removing line 92, but I think it will be the same uninitialized map error. I'll let you know what I find.

hive.go
SomniaStellarum: I think you're right. I'll move this code so it's under the same `if ok{` clause.

SomniaStellarum:
I can't comment on your last comment so I'm writing my thoughts here.

For propagating the error through to the json output, I can certainly do that. My only question is would you want the error from the final returned err (line 87 of hive.go) or the original err located within the buildError struct? Also, would you want additional context tagged on there (like `error object buildError created due to docker build fail: *then the error`)? Let me know what context you would want added.

I might refactor this slightly so that I'm calling a function that returns buildError within the images.go file. This will let me add this error to the case when the version.json file is missing. If you want this to apply to other errors throughout the execution of hive, I think it would be outside the scope of this issue, but I don't mind helping once this issue is completed.

In any case, I'm not sure how much time I'll have to finish this up this weekend, but Monday morning I should be able to resubmit the PR.

holiman:
> If you want this to apply to other errors throughout the execution of hive, I think it would be outside the scope of this issue, but I don't mind helping once this issue is completed. 

It's not needed. The problem I want to solve, which I have today, is that sometimes changes in the upstream clients causes the docker build to fail. And that means that they all of a sudden stop appearing at http://hivetests.ethstats.net/ . In order to figure out why they're not there any longer, I have to first notice that they're missing, and secondly ssh into the machine and dig up the right logfiles. 

So 1) showing that the build failed is the first improvement, and 2) showing the error message of why it failed is even better, saving me from having to ssh into the box. 

Other errors throughout the execution is not something that I've had to deal with a lot, so I'm not very concerned about that. 

I don't really care if the error is wrapped in another error or not, whichever is easiest. The main thing is that the original error is there somewhere.

SomniaStellarum:
I've made all the changes. I couldn't recreate the error not getting reported with the clients in the repository though. I think this is because the dockerfile references the version.json file, so it never gets past the docker build with those examples. I did add the build error struct to the two places where I think a missing version.json would error out, but I couldn't test that directly. Can you either let me know if this worked or give me a client that would give the error you're seeing (client: bazonk in your example).

holiman:
I'll check it out later. I just copied `aleth:nightly` into `gazonk`, then added `RUN rm /version.json` into the dockerfile to test that. And then I added a bad command to make the docker file fail for the other example.

SomniaStellarum:
I added the `RUN rm /version.json` to the aleth:nightly client (just on my side) and it seems to report the error correctly. I think everything should work now, so I'll submit work done on the gitcoin side, but let me know if you need any more changes before approving the work. Also, I had a blast working on this. Hopefully I can find some more ways to help out. Cheers.

holiman:
Great to hear! I'll try to check this out asap so you can get the bounty approved. I might be tied for time this week, going to berlin for a few days though, but I'll do my best. 

I'm sure there are other pieces that can be bountied, one thing being to add a configurable timeout for the containers that are being run. I'm having some problems right now, due to https://github.com/paritytech/parity-ethereum/issues/5571 and https://github.com/ethereum/aleth/issues/5243

holiman:
See https://github.com/karalabe/hive/issues/122

holiman:
Well done!

PR #130, opened 2018-10-05
Fixing Issue 122 Make the simulator able to specify a kill-timeout on containers

SomniaStellarum:
This is to fix issue 122. Make the simulator able to specify a kill-timeout on containers

### DIFF
--- a/simulator.go
+++ b/simulator.go
@@ -212,21 +212,35 @@ type simulatorAPIHandler struct {
 	lock   sync.RWMutex
 }
 
+// CheckTimeout is a goroutine that checks if the timeout has passed and stops
+// container if it has.
 func (h *simulatorAPIHandler) CheckTimeout() {
 	for {
 		for id, c := range h.nodes {
 			h.lock.Lock()
 			if !c.State.Running || (time.Now().After(h.nodesTimeout[id])) {
-				delete(h.nodes, id)
-				delete(h.nodesTimeout, id)
-				h.logger.Debug("deleting client container", "id", id)
-				if err := h.daemon.RemoveContainer(docker.RemoveContainerOptions{ID: id, Force: true}); err != nil {
-					h.logger.Error("failed to delete client ", "id", id, "error", err)
-				}
+				h.terminateContainer(id, nil)
 			}
 			h.lock.Unlock()
 		}
-		time.Sleep(time.Duration(*timeoutCheck) * time.Second)
+		time.Sleep(timeoutCheckDuration)
+	}
+}
+
+func (h *simulatorAPIHandler) terminateContainer(id string, w http.ResponseWriter) {
+	node, ok := h.nodes[id]
+	delete(h.nodes, id) // Almost correct, removal may fail. Lock is too expensive though
+	delete(h.nodesTimeout, id)
+
+	if !ok {
+		h.logger.Error("unknown client deletion requested", "id", id)
+		http.Error(w, "not found", http.StatusNotFound)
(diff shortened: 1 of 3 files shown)


simulator.go
```diff
@@ -202,13 +204,46 @@ type simulatorAPIHandler struct {
...
+}
+
+func (h *simulatorAPIHandler) terminateContainer(id string, w http.ResponseWriter) {
+	node, ok := h.nodes[id]
+	delete(h.nodes, id) // Almost correct, removal may fail. Lock is too expensive though
+	delete(h.nodesTimeout, id)
+
+	if !ok {
+		h.logger.Error("unknown client deletion requested", "id", id)
+		http.Error(w, "not found", http.StatusNotFound)
```
holiman: Forgot a `nil`-check for `w` here?

simulator.go
```diff
@@ -202,13 +204,46 @@ type simulatorAPIHandler struct {
...
 	result *simulationResult
 	lock   sync.RWMutex
 }
 
+// CheckTimeout is a goroutine that checks if the timeout has passed and stops
+// container if it has.
+func (h *simulatorAPIHandler) CheckTimeout() {
+	for {
+		for id, c := range h.nodes {
+			h.lock.Lock()
```
holiman: I think this is racy -- you should probably `RLock` h.nodes. Might be less hassle to simply lock it outside the loop.

simulator.go
```diff
@@ -202,13 +204,46 @@ type simulatorAPIHandler struct {
...
+			if !c.State.Running || (time.Now().After(h.nodesTimeout[id])) {
+				h.terminateContainer(id, nil)
+			}
+			h.lock.Unlock()
+		}
+		time.Sleep(timeoutCheckDuration)
```
holiman: Nitpick, it's good practice to have a close channel, which can be fired when the program is shutting down. Doesn't really matter though

SomniaStellarum: I moved the locks outside the loop. I'm not sure if they will take up too much time when it loops through all the containers to check for the timeout. If it does, we can likely do something with RLock and such, but yeah, it's complicated. I usually try to avoid mutex's for the most part (the share memory by communicating thing). Just let me know if you want me to change something.

simulator.go
SomniaStellarum: I moved the locks outside the loop. I'm not sure if they will take up too much time when it loops through all the containers to check for the timeout. If it does, we can likely do something with RLock and such, but yeah, it's complicated. I usually try to avoid mutex's for the most part (the share memory by communicating thing). Just let me know if you want me to change something.

simulator.go
SomniaStellarum: I added the nil check. For some reason I thought if it was a nil http.readwriter it would just not write the error and pass on it's way. To be honest I'm not entirely sure that it won't, but this is probably safer.

simulator.go
SomniaStellarum: I've seen a close channel when the function spins off it's own goroutine so that a user of the library knows there's a goroutine that was started. I could modify the function to do that, but l'm not sure if it's necessary. If you want that, I'll make that change.

Also, the goroutine that's started just after this one (line 189) is ignoring the error from http.Serve. Not sure if you knew about that or if it's intentional, but I thought I'd mention it since I saw it.

SomniaStellarum: Oh, and sorry for being a bit slow on these changes. I have a cold this week and it's just killing me.

PR #147, opened 2018-11-23
log output folder creation

FrankSzendzielarz:


holiman:
Nah, I want all the glory :trollface:

PR #163, opened 2018-12-19
Sync sim

FrankSzendzielarz:
This introduces some basic sync tests, where a node can load a chain and others of different types can attempt to use a similar genesis and sync.

It also attempt to fix some container timeout problems in the simulator.go

PR #187, opened 2019-03-05
Fix typo in go-ethereum_master that breaks all tests for that client 

meowsbits:
Was causing:

DBUG[03-05|13:25:54] starting container                       client=go-ethereum_master validator=smoke/genesis-chain-blocks id=b6ef6d4e
[b6ef6d4e] /geth.sh: line 66: [: : unary operator expected
[b6ef6d4e] Initializing database with genesis state...
[b6ef6d4e] WARN [03-05|19:26:19.326] Sanitizing cache to Go's GC limits       provided=1024 updated=666
[b6ef6d4e] INFO [03-05|19:26:19.327] Maximum peer count                       ETH=25 LES=0 total=25
[b6ef6d4e] INFO [03-05|19:26:19.328] Allocated cache and file handles         database=/root/.ethereum/geth/chaindata cache=16.00MiB handles=16
[b6ef6d4e] Fatal: Failed to write genesis block: genesis has no chain configuration
[b6ef6d4e] Fatal: Failed to write genesis block: genesis has no chain configuration

### DIFF
--- a/clients/go-ethereum_master/geth.sh
+++ b/clients/go-ethereum_master/geth.sh
@@ -63,7 +63,7 @@ fi
 
 
 
-if [ "$HIVE_USE_GENESIS_CONFIG" == ""]; then
+if [ "$HIVE_USE_GENESIS_CONFIG" == "" ]; then
 
 	# Override any chain configs in the go-ethereum specific way
 	chainconfig="{}"


whilei:
[This](https://github.com/ethereum/hive/pull/163/files#diff-b4ff80be2bda1f76661a5935639f8d0fR66) would have and should have been caught by running pretty much any test before merging. Is that not happening? Is there a reason there's no CI?

holiman:
@whilei no, it's not happening. That said, we don't often add new clients, nor really maintain the smoke tests. We're primarily using hive for consensus tests nowadays, though we have some more test suites coming up

PR #235, opened 2019-06-13
removing ChainLoader

tkstanczak:
not needed


tkstanczak:
also fixed chainspec creation

PR #249, opened 2019-08-28
istanbul definition for aleth+geth

holiman:
This PR adds istanbul definitions to clients

- [x] geth
- [x] aleth
- [ ] parity
- [ ] nethermind
- [ ] ethereumj
- [ ] trinity


holiman:
Fixed on master

PR #257, opened 2020-04-06
Minor fixes in README

lightclient:
I noticed a few small typos in the README -- as well as a few places where the wording could be a little more clear. This is by no means a thorough pass, just some low hanging fruit!


holiman:
Thanks!

PR #263, opened 2020-04-18
Make hive operational again

holiman:
A massive PR which brings along quite a lot of needed maintenance. To name a few: 

- Graphql tests now actually work (they previously failed, but were reported as successfull). Geth + Besu enabled. 
- Aleth fixed up for Istanbul+
- Ability to download simulator log file, which is quite important
- Expose log file even if client container doesn't open RPC port
- Several more cmd line options, e..g to make it possible to run tests more limited, for local testing
- Make "popups" use regular windows/tabs instead of pesky floating windows

Changes some cmd line params too: 

```
     sim-parallelism => sim.parallelism
     simlimiter => sim.testlimit
     sim-rootcontext => sim.rootcontext
     simloglevel => sim.loglevel
```


holiman:
I don't want this squashed. The commits are pretty unrelated to eachother. I might squash it a bit manually before I merge it

PR #287, opened 2020-05-07
add besu fast-sync config flags for contrived test chains

RatanRSur:
hyperledger/besu#797 brought up a fast sync issue that is a result of the test chain not conforming to certain configured minimums.

These flags are there to protect against chain reorgs and bad pivot-block-selection while fast syncing. I'm curious why geth's sync doesn't need these precautions! If you have a link to a resource that explains how geth's sync gets around these that would be lovely.

### DIFF
--- a/clients/besu/besu.sh
+++ b/clients/besu/besu.sh
@@ -54,7 +54,8 @@ else
 fi
 
 if [ "$HIVE_NODETYPE" == "full" ]; then
-	FLAGS="$FLAGS --sync-mode=FAST"
+	FLAGS="$FLAGS --sync-mode=FAST --fast-sync-min-peers=1 --Xsynchronizer-fast-sync-pivot-distance=0"
+
 fi
 if [ "$HIVE_NODETYPE" == "light" ]; then
     echo "Besu does not support light nodes"


holiman:
Well, if you explain why you need them, that'd be easier to answer :) 

Some reasons to _not_ have those things: 
- When private networks are used, they are often pretty small. So if you spin up a couple of internal nodes on a private network, it's nice if those can sync with eachother. 
- That said, we did see a fast-sync 'tar pit' attack on mainnet nodes ~a year ago. What happened was that someone was scanning the DHT for new nodes, and instantly peered with them, and presented them with a fast-sync: "Hey I've got the full state, at height `4`", and the victim very quickly performed a fast-sync against the malicious node. And then had to block-by-block import the remaining 9 million nodes. 
  - As a response to that, we made use of the CHT (canonical hash tables), and simply choose not to sync to a pivot point below the last CHT block. 

But please let me know what other dangers you see, and I can probably answer why geth does or does not need them.

holiman:
Ok, prod server updated, I guess we'll see if it works within ~24 hours or so

PR #290, opened 2020-06-02
Pass only request name to send instead of the whole request

adamschmideg:
Before:
```
t.sendPacket(toid, toaddr, req, packet, callback)
```
where `req` is the unencoded packet used only to get its name.

After
```
t.sendPacket(toid, toaddr, req.name(), packet, callback)
```
Now the intention is clear.

### DIFF
--- a/simulators/devp2p/udp.go
+++ b/simulators/devp2p/udp.go
@@ -353,7 +353,7 @@ func (t *V4Udp) SpoofedPing(toid enode.ID, tomac string, toaddr *net.UDPAddr, fr
 
 	}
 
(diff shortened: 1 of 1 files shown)


holiman:
Why is this an improvement?

adamschmideg:
It's not a big improvement by itself but a part of a bigger refactoring to write dockerless tests.  I thought it's better to break it down into smaller PRs.

It's to make it possible to test certain cases. The current test passes a dummy `&ping{}`  (so its name can be used) and passes a different packet with some junk metadata.

adamschmideg:
I'm taking a different path implementing it first outside Hive.

PR #302, opened 2020-06-29
Update Nethermind Dockerfile for hive specific docker 

SebastianDremo:


holiman:
What is the benefit of doing this -- and what is the difference between the two versions? We want hive to use the 'real' thing, if possible.  
If it's for dev/testing, you can just define another (`nethermind2`) client which uses whatever docker image you want.

SebastianDremo:
nethermind/nethermind:alpine is from up to date Nethermind's release (so I guess we can take it as 'real' thing) 

nethermindeth/nethermind-hive:alpine is nightly build from master branch

Let's assume the scenario in which I make some changes to hive-integration module but did some mistakes and pushed it to our master.
This bug goes to the newest release and ends up failing all of the Hive's tests on public board.
Right after I find out that the tests are not passing I push the fix to master but in order to see if the Hive's public tests are green now I have to wait until next release so like 2-3 weeks.

I agree that with this approach we will be looking at Nethermind's master branch and not the newest package so maybe creating the second client is better idea if Hive is more for production testing

holiman:
I see... Another approach which is pretty common, is to do as we've done with geth here: https://github.com/ethereum/hive/blob/master/clients/go-ethereum/Dockerfile#L8 

So we define only one geth-client, but if we run it woth `hive ... -client go-ethereum_latest`, then the base image will be `ethereum/client-go:latest`. Thus, when running it, we can decide if we want to run `latest` or `stable` or something else. 
So perhaps you can use the same approach? Geth uses `latest` as in "whatever is committed to master", and `stable` as "latest official release". 
Why do you use `nethermindeth` instead of `nethermind`?

tkstanczak:
@holiman we initially created a user with nethermind and then because you have to chose a different name for organization we picked nethermindeth. Tried to change it later with docker but they do not support any changes for the account names.

PR #318, opened 2020-09-02
Make clients discoverable so bonding can occur; Fix incorrect graphql port

davemec:
Make clients discoverable so bonding can occur
- Besu requires a bonded node ino order to start syncing

Fix incorrect graphql port
- Startup scripts were modified to use port 8550 so update graphql tests to also use that port


holiman:
Hm, the `nodiscovery` is to prevent the geth/OE node from reaching out into the world trying to discover peers, it doesn't make them un-bondable. Does changing this fix besu problems? What particular problems are fixed by this?

holiman:
But that doesn't make sense.. Geth _should_ stil respond to `PING` even when `--nodiscover` is used, it should just stop the geth node from starting it's own discovery mechanism. Or am I wrong, @fjl ?

holiman:
I mean, I don't particularly want to be the blocker on this, I just want to understand what is not working and why, and ensure that we fix it the right way

fjl:
The `--nodiscover` option in geth turns off the discovery listener entirely, so it won't be reachable. If you want to run the discovery but prevent it from talking to the internet, you need to use `--netrestrict`.

fjl:
I think we should try to make the test work without discovery, so the besu launcher needs to change to use RPC `admin_addPeer` for the connection.

davemec:
The test did use the rpc end point but was changed to use the boot nodes. I
can update the flag to --netrestrict but I don't see the need to run the
geth node as completely non-discoverable.

holiman:
I'm looking at the tests now.. So, afaict, the [code](https://github.com/ethereum/hive/blob/master/simulators/ethereum/graphql/graphql.go) does not spin up any geth-node, when `besu` is the client under test. So no bonding occurs, because there simply aren't any other clients involved. So I don't see how this could possibly fix any errors? https://github.com/ethereum/hive/blob/master/simulators/ethereum/graphql/graphql.go

I mean, aside from the port-change, that is all good.

holiman:
Ok, this was confusing. I somehow thought the entire PR was about fixing graphql, but it's fixing two separate issues. 

I'm fine with merging this, unless @fjl  has some other idea?

holiman (approved):

PR #319, opened 2020-09-08
chaintools: ineffectual error assignment preventing genesis file from being opened

renaynay:
This PR fixes an ineffectual error assignment that prevented a genesis file from being accepted with the command `chainGenesis`. 

The following error would result: 

```
CRIT[09-08|14:35:38] failed to read genesis json              error=nil
```

Now, if a user specifies the `chainGenesis` command after the `chainGenerate` command, a more descriptive error will be logged:

```
CRIT[09-08|14:37:33] failed to read genesis json              error="open : no such file or directory"
```

This does not solve the problem, however, that the order in which a user specifies `chainGenerate` and `chainGenesis` matters for whether the given filepath+name is actually recognized.

The valid order is: 
```
-chainGenesis <filepath> -chainGenerate <bool>
```

### DIFF
--- a/chaintools/chaintools.go
+++ b/chaintools/chaintools.go
@@ -1,7 +1,6 @@
 package chaintools
 
 import (
-	"fmt"
 	"io"
 	"io/ioutil"
 	"math/big"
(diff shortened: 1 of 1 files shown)


holiman (approved):
LGTM

PR #324, opened 2020-09-23
simulators/ethereum/eth: add eth protocol test suite

renaynay:
This PR adds the eth protocol test suite to hive. 

Run it with the following cmd:
```bash 
./hive --sim ethereum/eth --sim.rootcontext --client <client>_<branch> --docker-noshell
```


renaynay:
- geth working
- besu requests headers and disconnects
- nethermind docker container doesn't build
- openethereum needs protocol version 64
- trinity doesn't build

renaynay:
update: 

- besu working
- openethereum has a bug? (wrong forkID in status and immediately sends PING upon fresh peer connection)
- nethermind can't import chain from chain.rlp file
- parity EOF issue on connecting
- trinity cannot import chain, also for some reason does not support eth 64 (but supports eth63 and eth65)

simulators/ethereum/eth/Dockerfile
```diff
@@ -0,0 +1,20 @@
+# This simulation runs various graphql tests
```
holiman, line 1: Does it?

holiman (commented):
Minor nitpick, otherwise LGTM

simulators/ethereum/eth/Dockerfile
renaynay: No it does not :) you caught me copy pasting.

PR #325, opened 2020-09-24
clients/trinity: fix trinity docker build and startup script

fjl:
With these changes, the client builds and actually starts. It now
supports most of the options that other clients support as well. The
initial chain import is still broken unfortunately, so it can't run the
consensus tests yet.


holiman (approved):
Nice!

PR #326, opened 2020-09-24
clients/trinity: make consensus tests work

fjl:
This fixes a couple more bugs in the client definition.
The consensus test suite works now.


holiman (approved):
I don't want to stand in the way when you're on a roll!

PR #330, opened 2020-09-28
[WIP] docs: update outdated documentation

renaynay:


renaynay:
note to self: 

fix: `//TODO - update docs to say that folders under 'clients' should not use _ underscore` (L191, images.go)

PR #331, opened 2020-09-28
simulators: remove DAO fork tests and DAG generation functionality

renaynay:
This PR removes the DAO fork tests as they no longer work.

It also removes DAG generation functionality as this significantly slows down other simulations (where DAG generation is unnecessary) as well as takes up a non-negligible amount of disk space.


clients/go-ethereum/Dockerfile
```diff
@@ -46,11 +46,4 @@ ADD genesis.json /genesis.json
 # Export the usual networking ports to allow outside access to the node
 EXPOSE 8545 8546 8547 30303 30303/udp
 
-# Generate the ethash verification caches
-RUN \
-  /usr/local/bin/geth makecache     1 ~/.ethereum/geth/ethash && \
-  /usr/local/bin/geth makecache 30001 ~/.ethereum/geth/ethash
-
-
-
```
fjl, line 55: This command is fine and needs to be left in. The cache is much smaller than the DAG and is needed for verification of blocks. Baking it into the client image just makes it run a bit faster.

fjl (changes requested):

holiman (approved):

PR #334, opened 2020-09-30
hive: add error handling for if dockerfile/image can't build or is not found

renaynay:
This PR adds error handling in the cases that: 

**client:**
- can't find user-specified docker image in `clients` dir (in which case, continue test with the clients that have been successfully built, but log critical error)
- no clients get spun up at all (in which case, kill the test)

**simulation:** 
- simulation is not found (in which case, kill the test)
- simulation image build failed (in which case, kill the test)

The errors will be logged both to stderr and to a file `containerErrorReport.json`

### DIFF
--- a/hive.go
+++ b/hive.go
@@ -115,8 +115,8 @@ func main() {
 	}
 	//set up clients and get their versions
 	if err := initClients(cacher); err != nil {
-		log15.Crit("failed to initialize client(s), exiting...")
-		os.Exit(1)
+		log15.Crit("failed to initialize client(s), terminating test...")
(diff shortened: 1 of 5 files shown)


holiman:
This is a good start, but what we really need in the end is also a way to export these failures to a similar json-struct as successfull tests, so that the UI can list them (possibly as a separate table of errors. It might be easier to use a different 'model' for the errors, since the error-runs don't necessarily have things like 'client' and 'testresults')

holiman (approved):
LGTM!

PR #335, opened 2020-10-12
add ability to create networks and connect containers to them

renaynay:


fjl:
Would be better to just create two networks automatically instead of having to specify them on the command line.

renaynay:
Meaning we'd spin up 2 docker networks for every test? @fjl

fjl:
Yes, that's what I meant. I think it should basically create two networks when starting hive and simply connect all containers which are launched to these two networks.

PR #343, opened 2020-10-19
hive: improve client name matching

fjl:
This allows running hive with client names that do not contain the
branch explicitly (e.g. "go-ethereum" instead of "go-ethereum_latest").
If there is no branch name, the default from the Dockerfile is used.

This also removes the substring-based matching of client names and
requires an exact match instead. It was previously possible to run hive
with weird patterns such as `--client e_latest`, matching all client names
containing 'e'. The client names now need to be spelled out in full, which
is more predictable.


renaynay (approved):
LGTM

PR #345, opened 2020-10-20
cmd/hivechain: chaintools cli tool

fjl:
This turns the chaintools package into a stand-alone command-line tool.
The new tool includes a command to generate chains like hive previously
did, but can also display the content of chain.rlp files. It also includes a utility
to display the post state root of genesis.json files.

I needed this to debug some client integration issues.


holiman (approved):
LGTM

PR #351, opened 2020-10-22
cmd/hiveview: fix more navigation issues

fjl:
This makes it possible to click 'Load' when a test suite was already
loaded once. This issue was always present, but we didn't notice because
it was impossible to navigate back and forth before the navigation fix.

I also fixed a bunch of issues in the file viewer, it's now more robust and
has a loading indicator.


holiman (commented):
LGTM

PR #356, opened 2020-10-27
hive: add simulation API for creating docker networks

renaynay:
This PR adds functionality to create docker networks and connect client and simulator containers to them, allowing for more versatile networking if necessary in simulations.


renaynay:
This PR is a copy of #335 without the additional IP test (which was only used for debugging).

Keeping #335 open for now to preserve discussion.

simulator.go
```diff
@@ -285,12 +297,98 @@ func nodeInfoGet(w http.ResponseWriter, request *http.Request) {
 	io.WriteString(w, fixedIP.URLv4())
 }
 
+// getSimulatorID gets the container ID of the simulation container.
+func getSimulatorID(w http.ResponseWriter, request *http.Request) {
```
fjl, line 301: I think adding this is a bit awkward. The only reason we need this is because the simulation container needs to
connect itself to the network. Maybe we could add another API endpoint that connects the simulation to a named network:

- `/testsuite/{suite}/node/{node}/network/{network}` connects the node
- `/testsuite/{suite}/network/connect` connects the simulation

fjl:
It doesn't need to be complicated, just launch any client from the available ones and hit the new endpoints. Maybe the test should be like:

- Launch one client
- Create network
- Connect client to network
- Connect simulation to network
- TCP dial to `<IP>:8545`
- Delete network again

simulator.go
fjl: We could also remove the extra APIs specific to the simulation and just give special meaning to the `"simulation"` node container ID. Just really trying to avoid the situation where the sim container needs to learn about its own docker container ID. It feels wrong.

So instead of having two APIs for 'connecting nodes' and 'connecting simulation', we could have

`GET /testsuite/{suite}/network/{network}/{node}`
`POST /testsuite/{suite}/network/{network}/{node}`
`DELETE /testsuite/{suite}/network/{network}/{node}`

and if `node` is `"simulation"`, it gets-ip/connects/disconnects the simulation container.

simulator.go
```diff
@@ -172,16 +181,21 @@ func startTestSuiteAPI() error {
...
+	router.HandleFunc("/testsuite/{suite}/test/{test}/pseudo", pseudoStart).Methods("POST")
+	router.HandleFunc("/testsuite/{suite}/test", testStart).Methods("POST")
+	router.HandleFunc("/testsuite/{suite}/test/{test}", testDelete).Methods("POST") //post because the delete http verb does not always support a message body
+	router.HandleFunc("/testsuite", suiteStart).Methods("POST")
+	router.HandleFunc("/testsuite/{suite}", suiteEnd).Methods("DELETE")
+	router.HandleFunc("/testsuite/{suite}/network/{network}", networkCreate).Methods("POST")
+	router.HandleFunc("/testsuite/{suite}/network/{network}", networkRemove).Methods("DELETE")
+	router.HandleFunc("/testsuite/{suite}/network/{network}/node/{node}", networkIPGet).Methods("GET")
+	router.HandleFunc("/testsuite/{suite}/network/{network}/node/{node}", networkConnect).Methods("POST")
+	router.HandleFunc("/testsuite/{suite}/network/{network}/node/{node}", networkDisconnect).Methods("DELETE")
```
fjl, line 198: I think these could just be `/testsuite/{suite}/network/{network}/{node}`

go.mod
```diff
@@ -3,54 +3,47 @@ module github.com/ethereum/hive
...
+	github.com/aristanetworks/goarista v0.0.0-20201012165903-2cb20defcd66 // indirect
+	github.com/btcsuite/btcd v0.21.0-beta // indirect
 	github.com/cespare/cp v1.1.1 // indirect
+	github.com/containerd/containerd v1.4.1 // indirect
+	github.com/containerd/continuity v0.0.0-20200928162600-f2cc35102c2a // indirect
 	github.com/deckarep/golang-set v1.7.1 // indirect
 	github.com/edsrzf/mmap-go v1.0.0 // indirect
-	github.com/elastic/gosigar v0.10.4 // indirect
-	github.com/ethereum/go-ethereum v1.9.1
+	github.com/ethereum/go-ethereum v1.9.23
```
fjl, line 16: It seems that you somehow upgraded all the dependencies here. Please revert the go.mod changes.

renaynay: ah yeah i ran a go mod tidy

PR #360, opened 2020-10-30
clients: rewrite client genesis translators with jq

fjl:
This replaces the client genesis translator scripts with a single jq program
for each client. Previously, the client bash scripts constructed the jq program
by concatenating snippets in bash. Having the jq script in a jq file is better
because there is no need to escape quotes, and advanced conversions (e.g. hex
encoding of fork numbers and chainID) can be applied.

This also makes the GraphQL tests work again with go-ethereum. 

The sync and eth protocol tests now work with OpenEthereum because the chainID
is now translated correctly.


holiman (approved):
Looks very good to me!

PR #361, opened 2020-10-30
cmd/hiveview: remove description in listing

fjl:
This removes the `description` key from the listing file because it's not used
by the viewer. Also changes the default sort order of the tests table to show the
failed tests first.

### DIFF
--- a/cmd/hiveview/assets/app.js
+++ b/cmd/hiveview/assets/app.js
@@ -440,6 +440,7 @@ function onSuiteData(data, jsonsource) {
     var thetable = $('#execresults').DataTable({
         data: cases,
         pageLength: 100,
+        order: [[2, 'desc']],
         columns: [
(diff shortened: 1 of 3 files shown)


holiman:
It's not used? Are you sure it isn't used as part of this? 
Screenshot_2020-10-30 https hivetests ethdevops io

fjl:
Yes, but on that page, the description is taken from the suite result file. The listing file is just the overview of all the suites, and the description is not displayed there.

fjl:
The description being removed in the listing JSON is this one:

Screenshot 2020-10-30 at 12 46 15

This is only on the suite results page, and not on the main overview page.

holiman:
Ah ok, so this change still leaves it in place on the result-page? Then it's all good

holiman (approved):

PR #363, opened 2020-11-02
simulators/devp2p/discv4: use discv4 test suite from go-ethereum

renaynay:
This PR replaces the current devp2p tests in hive with the new discv4 test suite. It uses the new network API so a pseudoclient is not required to relay traffic to simulate a separate IP address.


simulators/common/testmanager.go
```diff
@@ -143,6 +143,7 @@ func (manager *TestManager) CreateNetwork(testSuite TestSuiteID, networkName str
 	// create network
 	network, err := manager.dockerClient.CreateNetwork(docker.CreateNetworkOptions{
 		Name:           networkName,
+		Driver:         "bridge",
```
renaynay, line 146: @fjl should this be a default, or should this be specified by the user? Maybe at one point we'll want two totally isolated networks or something? Wdyt?

fjl: I think it's OK to just put this in for now, we can always add more config possibilities later when we actually need it.

simulators/discovery/version4/discv4.go
```diff
@@ -0,0 +1,151 @@
...
+		err = host.ConnectContainer(suiteID, networkID, "simulation")
+		if err != nil {
+			endTest(&common.TestResult{
+				Pass:    false,
+				Details: err.Error(),
+			})
+			fatalf("could not connect container to network: %v", err)
+		}
+		// get bridge network ID
+		bridgeID, err := host.GetNetworkID(suiteID, "bridge")
```
fjl, line 103: This part I don't like. It's a lot like the situation we had before with the simulation container. I think we should solve this by having a special, `"default"` network ID, which is recognized by the API and translated into the actual network ID of the default network.

simulators/ethereum/eth/eth.go
```diff
@@ -14,7 +14,7 @@ func main() {
 	host := hive.New()
 
 	//	get clients
-	// for loop that runs tests against each client type
+	// loop that runs tests against each client type
```
fjl, line 17: Please revert this. I have a commit where I removed this comment completely, and it will just cause a conflict.

simulators/discovery/version4/discv4.go
```diff
@@ -0,0 +1,151 @@
+package main
```
fjl, line 1: I think this should be in simulators/devp2p. The eth test should also be there, but we can move it some other time.

simulators/common/testmanager.go
```diff
@@ -189,11 +190,16 @@ func (manager *TestManager) ContainerIP(testSuite TestSuiteID, networkID, contai
...
 	}
 	// if the containerID is "simulation", use simulation container ID
-	containerID, err := manager.isSimulation(containerID)
+	var err error
+	containerID, err = manager.isSimulation(containerID)
+	if err != nil {
+		return "", err
+	}
+	// if the networkID is "bridge", use bridge networkID
+	networkID, err = manager.isBridge(networkID)
```
fjl, line 199: `isBridge` is not the best name, would be better to call it `findNetwork`.

fjl: Also, I'm not sure if it will work. The idea with this function seems to be resolving the network ID from the network name, but `isBridge` just returns "bridge" for the input "bridge". Might be better to restructure things so the network ID is never required and all operations are by-name.

PR #366, opened 2020-11-03
Simulation API Rewrite

fjl:
This change restructures the simulation API server and client.

Before this change, the server implementation lived in the top-level 'package main' of hive, and the client was located in simulators/common/providers/hive. For all Go-based simulations, we basically copied this directory into the docker container so that the simulation implementation could link it.

There are a couple of problems with this design:

- When building simulation containers, the docker build context has to be simulators/ and not the actual simulation directory because the Dockerfile can't access simulators/common otherwise. This is why we have `--sim.rootcontext`, which is very hard to understand for simulation authors.

- All Go-based simulations are technically part of the main 'github.com/ethereum/hive' Go module. This is annoying because it adds a lot of specialized dependencies to the module even though they are only used by the simulation. However, when building the simulation inside of its container, the go.mod of hive is no longer there and we need to run `go mod init` in the simulation directory or build with Go modules disabled. This makes it impossible to use dependency version pinning and other go modules features in simulations.

- On top of all that, the API exposed by simulators/common is actually horrible and writing simulations isn't fun at all.

- It's also not possible to create real unit tests for the API client because the server implementation lives in hive's package main, which cannot be imported.

Here's how all of these issues can be resolved:

The simulation API server now lives in a separate package,`internal/libhive`. I've also fixed a couple of annoying bugs in the server, which I found because we now have some basic unit tests for the API.

The new API client lives in the top-level package `hivesim`. This new package has a much cleaner API than simulators/common/providers/hive. For example, it is now possible to write simulations that look like Go unit tests.

As of this PR, the new `hivesim` client is not used by any simulations. This is because I'm planning to incrementally convert all simulations into Go modules which import `hivesim` in the normal Go module way. In order for this to work, we need to have the client committed to the master branch first.

There is a downside to this new approach: changes to the simulation API client can no longer be made in the same PR that also updates simulations. Instead, any changes to `hivesim` will first need to hit the master branch, and then the module dependency will need to be updated in simulations. I think we can make this easier by automating the dependency update with a shell script though.


hivesim/hive.go
```diff
@@ -0,0 +1,307 @@
...
+// parameter must be named CLIENT and should contain one of the client types from
+// GetClientTypes. The input is used as environment variables in the new container.
+// Returns container id and ip.
+func (sim *Simulation) StartClient(testSuite SuiteID, test TestID, parameters map[string]string, initFiles map[string]string) (string, net.IP, error) {
+	// vals := make(url.Values)
+	// for k, v := range parameters {
+	// 	vals.Add(k, v)
+	// }
+	//vals.Add("testcase", test.String())
+	//	parameters["testcase"] = test.String()
```
renaynay, line 132:
```suggestion
```

internal/hive/data.go
```diff
@@ -0,0 +1,62 @@
...
+// TestResult is the payload submitted to the EndTest endpoint.
+type TestResult struct {
+	Pass    bool   `json:"pass"`
+	Details string `json:"details"`
+}
+
+// ClientInfo describes a client that participated in a test case.
+type ClientInfo struct {
+	ID              string    `json:"id"`
+	IP              string    `json:"ip"`
```
renaynay, line 55: Maybe this should rather be a `map[string]net.IP` where key is the network name and value is the ip addr of the client on that network considering a client could be connected to several networks?

fjl: We could change that later, but for now, we can't make any breaking changes to the API. It still needs to be compatible with simulators/common so the simulations keep working.

PR #370, opened 2020-11-03
hivesim, internal/libhive: update docker network API

renaynay:
This updates the simulation API for docker networks. Networks are now scoped to the
test suite and will be removed when the suite ends. Simulations can now refer to their
networks by name instead of requiring them to keep track of the docker network ID.


internal/libhive/docker.go
```diff
@@ -187,33 +188,29 @@ func (b *dockerBackend) StopContainer(containerID string) error {
...
-		return "", err
-	}
-	// check for existing networks with same name, and if exists, remove
-	for _, exists := range existing {
-		if exists.Name == name {
-			if err := b.client.RemoveNetwork(exists.ID); err != nil {
-				return "", err
-			}
+func (b *dockerBackend) CreateNetwork(name string) (string, error) {
+	existing, err := b.NetworkNameToID(name)
```
fjl, line 192: I'm not sure about this. I think it should just fail if the network already exists.

internal/libhive/testmanager.go
```diff
@@ -143,11 +143,10 @@ func (manager *TestManager) Terminate() error {
 		}
 		// ensure the db is updated with results
 		manager.doEndSuite(suiteID)
+		// remove the test suite's left-over docker networks.
+		manager.PruneNetworks(suiteID)
```
fjl, line 147: This needs to move into `doEndSuite` so that it will also remove the networks when the suite ends in the normal way.

internal/libhive/testmanager.go
```diff
@@ -169,55 +168,84 @@ func (manager *TestManager) GetNodeInfo(testSuite TestSuiteID, test TestID, node
...
 	if err != nil {
-		manager.networks[netID] = networkName
+		return err
 	}
-	return netID, err
+	if _, exists := manager.networks[testSuite]; !exists {
+		// initialize network map for individual test suite
+		manager.networks[testSuite] = make(map[string]string)
+		// add bridge network information
+		bridgeID, err := manager.backend.NetworkNameToID("bridge")
```
fjl, line 191: I'm not sure about this. I think it would be easier to resolve "bridge" the same way as before, i.e. just check for the special name in ContainerIP.

fjl (approved):

PR #378, opened 2020-11-04
simulators/devp2p/eth: integrate eth protocol test TAP output

fjl:
This changes the eth simulation to use the new TAP output feature of the
test tool. The simulation now reports results from individual protocol
tests to hive.

The simulation now uses package hivesim and is a stand-alone Go
module. I also moved it to simulators/devp2p because we wanted to have
it there anyway, and renaming it in this PR makes the Go module migration
easier.


fjl:
Before:

Screenshot 2020-11-04 at 19 19 11

After:

Screenshot 2020-11-04 at 19 18 50

PR #379, opened 2020-11-04
cmd/hiveview: show test details as preformatted text, not markdown

fjl:
The markdown is nice, but requires a lot of effort in the test to format
it correctly. With the newer simulation API client where you can basically
write simulations like Go unit tests, outputting lines with `t.Log(...)` gets really
messy if it has to generate valid markdown.

This simplifies it a bit by just showing the text with `<pre>`.

I think the most useful feature of the markdown conversion was being able
to link to stuff. This is preserved with this change by making all URLs in the
output clickable.


fjl:
Before:

Screenshot 2020-11-04 at 19 19 11

After:

Screenshot 2020-11-04 at 19 18 50

fjl:
Here's what a test description containing a link looks like. I think it's OK and preserves the useful functionality completely.

Screenshot 2020-11-05 at 01 51 57

cmd/hiveview/assets/app.js
```diff
@@ -14,6 +14,19 @@ utils = {
...
+    // Replaces URLs in input with HTML links.
+    urls_to_links: function(str) {
+        // Thanks, http://urlregex.com/
+        let re = /(((?:http[s]?:\/\/)(?:[\-;:&=\+\$,\w]+@)?[A-Za-z0-9\.\-]+|(?:www\.|[\-;:&=\+\$,\w]+@)[A-Za-z0-9\.\-]+)((?:\/[\+~%\/\.\w\-_]*)?\??(?:[\-\+=&;%@\.\w_]*)#?(?:[\.\!\/\\\w]*))?)/
+        return String(str).replace(re, function (match) {
+            var link = '<a href="';
+            link += utils.attr_encode(match) + '">';
+            link += utils.html_encode(match);
+            link += '</a>';
+            return link;
```
holiman, line 26: how about 
```suggestion
return utils.getLink(match, match)
```

holiman (commented):
Minor comment

cmd/hiveview/assets/app.js
fjl: Fixed!

holiman (approved):

PR #390, opened 2020-11-05
simulators/ethereum/consensus: use package hivesim

fjl:
This makes the consensus tests use the new 'hivesim' simulation API
client. The simulation is now a stand-alone module and no longer
requires the --sim.rootcontext option. While here, I also fixed some
staticcheck warnings in the simulation code.

Note that the simulation will behave slightly differently from before
when run with multiple client types: before this change, a separate
suite was started for every client. With the new version, the simulation
will only create a single suite and put all client test results into it.

This behavior change will not be noticed on the production Hive instance
though, because we invoke the simulation separately for every client.
Also, it seems running the simulation against multiple clients didn't actually
work before because their artefacts directories collided. It works now.

I've tested this using the following command, and it does work as expected:

  ./hive --sim consensus --client go-ethereum --docker-noshell --sim.parallelism 8 --sim.testlimit 50


simulators/ethereum/consensus/go.sum
```diff
@@ -0,0 +1,582 @@
+bazil.org/fuse v0.0.0-20160811212531-371fbbdaa898/go.mod h1:Xbm+BRKSBEpa4q4hTSxohYNQpsxXPbPry4JJWOB3LB8=
```
holiman, line 1: Damn, this was big. Is all of it needed?

holiman (approved):
LGTM

simulators/ethereum/consensus/go.sum
fjl: Yeah, it's kind of an annoying issue at the moment. It happens because the hivesim package
imports internal/libhive *in its unit tests*, and Go somehow decides to include all the
dependency checksums of all hive dependencies into this file. It happened with all other
simulations as well. I guess it's best to just not look at go.sum.

PR #394, opened 2020-11-05
simulators/common: delete old simulation API client

fjl:
The new Go client for the API lives in the top-level 'hivesim' package,
and all simulations use it. This means we can now delete the old client.


renaynay (approved):
BEST PR

PR #396, opened 2020-11-06
simulators/smoke/network: fix network smoke test

renaynay:
This PR fixes the network smoke test to match the updates made to the network API.


fjl:
Please add a check for the "bridge" network name. You can even check if the IP is equal to `Client.IP`.

PR #399, opened 2020-11-06
simulators/devp2p/discv4: update to new docker networks API

fjl:
This updates the discovery simulation to work with the new 'by-name'
docker networks API. It also fixes running the simulation with multiple
client types. This previously failed because it tried to create the
network more than once.


renaynay (approved):
LGTM

PR #401, opened 2020-11-09
hive: remove shell container option

renaynay:
This PR deprecates the shell container option, but keep flag as a no-op to not interfere with scripts.


fjl (commented):
Please also remove `buildShell`, and remove the `--docker-noshell` option in the CI script.

PR #402, opened 2020-11-09
documentation: update docs to reflect new APIs / refactoring

renaynay:


README.md
```diff
@@ -1,323 +1,299 @@
...
 
-`--sim.rootcontext` a boolean, which when set tells the compiler to build the docker image with 'simulators'
-as the root of the context, allowing the simulators\common and simulators\devp2p common code to be included
-in the simulator. 
+To get information about the client that is likely necessary for test execution, you can use `hivesim.Client`within the aforementioned test execution function (`myTestFunction`).
 
-Sim.rootcontext needs to be set differently depending on the type of simulation being run. For the consensus tests
-the base simulator image relies on files to be added from a folder local to the image. For developing new simulations,
-or extending the existing ones, it is recommended to use sim-rootcontext as true. 
+**Enode ID**
```
fjl, line 213: Enode ID is a bad term, and we should avoid using it. The correct term for this is 'enode URL'.

README.md
```diff
@@ -404,340 +364,39 @@ Simulation results:
...
-
-# Generating test blockchains
-
-`Hive` allows you to create rlp encoded blockchains for inclusion into sync tests 
-
-## To generate a blockchain offline
-
-`Hive` can be run with the `chainGenerate` setting to generate a blockchain according to specification
-and then exit. The current version only generates blocks with empty transactions, but this will be
+The current version only generates blocks with empty transactions, but this will be
```
fjl, line 380: This is not true, hivechain will generate transactions if certain accounts are present in the genesis block. Check source for the account addresses.

README.md
```diff
@@ -1,323 +1,299 @@
...
-      --chainBlockTime 30
+If you want to rebuild both, separate the names with a `,` as such: 
+```text
+--docker-nocache devp2p,go-ethereum_latest
+```
 
-## Iterating on bugfixes locally
+# Adding a Simulation
+There are two components to a simulation: 
+1. a **simulation program written in Go** using the `hivesim` test API to coordinate the execution of your desired test; and
```
fjl, line 72: I would strongly prefer that we don't say `hivesim` is required. From hive's point of view, a simulation is any program using the simulation API. It doesn't need to be written in Go, and doesn't need to use package hivesim. For documenting package hivesim, we should add the tutorial content you wrote here to the GoDoc instead, i.e. it should be in hivesim/doc.go. Then you can just link to that from here.

README.md
```diff
@@ -1,323 +1,299 @@
...
+need to run with different initial chain setups, a simulator needs to be able to specify these per
+client. To avoid file uploads, `hive` solves this by defining a set of API variables that allow a
+simulator to specify the source paths to use for specific init files which will be extracted from the
+live container:
 
-# Running on Windows
+ * `HIVE_INIT_GENESIS` path to the genesis file to seed the client with (default = "/genesis.json")
+ * `HIVE_INIT_CHAIN` path to an initial blockchain to seed the client with (default = "/chain.rlp")
+ * `HIVE_INIT_BLOCKS` path to a folder of blocks to import after seeding (default = "/blocks/")
+ * `HIVE_INIT_KEYS` path to a folder of account keys to import after init (default = "/keys/")
```
fjl, line 93: These variables are not used anymore. The files are now supplied using the multipart file-upload request to start a node.

fjl:
What we need really badly is technical documentation of the simulation HTTP API.

renaynay:
Yes that makes sense @fjl The plan was to go over it today. I don't like the way the Add Simulation section is written either, so I will do what is mentioned here: 
https://github.com/ethereum/hive/pull/402#discussion_r522005459

hivesim/doc.go
```diff
@@ -1,3 +1,126 @@
-// Package hivesim is a Go wrapper for the Hive Simulation API.
-// You can use this package to write simulations for Hive in Go.
+/*
+	Package hivesim is a Go wrapper for the Hive Simulation API.
```
holiman, line 2: We might, in the README, point to godocs for additional package docs. Just to make it easier for people to find the package-specific documentation, which until now, maybe we've been kind of bad at maintaining

holiman (approved):
Good job @renaynay !

hivesim/doc.go
renaynay: @holiman I added a `godocs` button to the README.md, do you think that's sufficient?

PR #404, opened 2020-11-10
clients/besu: add AOT compilation of besu jars

fjl:
This reduces the startup overhead of the JVM. In my tests, besu starts
quite a bit faster with the AOT libraries enabled: it takes 9s without,
5s with AOT. The downside to this is that the precompilation takes a lot
of time.

### DIFF
--- a/clients/besu/Dockerfile
+++ b/clients/besu/Dockerfile
@@ -2,10 +2,14 @@ ARG branch=develop
 FROM hyperledger/besu:$branch
 
 USER root
-RUN apt-get update && apt-get install -y curl
+RUN apt-get update && apt-get install -y curl binutils
 RUN curl -L -o /usr/local/bin/jq 'https://github.com/stedolan/jq/releases/download/jq-1.6/jq-linux64' && \
     echo 'af986793a515d500ab2d35f8d2aecd656e764504b789b66d7e1a0b727a124c44  /usr/local/bin/jq' | sha256sum -c && \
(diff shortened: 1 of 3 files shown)


fjl:
@davemec, pinging you here since you seem to be handling hive things for besu.
Please check if this makes any sense.

In general, it would be very nice to have a combined way of importing the blocks
and starting the client for besu. At the moment, running all 22k tests takes more
than a full day with besu because we start the client twice for each test: once to
load the blockchain into the database, then once more to start the RPC server.
If we had a command to 'import then start', we could further reduce the overhead
here.

holiman:
Might be good to ping @shemnon too

fjl:
@davemec told me that besu does actually have a way to import-then-start, it's the `--run` option for the import command.
In #405, I've made the necessary changes to enable it for hive. Using the `--run` option, we can achieve a speedup of 9s -> 5s.

I have also tested combining the `--run` change with this PR. With the AOT compilation enabled as well, we can achieve a startup time of ~3s.

shemnon:
Starting the JVM for each tests is indeed expensive.  

We run in ~7-8min for non-CPU expsneive tests for all reference tests if run in a single JVM. (https://app.circleci.com/pipelines/github/hyperledger/besu/6099/workflows/0cbf327a-6908-4f8d-bb3b-e8a469616047/jobs/29494)

However, Hive has a different model where the stand alone client has the blocks loaded up and then the transaction executed.  Much more black box than our standard tests.

Issue #407, opened 2020-11-23
graphql: update client definitions with gas price config


renaynay:
Update client definitions with gas price config to fix the inconsistency between Besu and Geth below: 

```
Name
07_eth_gasPrice (go-ethereum)

Description
Test case source: https://github.com/ethereum/hive/blob/master/simulators/ethereum/graphql/testcases/07_eth_gasPrice.json

Details

Test failed.
HTTP response code: 200 OK
expected: {
  "data": {
    "gasPrice": "0x10"
  }
}
got: {
  "data": {
    "gasPrice": "0x1"
  }
}
```

fjl:
I have fixed this in #418. All clients now use 16 gwei gas price. Not sure if this requires any changes in graphql suite.

renaynay:
Oh yes this issue can be closed.

PR #413, opened 2020-12-10
simulators/ethereum/graphql: allow multiple valid expected responses for tests

renaynay:
This PR expands graphql tests to allow more than one valid expected response value.


simulators/ethereum/graphql/graphql.go
```diff
@@ -156,24 +156,41 @@ func (tc *testCase) run(t *hivesim.T, c *hivesim.Client) {
...
+				}
+				t.Log("expected value(s):")
+				for _, expected := range tc.gqlTest.Responses {
+					prettyExpected, _ := json.MarshalIndent(expected, "", "  ")
+					t.Log(string(prettyExpected), "\n_____________________\n")
+				}
+
+				t.Log("got:", string(prettyResponse))
+				t.Fail()
+			}()
```
holiman, line 190: I think this would be simpler if you first did the for-loop, instead of checking the index and adding a deferred function. Something like this:

```suggestion
	for i, response := range tc.gqlTest.Responses {
		if reflect.DeepEqual(response, got) {
			return nil
		}
	}
	prettyQuery, ok := reindentJSON(tc.gqlTest.Request)
	prettyResponse, _ := json.MarshalIndent(got, "", "  ")
	
	t.Log("Test failed.")
	t.Log("HTTP response code:", respStatus)
	if ok {
		t.Log("query:", prettyQuery)
	}
	t.Log("expected value(s):")
	for _, expected := range tc.gqlTest.Responses {
		prettyExpected, _ := json.MarshalIndent(expected, "", "  ")
		t.Log(string(prettyExpected), "\n_____________________\n")
	}
	t.Log("got:", string(prettyResponse))
	t.Fail()
```

renaynay:
Will close #407

PR #417, opened 2021-01-08
cmd/hiveview: fix typo

rjl493456442:


fjl:
This is actually not a typo, the file really has this extension. The JS app loads it here: https://github.com/ethereum/hive/blob/master/cmd/hiveview/assets/app.js#L316

PR #421, opened 2021-01-19
hive, internal/libhive: refactoring of internals in preparation for track file support

fjl:
This PR is a large refactoring of hive internals. It is preparation work for the
introduction of the 'tracks' feature, which will allow defining lists of simulation runs
in YAML files.

User-facing changes in the hive command:

- Hive has a new flag `-docker.pull`, which instructs it to pull base images
  when building docker images of clients and simulators.
- The new `-docker.output` flag enables printing of docker output during image builds
  and container execution. Docker output was previously enabled at log level > 3,
  which meant one couldn't see hive's debug logs without also getting
  lots and lots of docker output.
- The deprecated flags `-debug`, `-override`, `-docker-noshell`, `-docker-hostalias`,
  `-sim.rootcontext`, `-hivemaxtestcount` have been removed. Most of them didn't work,
  except for `-hivemaxtestcount`, but the way it worked was really sketchy and it's better
  to use `-sim.testlimit` in its place.
- Hive now stops and removes all of its containers on interrupt.
- Simulations now run in unprivileged containers, just like clients.

There are many changes in the underlying library:

- The internal dependency structure has been cleaned up. I already moved most of the logic
  into internal/libhive in #366, but the way I did it was super messy and didn't really
  achieve a clean separation between the docker backend and all of the high-level code.
  This is now much improved. The docker interfaces are now defined in internal/libhive
  and implemented in a new library package internal/libdocker. As a side effect of the
  much cleaner separation, the 'hivesim' package used by Go simulators no longer depends
  on the docker client. Yay!
- The image building code is now available in library form as part of libdocker. It used
  to be in hive's package main and was very entangled with the handling of the `-sim` and
  `-client` flags.
- Hive now uses the ContainerBackend of libdocker to manage the simulation container. This
  removes the duplicated copy of the container launch code that I left behind in #366. We
  were already using the backend for client containers and I had to make a lot of API
  changes to get this to work, but it really improved the API as well.
- Container launching and image builds now support `Context` for cancelation of
  long-running operations such as the client TCP port 8545 ready check.
- The container backend had a couple of potential resource leaks (i.e. log file handles,
  background goroutines). These should now be fixed.


internal/libdocker/container.go
```diff
@@ -0,0 +1,442 @@
...
+		}
+		log, err := os.OpenFile(logfile, os.O_WRONLY|os.O_CREATE|os.O_SYNC|os.O_TRUNC, 0644)
+		if err != nil {
+			return nil, err
+		}
+		closer.addFile(log)
+		stream = log
+
+		// If console logging was requested, tee the output and tag it with the container id.
+		if b.config.ContainerOutput != nil {
```
fjl, line 328: This is slightly incorrect, ContainerOutput should be independent of logfile.

PR #424, opened 2021-01-26
simulators/ethereum/sync: add test for light client sync

rjl493456442:
This PR adds the test suite for les syncing.

It's still problematic

```
[76276231] DEBUG[01-26|05:37:16.160] invalid neighbour (172.17.0.2) from 0871040109a99f25@172.17.0.9:30303: is self
[76276231] DEBUG[01-26|05:37:16.261] Bad discv4 packet                        addr=172.17.0.9:30303 err="bad hash"
[76276231] DEBUG[01-26|05:37:16.261] invalid neighbour (172.17.0.2) from 0871040109a99f25@172.17.0.9:30303: is self
[75d04963] DEBUG[01-26|05:37:16.331] Served eth_getBlockByNumber              conn=172.17.0.7:60820 reqid=3 t="132.145µs"
[76276231] DEBUG[01-26|05:37:16.362] Bad discv4 packet                        addr=172.17.0.9:30303 err="bad hash"
[76276231] DEBUG[01-26|05:37:16.362] invalid neighbour (172.17.0.2) from 0871040109a99f25@172.17.0.9:30303: is self
[76276231] DEBUG[01-26|05:37:16.462] Bad discv4 packet                        addr=172.17.0.9:30303 err="bad hash"
[76276231] DEBUG[01-26|05:37:16.462] invalid neighbour (172.17.0.2) from 0871040109a99f25@172.17.0.9:30303: is self
[76276231] DEBUG[01-26|05:37:16.563] Bad discv4 packet                        addr=172.17.0.9:30303 err="bad hash"
[76276231] DEBUG[01-26|05:37:16.563] invalid neighbour (172.17.0.2) from 0871040109a99f25@172.17.0.9:30303: is self
[76276231] DEBUG[01-26|05:37:16.664] Bad discv4 packet                        addr=172.17.0.9:30303 err="bad hash"
[76276231] DEBUG[01-26|05:37:16.664] invalid neighbour (172.17.0.2) from 0871040109a99f25@172.17.0.9:30303: is self
[76276231] DEBUG[01-26|05:37:16.764] Bad discv4 packet                        addr=172.17.0.9:30303 err="bad hash"
[76276231] DEBUG[01-26|05:37:16.765] invalid neighbour (172.17.0.2) from 0871040109a99f25@172.17.0.9:30303: is self
```

Do you have any idea? @fjl


rjl493456442:
I think I find the issue. The les server will reject all the light connections if it's not synced yet.

However in this test suite, no more extra blocks will be imported, so the status is always "non-synced".

One feasible solution is the enable serving no matter the server is synced or not by an extra CLI flag.

fjl:
Since this test is basically a copy of the normal 'sync' test, we could just integrate it into the 'sync' test as a special test case.

rjl493456442:
@fjl Thanks for reviewing it. I do have a question want to ask. The original scheme for testing is: run the test spec against all the specified clients. While to les-related tests, it's a little special, we need a server-client structure. I am wondering what's the best way to build the structure with some specified "servers" and "clients".

But the situation is: LES is only implemented by Geth, so it's only useful to test the syncing against different client version.

rjl493456442:
@fjl Regarding integration. Do you think it's acceptable that I add some additional checks(e.g. the sync source and sync destination should at least be Geth client because LES is only implemented by Geth client).

Or we can leave the `lessync` there to keep the original `sync` simulator clean.

simulators/ethereum/sync/sync.go
```diff
@@ -73,18 +88,68 @@ func runSourceTest(t *hivesim.T, c *hivesim.Client) {
 		Description: fmt.Sprintf("This test attempts to sync the chain from a %s node.", source.Type),
 		Parameters:  sinkParams,
 		Files:       sinkFiles,
-		Run:         runSyncTest,
+		Run:         runETHSyncTest,
 	})
 }
 
-func runSyncTest(t *hivesim.T, c *hivesim.Client) {
+func runETHSyncTest(t *hivesim.T, c *hivesim.Client) {
```
fjl, line 95: Please don't rename this.

simulators/ethereum/sync/sync.go
```diff
@@ -73,18 +88,68 @@ func runSourceTest(t *hivesim.T, c *hivesim.Client) {
...
 	err := node.checkSync(t, testchainHeadNumber, testchainHeadHash)
 	if err != nil {
 		t.Fatal("sync failed:", err)
 	}
 }
 
+func runLESSourceTest(t *hivesim.T, c *hivesim.Client) {
+	// Short circuit if the source is not Geth client.
+	if !isGeth(c.Type) {
+		return
```
fjl, line 106: I think this should be done a bit differently. If you use `ClientTestSpec`, it will launch an instance of the test for all client types. But this test will only work with geth. What you should to instead is:

```
sim := hivesim.New()
clientTypes, err := sim.ClientTypes()
if err != nil {
    panic(err)
}

suite := hivesim.Suite{ ... }
suite.AddTest(hivesim.ClientTestSpec{ ... eth test ... })

// Add the light client sync test. This only works with geth for now.
var gethClients []string
for _, name := range clientTypes {
    if strings.HasPrefix(name, "go-ethereum") {
        gethClients = append(gethClients, name)
    }
}
for _, name := range gethClients {
    name := name
    suite.AddTest(hivesim.TestSpec{
         Name: fmt.Sprintf("%s as LES server", c),
         Run: func (t *T) { runLESTest(t, name) },
    })
}
```

fjl: Then, in `runLESTest`, you can use `t.RunClient` with all geth client types instead of using `t.RunAllClients`.

fjl: Sorry, this is a bit complicated right now. We can make this simpler in a future version of package hivesim.

rjl493456442:
@fjl Please take another look. 

If you want to test, please repalce the base Geth image with `rjl493456442/go-ethereum`, the `light.nosyncserve` is necessary for testing

clients/go-ethereum/geth.sh
```diff
@@ -137,6 +138,11 @@ if [ "$HIVE_MINER_EXTRA" != "" ]; then
 fi
 FLAGS="$FLAGS --miner.gasprice 16000000000"
 
+# Configure any les flags
+if [ "$HIVE_LIGHTSERVE" != "" ]; then
+  FLAGS="$FLAGS --light.serve $HIVE_LIGHTSERVE --light.nosyncserve "
```
rjl493456442, line 143: The `light.nosyncserve` flag is only available after 1.9.25. We should ban all geth client less than 1.9.25 to run as the les server.

simulators/ethereum/sync/sync.go
```diff
@@ -43,17 +43,46 @@ func main() {
...
+		sim     = hivesim.New()
+	)
+	clientTypes, err := sim.ClientTypes()
+	if err != nil {
+		panic(err)
+	}
+	for _, name := range clientTypes {
+		if !strings.HasPrefix(name, "go-ethereum") {
+			continue
+		}
```
rjl493456442, line 71: Would make sense to filter out the geth client less than 1.9.25

fjl: We can't really do that here because the client version is not available to the simulator.

fjl:
I need to think about these changes some more. It will be merged, but not right now.

fjl:
I have decided to not integrate LES into the sync test because the rpc test for LES already covers the same thing.

PR #425, opened 2021-01-27
simulators/ethereum/graphql/testcases: remove test cases with queries on account

renaynay:
This PR removes tests with queries on account as the [graphql spec](https://eips.ethereum.org/EIPS/eip-1767) does not include queries on account.


renaynay:
Closing in favor of #426

PR #426, opened 2021-01-28
simulators/ethereum/graphql: change account queries to be on block, not account

renaynay:
This PR changes several testcases from querying on account to querying on block for account information. Queries on account are not required by the [graphql spec](https://eips.ethereum.org/EIPS/eip-1767). While Besu implements queries on account, geth does not as it is not required.


fjl:
I don't fully understand which queries are unsupported. Do you mean `account { ... }` is not supported, but `block { account { ... } }` is supported?

The JSON-RPC compatibility table in EIP-1767 contains this row:

```

renaynay:
It was never required by the graphql spec though and we don't have it implemented. Should we implement support for queries on account then in geth?

fjl:
We talked about this in a call and decided couple things:

- Basic change to add `block { }` around account query in tests is OK and besu guys also agreed with this change
- Let's not add the multiple-status-code feature to tests. It will cause all tests to pass with all clients even though the clients disagree on basic things.
- Account does not exist exception should be removed in besu

renaynay:
Account exception was removed in Besu and status code returned is now 200, after fixing merge conflicts, can we merge this?

holiman:
Oh, did I merge them in wrong order? Now there are conflicts (oops)

renaynay:
Yes, but it's okay :) I can fix them, thanks @holiman

PR #429, opened 2021-01-31
simulators/devp2p: update discv4 and eth test dependencies

renaynay:
This PR updates discv4 and eth test dependencies


fjl:
This update didn't go so well, I think. If you want to update the dependencies of a simulator, then the main go.mod should not be modified at the same time, and vice versa. What we should do is upgrading go-ethereum to v1.9.25 in the main hive module, then commit + merge that. Then, in a subsequent PR, we can pull this new commit of hive into simulators.

BTW, if you are doing this upgrade to just include the API change you made in hivesim, you don't need to pull that in immediately. Since it's a bit of a hassle to do those updates, we can defer them until later.

fjl:
I will update the simulators later when more API changes are in hive base.

PR #430, opened 2021-02-02
simulators/eth2: *experimental*, eth2 testnets and p2p tests [WIP]

### LINKED ISSUE #428: eth2 tests integration
This issue is about integrating eth2 clients into hive.

hive changes:
  - think about port 8545 check
    - docker health check?
    - provide lower level non-blocking client startup in sim api
      - make the simulator poll
    - just make it work in eth2 client images with this port
  - enode.sh is eth1 specific
    - no need to worry about it because eth2 has a standard API for this
    - there should be a way to get the multiaddr in eth2
  - **need way to access files of running container or run script in container**
    - i would prefer script approach, in the following way
      - client has defined directory `/hive-bin/` containing scripts
      - sim API will only launch scripts in this directory

eth2 specifics:
  - eth2 has two roles:
    - beacon node
    - validator node
  - we need to package both, e.g. `clients/eth2-validator/prysm`, `clients/eth2-beacon/teku`
  - [eth2 rest API](https://github.com/ethereum/eth2.0-APIs) needs to be started by client
  - want to test validator key move between client instances, because its complicated w.r.t slashing
    
env:
  - eth2 spec configurables
  - bootnodes
  - node identity (maybe make this optional)

files:
  - genesis file (ssz)
    - deposits
  - validator keys (json)

protolambda:
This is DRAFT, do not merge it yet.

Eth2 is not a simple system, and some additions to Hive are necessary.

If new Hive features look good I can refactor them into a new PR and get them merged. But generally new Hive features will not be 
stable until the eth2 simulator runs.

To name a few differences:
- "client types" consist of a Beacon Node container, a Validator Client container. And they can be interchanged, or either of the two can be developed. Both are required to run a testnet.
- Eth2 uses a HTTP API, not an Eth1 RPC. We do have [API specs](https://github.com/ethereum/eth2.0-APIs) and [bindings](https://github.com/protolambda/eth2api).
- The VC needs to learn the BN API address to do its work.
- No enode URL format. Only ENR and multi-addrs. Generalizing the `enode.sh` to run `/hive-bin/<name>` via an API like `/some/hive/endpoint/<container>/<name>` would help a lot.
- The above is also important for advanced tests like keystore-transfer with slashing-protection, some script access to a running container is super powerful.
- Eth2 has its own configuration standard, mostly standardized between clients, but some of it is compile-time only and some require additional configuration. Using the same env vars trick as eth1-hive sims, each client can write their own config.
- Node startup is different, API not directly up in pre-genesis case
- An Eth1 RPC for deposits is another node type to add (integration tests), or can be mocked (in general tests)
- State can be imported as SSZ file, different client-prep in general.
- Keystores need to be imported into clients, with strict permission settings

TODO:
- [x] Improve client-start with TAR support:
   - Option pattern for forwards-compability, maybe we add more options later.
   - TAR is a special kind of file (uses MIME header to define it as such), works just like existing files, but the TAR is unpacked in the multi-part -> TAR step in the docker glue code.
   - When dealing with thousands of keystores, permission settings, large state dynamically-generated state files, etc. it's nice to just have TAR support like Docker itself.
- [x] Port over genesis-state-generation from previous Hivemind experiment to Hive eth2 simulator
- [x] Port over keystore-generation from previous Hivemind experiment to Hive eth2 simulator
- [ ] Hook up Eth2 API bindings in Eth2 client type for easy API calls during tests
- [x] Define config vars spec (hive env names, the runtime subset of eth2 config vars)
- [x] Define Eth2 client docker containers that start BNs and VCs with given state, keystores and env vars.
- [ ] Implement testnet simulator that checks validator performance (i.e. on-time finalization right after genesis), with mocked eth1
- [ ] Implement Eth2 sync tests
- [x] Add client-metadata, and role-filtering, so not all Eth1 simulators try to run tests with the eth2 clients, and we can compose different tests with clients of different roles (e.g. an eth2 tests with an eth1 endpoint, beacon node, and validator client)

Advanced ideas (probably later PR)
- Testnet with faster slot time
- Validator keystore transfer
- Add Slasher node type
- Add malicious validators to test attacks.
- Add Eth1 node for deposit chain integration tests
- Eth1 merge testing

This fixes #428


fjl:
I don't understand the need for the special handling of tar archives. You can already upload an arbitrary number of files using the existing API, and sending them as a single file doesn't really make a difference. If you want to be able to upload files in hivesim without writing them to disk first, this is something we can implement, but it can be done without changing the HTTP API.

protolambda:
> You can already upload an arbitrary number of files using the existing API

But lacking any file metadata (it's nice to just permissions on validator keystores, and not have to script that in N different places).

> sending them as a single file doesn't really make a difference

For the convenience of sending 1000s of tiny keystore and secret files, I disagree. Maybe not in performance, but it's more practical to just write a tar and pass around some bytes, than thinking about multi-part forms.

> this is something we can implement, but it can be done without changing the HTTP API.

That's what I did. API works pretty much the same. Docker uses tar, I don't see why we shouldn't if we want (optionally in-memory) archives of files via API.

And sure, we can polish it (fix reader close, fix tests that were matching error string contents, etc.).

fjl:
I'll just wait for you to finish up your draft and then complain later. IMHO the tar archive stuff is not needed and is just extra complexity. The only new thing this provides is control over the file permissions. I really hope we won't need to ever think about file permissions in hive tests.

protolambda:
Pain with multi-module repo: the eth1 simulators have module files that reference upstream hive, instead of local module. Adding replace directives for now to work around that. Once it's ready we can squash commits and get rid of the temporary replace directives.

protolambda:
Refactored out some components, based on call discussion, to gradually update hive before introducing an Eth2 simulator.

See:
- #434 run scripts in containers (updated)
- #442 StartOption, bundling of file options (without TAR)
- #443 Client roles / metadata

Other things:
- Build targets (maybe)
- Eth2 API client (like Eth1 RPC is already exposed). Maybe better to introduce in simulator first, and share between simulators later, when merge-sim is introduced.

protolambda:
The more experimental eth2 simulator changes are currently on https://github.com/protolambda/hive/tree/build-targets
Next wek I'll rebase this on top of a hypothetical merge of the other PRs, then continue eth2 sim work.

PR #432, opened 2021-02-02
simulators: integrate les light client into the RPC test suite

rjl493456442:
It's a following PR based on #424


holiman (approved):
LGTM

holiman:
> It's a following PR based on #424

Does this require that other PR or can this be merged separately?

rjl493456442:
@holiman The problem of the previous PR(#424) is that: 

LES Server has this property that it will reject all client connections by default **if it's not synced yet**. 
While the indicator of "node synced" is setup by (1) inserting a new block (2) start mining.

For (1) it's impossible/ugly in the hive test suite, we insert the pre-configured blocks by offline command and no more block involved. So the indicator won't be marked forever.

For (2) it's feasible to start mining with `thread=0`. It won't waste CPU resource but can mark the node as "synced". But it's not elegant enough?

So I add a new [feature](https://github.com/ethereum/go-ethereum/pull/22250) in Geth, so that LES server can start serving even the les server is not synced. It can work perfectly with new version Geth, but it won't work for the legacy Geths. So it blocks the PR.

We can either filter out the legacy Geth version for LES testing or use approach (2) so that it can work for all Geths. What's your option about it? 

cc @fjl

fjl:
I think we need to wait a couple more days before merging this. I'm working on integrating the client roles from PR #443 at the moment. When this is done, we can remove the client name logic in this PR and just define roles for les-capable clients.

fjl:
Finally added this in https://github.com/ethereum/hive/commit/1f7f6b4b195b7b4b28a5d60bb0fc2cad635c0302 and https://github.com/ethereum/hive/commit/9e744e79d62b2cd5a9fbae1895c7fbf40eb8aa25.

PR #434, opened 2021-02-02
internal/libhive, hivesim: add client script exec API

protolambda:
Changes:
- ClientRunProgram in hivesim can run a command in a container
- Hive API endpoint to run program in node
- Test backend implements basic mock + hook
- Test to see API work

This is a first attempt to generalize the "enode.sh" functionality (idea from earlier discussion with @fjl).
It misses the scope limit to `/hive-bin/<program>` still, I didn't find a clean way to limit the command (especially if it's shell code).

TODO The API is tested, but the docker backend implementation is not.

This helps:
- Run more advanced eth2 tests, e.g. stop validator, export keystores and slashing DB out of a container, then start again. (and import it into another validator). E.g. `/hive-bin/export-keys.sh`
- Connect eth2 clients. Eth2 nodes don't have an enode-url formatted address, and the beacon API address also needs to be communicated. We can put a `/hive-bin/multiaddr.sh`, `/hive-bin/`

### DIFF
--- a/hivesim/hive.go
+++ b/hivesim/hive.go
@@ -196,14 +196,18 @@ func (sim *Simulation) ClientEnodeURL(testSuite SuiteID, test TestID, node strin
 }
 
 // ClientExec runs a command in a running client.
-func (sim *Simulation) ClientExec(testSuite SuiteID, test TestID, nodeid string, cmd string) (*ExecInfo, error) {
-	params := url.Values{}
-	params.Add("cmd", cmd)
-	p := fmt.Sprintf("%s/testsuite/%d/test/%d/node/%s/exec?%s", sim.url, testSuite, test, nodeid, params.Encode())
-	req, err := http.NewRequest(http.MethodPost, p, nil)
+func (sim *Simulation) ClientExec(testSuite SuiteID, test TestID, nodeid string, cmd []string) (*ExecInfo, error) {
+	type execRequest struct {
+		Command []string `json:"command"`
+	}
+	enc, _ := json.Marshal(&execRequest{cmd})
+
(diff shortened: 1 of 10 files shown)


hivesim/hive.go
```diff
@@ -154,6 +155,40 @@ func (sim *Simulation) ClientEnodeURL(testSuite SuiteID, test TestID, node strin
...
+	StdErr   string `json:"err"`
+	ExitCode int    `json:"code"`
+}
+
+// ClientRunProgram runs a command in a running client.
+func (sim *Simulation) ClientRunProgram(testSuite SuiteID, test TestID,
+	nodeid string, privileged bool, user string, cmd string) (stdOut string, stdErr string, exitCode int, err error) {
+
+	params := url.Values{}
+	params.Add("privileged", strconv.FormatBool(privileged))
```
fjl, line 169: I don't want to add support for privileged exec because I just changed all container execution to unprivileged.
Also think it's not needed to specify the user. The API here should really just be

```
func (sim *Simulation) ClientExec(s SuiteID, t TestID, node string, cmd string, ...args string) (int, string, error)
```

protolambda:
Rebased and simplified

hivesim/hive.go
protolambda: Still like having std-err and std-out separate. Combining them often becomes a mess, and I've spend more time debugging logs of just std-out or std-err, hiding the data I need, than I'd like. For a simple enode-like script it's not necessary, but when catching a panic after a complex keystore-export operation, it's great.

Removed the user/privilege options, and bundled the outputs.

PR #438, opened 2021-02-05
hive: don't abort entire run for client build error

fjl:
This changes the client build to behave like the documentation says:
it's OK for some client builds to fail, and they'll just be excluded from the run.
However, at least one client must build successfully.


renaynay (approved):
LGTM but there's a merge conflict

PR #443, opened 2021-02-12
internal/libhive: client definitions, metadata file loading, client-roles

protolambda:
Changes:
- Load clients into single collection of definitions, don't separate versions, images, metadata into different maps.
- List the definitions in the API clients listing, except docker image: metadata such as the role is important for test-suites, docker-strings not so much.
- Update client loading, no changes required to clients. Default metadata is an "eth1" role. 

Eth1 simulators should update like:
```diff
	suite.Add(hivesim.ClientTestSpec{
+               Role: "eth1',
		Run: runDiscoveryTest,
		Parameters: hivesim.Params{
			"HIVE_LOGLEVEL": "5",
		},
	})
```
To ensure the test only ever runs against the intended kind of client. If the test runner does not provide eth2 client names by accident, the old behavior still works.

Next, we can make other tests that consume "beacon", "validator", "shard-node", etc. roles.

For an Eth2 testnet, or Eth1-Eth2 merge test, multiple roles may be involved. The CLI does not change, just give it a list of involved clients. The simulator can e.g. find the first beacon node, first validator, first shard-node to run a simple testnet. Or parametrize by role, to try different combinations (e.g. for Eth2 validator <> beacon API tests, and beacon <> eth1 RPC tests).

TODO:
- Update docs to describe `hive.yaml` in clients.
- More testing. Chicken-egg problem when using it in a simulator and having it in hive.
- Suite files only list the client versions. I kept that as-is, since changing it breaks viewing of existing suite-outputs, and the recent versions support broke the old format already (but gracefully). We have full client info now though, maybe worth listing it in the suite output?
- Thinking of best build-target approach. The metadata file is useful to define builds of the same client and version, but different compile-time arguments (e.g. eth2 minimal/mainnet configuration choice). For another later PR.

### DIFF
--- a/docs/clients.md
+++ b/docs/clients.md
@@ -41,11 +41,27 @@ configurable in certain ways. In order to run tests against multiple Ethereum cl
 example, the simulator needs to be able to configure all clients for a specific blockchain
 and make them join the peer-to-peer network used for testing.
 
+## Client Metadata
+
+Metadata is used to express client differences. E.g. clients can have different roles
+within the Ethereum ecosystem, each of which can have different tests,
+some even capturing interactions between multiple roles.
+
+Client metadata is defined with a `hive.yaml` file in the client directory (next to the 
+`Dockerfile`). This optional, by default each client is assumed to only have an `eth1` role.
+
+The YAML fields are:
+```yaml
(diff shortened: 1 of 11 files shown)


fjl:
I think we should support multiple roles per client.

PR #458, opened 2021-03-11
simulators: decrease eth simulator loglevel a bit

holiman:
The lastest besu logfile from the eth protoocol is `59Mb`. Let's decrease that a bit

### DIFF
--- a/simulators/devp2p/eth/eth.go
+++ b/simulators/devp2p/eth/eth.go
@@ -26,7 +26,7 @@ Results from the test tool are reported as individual sub-tests.`,
 			"HIVE_FORK_TANGERINE": "0",
 			"HIVE_FORK_SPURIOUS":  "0",
 			"HIVE_FORK_BYZANTIUM": "0",
-			"HIVE_LOGLEVEL":       "5",
+			"HIVE_LOGLEVEL":       "4",
 		},
 		Files: map[string]string{
 			"genesis.json": "./init/genesis.json",


holiman:
Hm, this might not actually make a _lot_ of difference. The log file contains a lot of 
```
 PeerInfo{version=5, clientId='', capabilities=[eth/64, eth/65, eth/66], port=0, nodeId=0x4b113616f7a07ba3d63b722e1e7d6e8555aa16359856ed9d50d832b8dcca9f9dc29577f18010470c31b09ebdfed726960c03b425b5feb025dba5aaf22aeead76a
3a38118dba3b07be61749637972bd8e97f629aeb150565b199ad371c24f61e3217119910475abd6d4477d6be6a9a48d9b05985120218b29676778ff722b8da535872bd8f101c012fdc81ba6c4112b2cbc806e48118594056ee77de1f9...
```
( a couple of MB of `nodeId`, and other messages where they spit out the `capabilities` which are stuffed with data from the malicious tests. So @RatanRSur you might want to add some sanity checks for all arbitrary-size integers to ensure they make sense.

holiman:
Another example: 

`
2021-03-09 22:10:29.353+00:00 | nioEventLoopGroup-3-7 | DEBUG | DeFramer | Received HELLO message: PeerInfo{version=5, clientId='', capabilities=[0xbcb1768aec9508efcb3c17b92950c7d66ff81e73e1efc3f6a21dc10557a21c05f52dbe58fd8406efb12a033cdfb5b50d133962277ea401a824bd793f04fb1ad8530386d0481668ba7def9263d37b525b5e382f323
...
`

atoulme:
Do you really need to log at debug level?

atoulme:
Another idea - you can configure logging per package in java. You can shut off specific packages if they're too noisy.

PR #461, opened 2021-03-26
docs: improve documentation

tomichec:
Describes the initial system setup in more detail.
Corrects the documentation about the simulators so that the tests pass now.

### DIFF
--- a/docs/simulators.md
+++ b/docs/simulators.md
@@ -41,7 +41,7 @@ suite.
 
 While simulators may be written in any language (they're just docker containers after
 all), hive provides a Go library that wraps the simulation API in a way that resembles the
-standard library "testing" package. Be sure to check the [Go API reference of package
+standard library "testing" package. Be sure to check the Go API reference of [package
 hivesim] for more information about writing simulators in Go.
 
 Simulators are contained in the hive repository as independent Go modules. To create one,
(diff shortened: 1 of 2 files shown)


docs/simulators.md
```diff
@@ -41,50 +41,107 @@ suite.
...
+[launch the simulation]: [./overview.md#running-hive]
+
+### Checking the results
+
+The test results are going to be written to the standard output and
+you can also review the results [on your server in case you run the
+`hiveview`][hiveview]. If you set everythin as described, your test
+should pass and you have your testing environment set up correctly.
+
+[hiveview]: ./commandline.md#viewing-simulation-results-hiveview
```
fjl, line 144: I think a lot of this content you added here overlaps with the other documents.

tomichec: Well, you might be right, but I've read the whole documentation in this folder and many thinks I've added there were not mentioned yet. I've tried to formulate some of the overlapping information more clearly. Also, as I've got through the initial setup following the documentation, and the tests failed. So, I've completed the code including a dummy functions for instance, so that that allow the tests to pass. Finally, I've changed the name of the files from `my-test` to `my-simulation` as @protolambda suggested that the name `test` is reserved for other things.

PR #462, opened 2021-03-26
cmd/hivechain: use GetBalance() method for tx generation

renaynay:
This PR is dependent on [this PR to add a `GetBalance()` method to block generation in geth](https://github.com/ethereum/go-ethereum/pull/22589), and now allows the hivechain tool to generate chains with a given number of transactions per block (capped at 2000 **@TODO this is arbitrary, should we increase the cap?**) instead of just 1.


cmd/hivechain/generate.go
```diff
@@ -86,23 +87,22 @@ func (cfg generatorConfig) addTxForKnownAccounts(i int, gen *core.BlockGen) {
...
+				continue
+			}
+			log.Printf("TX FROM ACCOUNT: %v", addr)
 			txCost := big.NewInt(1)
-			spent := new(big.Int).Mul(txCost, big.NewInt(int64(i*cfg.txInterval)))
-			gbal := new(big.Int).Set(a.Balance)
-			if gbal.Sub(gbal, spent).Cmp(txCost) < 0 {
-				continue // no funds left in this account
+			if gen.GetBalance(addr).Cmp(txCost) > 0 {
+				tx := generateTx(txType, key, &cfg.genesis, gen)
```
fjl, line 100: You can do it better by always creating the tx, then comparing against `tx.Cost()`

go.mod
```diff
@@ -27,13 +28,13 @@ require (
...
 	golang.org/x/net v0.0.0-20201026091529-146b70c837a4 // indirect
 	golang.org/x/sync v0.0.0-20201020160332-67f06af15bc9 // indirect
-	golang.org/x/sys v0.0.0-20201109165425-215b40eba54c // indirect
 	google.golang.org/genproto v0.0.0-20201109203340-2640f1f9cdfb // indirect
 	google.golang.org/grpc v1.33.2 // indirect
 	gopkg.in/inconshreveable/log15.v2 v2.0.0-20200109203555-b30bc20e4fd1
 	gopkg.in/yaml.v2 v2.3.0
 )
+
+replace github.com/ethereum/go-ethereum => github.com/renaynay/go-ethereum v1.9.12-0.20210326162836-ac99fed7c3f0
```
fjl, line 40: You can remove this now

PR #446, opened 2021-02-25
simulators/eth2: Run short eth2 testnets in Hive

protolambda:
The idea:
1. Improve Hive to handle Eth2 clients (via separate PRs, this PR builds on top of them):
    - [ ] internal/libhive: client definitions, metadata file loading, client-roles. PR: #443
    - [ ] hivesim: StartOption for client configuration. PR: #442
    - [ ] hivesim: generalized enode.sh: new ClientRunProgram to exec cmds in containers. PR: #434
    - [ ] multiple build targets per client (WIP)
2. Add Hive simulator to run short Eth2 testnet
    - Make sure Eth2 support works. Others can then build more Hive simulators to test more Eth2 features.
    - Enable integration testing with Eth1 clients: for monitoring deposits in phase0, but then Eth1-Eth2 merge later!

This replaces #430, and builds the Eth2 simulator on top the separate Hive PRs. The PR diff will get smaller as those features are merged. This PR stays a Draft until the other PRs land and the testnet functionality works.

PR #481, opened 2021-09-15
Eth2 Hive - take 2

protolambda:
Replaces the original eth2 draft #446

This new simulator:
1. Takes list of available client definitions, organizes by role
2. Checks if it has 1 of each role (eth1/beacon/validator). We could mix more later
3. Prepares a testnet with `2**14` (= approx. 16k) validators, split between 4 beacon nodes, each with an eth1 node and validator client. Fork version and other runtime configuration is customized, but otherwise mainnet-like.
    - If clients could publish `minimal` (faster preset used in consensus unit-testing) docker builds, and if we introduce build-target support in Hive, then we could benefit from the quicker test results of `minimal` (less slots per epoch, shorter slots, shorter eth1 deposit-data voting period, more interesting behavior)
4. Creates a testnet, basically a collection of client instances that is built incrementally (simple loop that adds 1 of each type here, but could reference any previous instance for bootnode/api/etc. coupling)
5. Monitors the testnet for 10 minutes, each slot, and tracks the head-block and finality of every beacon node.
    - Go API client bindings for full consensus API available, can monitor more data, if anyone has good ideas. Maybe track eth1 voting of each block against the actual eth1 chain?

Also, one minor change in Hive itself: use the `HIVE_CHECK_LIVE_PORT` client-param to change/disable the live-check of started client containers. This is optional, and it still defaults to the old hardcoded 8545 value. This way we can adjust it to the ports that beacon/validator clients use.

Lighthouse beacon/validator client supported so far. Other consensus clients next (will prioritize those that support the HTTP standard API nicely). Once we can run all consensus clients we can start composing mixed-vendor testnet cases as well. And then Merge testnets.

How to run:
```bash
# build hive binary
go build .

# run simulator
./hive -docker.output --sim.loglevel=3 --loglevel=5 --sim eth2/testnet --client go-ethereum,lighthouse-bn,lighthouse-vc
```


simulators/eth2/README.md
```diff
@@ -0,0 +1,97 @@
...
+# Some clients have a hard limit on slots that can be skipped without block. 
+HIVE_ETH2_MAX_SKIP_SLOTS: 1000
+
+# Comma separated list of Eth1 nodes to communicate with.
+# Clients should strip off everything after first comma if they do not load-balancing between them.
+# If it is left empty, the beacon node should use a "dummy eth1" mode, where it fills Eth1 votes with mock data.
+HIVE_ETH2_ETH1_RPC_ADDRS: ""
+
+# Port to expose standard HTTP API on 
+HIVE_ETH2_BN_API_PORT: 4000
```
fjl, line 91: I think you should just require that clients run this API on port 8545.

clients/go-ethereum/mapper.jq
```diff
@@ -28,6 +28,49 @@ def to_bool:
...
+        "0x0000000000000000000000000000000000000000000000000000000000000039": "0x31206fa80a50bb6abe29085058f16212212a60eec8f049fecb92d8c8e0a84bc0",
+        "0x000000000000000000000000000000000000000000000000000000000000003a": "0x21352bfecbeddde993839f614c3dac0a3ee37543f9b412b16199dc158e23b544",
+        "0x000000000000000000000000000000000000000000000000000000000000003b": "0x619e312724bb6d7c3153ed9de791d764a366b389af13c58bf8a8d90481a46765",
+        "0x000000000000000000000000000000000000000000000000000000000000003c": "0x7cdd2986268250628d0c10e385c58c6191e6fbe05191bcc04f133f2cea72c1c4",
+        "0x000000000000000000000000000000000000000000000000000000000000003d": "0x848930bd7ba8cac54661072113fb278869e07bb8587f91392933374d017bcbe1",
+        "0x000000000000000000000000000000000000000000000000000000000000003e": "0x8869ff2c22b28cc10510d9853292803328be4fb0e80495e8bb8d271f5b889636",
+        "0x000000000000000000000000000000000000000000000000000000000000003f": "0xb5fe28e79f1b850f8658246ce9b6a1e7b49fc06db7143e8fe0b4f2b0c5523a5c",
+        "0x0000000000000000000000000000000000000000000000000000000000000040": "0x985e929f70af28d0bdd1a90a808f977f597c7c778c489e98d3bd8910d31ac0f7"
```
fjl, line 67: This is definitely not the right place for this. You should put these storage keys into the input genesis data.

simulators/eth2/README.md
protolambda: They are completely different APIs, that might even run side by side in the same container in the future (if nimbus or consensys merge their eth1/eth2 client projects). No one uses the 8545 port for eth2 APIs and it is very confusing to retrofit it like that.

clients/go-ethereum/mapper.jq
protolambda: Where can I find a good example of this? I thought it would be convenient to just insert a ready deposit contract if the deposit contract address is specified.

fjl: The genesis data is translated into client-specific formats by those mapper.jq scripts. So if you would add it like this, we would have to change all the eth1 client definitions to maintain this account in all the possible formats. It's really the wrong approach.

For eth1 clients, the genesis block is part of the client input data. You can find an example here: https://github.com/ethereum/hive/blob/master/simulators/ethereum/graphql/init/testGenesis.json#L12

If you want this account to be part of the genesis, just generate the input genesis.json to include it. That way, it will be translated for all clients and you only need to maintain this info in one place. If you want to do this with go, you can use the `core.Genesis` type from go-ethereum, it maps to the JSON encoding.

clients/lighthouse-bn/lighthouse_bn.sh
```diff
@@ -0,0 +1,62 @@
...
+    0|1) LOG=error ;;
+    2)   LOG=warn  ;;
+    3)   LOG=info  ;;
+    4)   LOG=debug ;;
+    5)   LOG=trace ;;
+esac
+
+echo "bootnodes: ${HIVE_ETH2_BOOTNODE_ENRS}"
+
+hostname -i
```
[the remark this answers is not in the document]
protolambda: left from debugging, will clean up, thanks

simulators/eth2/README.md
```diff
@@ -0,0 +1,97 @@
+# Eth2 in Hive
```
zilm13, line 1: Shouldn't it be in ` simulators/eth2/testnet/`?

holiman:
@protolambda let's un-draft this?

MariusVanDerWijden (approved):
LGTM

PR #489, opened 2021-11-27
Add Geth Engine (PoS) API Compatible Client + Engine Simulator

marioevz:
This draft PR adds a custom client which is a geth version that contains Engine API compatibility, and allows block production using PoS calls.

The client uses https://github.com/MariusVanDerWijden/go-ethereum/tree/eip-4399

An "engine" simulator is also added, which is basically a copy of ethereum/rpc, but the blocks are produced by using the PoS calls instead of mining.

The TerminalTotalDifficulty is set in order for the client to deprecate mining in favor of PoS engine calls.

The simulator produces PoS blocks asynchronously by running the following calls in sequence:
- engine_forkchoiceUpdatedV1, using the latest block head known, and request a payload with `last block time + 1`
- engine_getPayloadV1, to get the generated payload
- engine_executePayloadV1, to execute the generated payload
- engine_forkchoiceUpdatedV1, to set the payload as head.

This is only a proof of concept and more tests shall be added in the future.


marioevz:
Added CL mock, which selects an Execution client to produce a block, and then broadcasts the resulting payload to the rest of the clients using the Engine API, mimicking the Consensus Layer.

marioevz:
Closing as I will be opening a new PR with only the simulator.

PR #495, opened 2021-12-14
simulators/eth2/testnet: merge transition testnet

lightclient:
Currently working on a PR to add a simulator for full PoW->PoS merge. Building off the work in #481.

---

The MVP of the this will be testing the "Happy Case":

#### Description

A mock client mines a normal PoW chain which reaches `TotalTerminalDifficulty`, CL+EL continue to finalization with live `ExecutionPayloads`.

#### Steps
* Mock client mines a terminal PoW block and propogates it to all EL clients.
* The next proposer constructs a beacon block with a `TRANSITION_BLOCK`.
* The proposer after that should build a block which builds an execution payload on top of the `TRANSITION_BLOCK`.
* So on and so forth until the chain finalizes a checkpoint that includes a non-zero `ExecutionPayload`.

#### Checks
- [x] Verify the chain post-transition finalizes.
- [x] Verify every CL client agrees on the finalized.
- [x] Verify attestation participation of finalized is above 95%.
- [x] Verify `execution_payload.block_hash` of a finalized checkpoint is in the canonical chain of all ELs.
- [x] Verify all beacon clients have proposed a block with a valid `ExecutionPayload`.


lightclient:
@fjl @holiman I think this simulator is ready for a review!

simulators/eth2/testnet/mock/p2p.go
```diff
@@ -0,0 +1,294 @@
...
+}
+
+type Pong struct{}
+
+func (p Pong) Name() string { return "pong" }
+func (p Pong) Kind() int    { return 0x03 }
+
+// Dial attempts to dial the given node and perform a handshake,
+// returning the created Conn if successful.
+func Dial(e *enode.Node) (*Conn, error) {
```
fjl, line 50: What is this file used for? I think it's not a good idea to add another half-baked eth protocol implementation just for this one test. Can we avoid this somehow?

fjl:
I think you have added the simulator binary accidentally, please remove.

fjl:
This PR is pretty large and contains all kinds of changes to clients, etc. I think I would prefer to get https://github.com/ethereum/hive/pull/496 in first and make eth1 client definitions ready for merge testing. We can work on the eth2 side concurrently to get things into shape. It would be nice to see another eth2 client definition like teku or prysm on here.

I want to prevent a scenario where we create a very messy WIP simulator that only works with geth+lighthouse.

fjl:
Also, another question: since this modifies the eth2 testnet simulator, does that mean there will no longer be a way to run stand-alone eth2 testnets without the merge?

simulators/eth2/testnet/mock/p2p.go
lightclient: Do you have a recommendation on how best to connect to a peer over devp2p and announce blocks? It seemed like this was a pretty minimal implementation of that, but would be better to use something that already exists.

fjl:
We have decided to create new client definitions temporarily while work to get merge tests running is ongoing. @marioevz will submit definitions for `merge-geth` and `merge-nethermind` shortly.

I will also update client definitions with support for Arrow Glacier.

It would be nice to get the eth2 client spec in separately. Please open a PR with just the `docs/` changes if possible.

simulators/eth2/testnet/main.go
```diff
@@ -1,18 +1,13 @@
...
 
-func jsonStr(v interface{}) string {
-	dat, err := json.MarshalIndent(v, "  ", "  ")
-	if err != nil {
-		panic(err)
-	}
-	return string(dat)
+var tests = []testSpec{
+	{Name: "transition-testnet-basic-geth-lighthouse", Fn: Bscmtt{node{"go-ethereum", "lighthouse"}}},
+	{Name: "transition-testnet-basic-geth-teku", Fn: Bscmtt{node{"go-ethereum", "teku"}}},
```
marioevz, line 10: I would suggest we use the provided clients to generate the client combinations instead of hard-coding the combinations here.

E.g. if we have:
Execution Clients: A
Beacon Client: B, C, D
Validator Client: E, F

The resulting testnet could be comprised of client combos:
A/B/E + A/C/E + A/D/E + A/B/F + A/C/F + A/D/F

clients/merge-go-ethereum/geth.sh
```diff
@@ -81,7 +81,7 @@ if [ "$HIVE_NODETYPE" == "archive" ]; then
     FLAGS="$FLAGS --syncmode full --gcmode archive"
 fi
 if [ "$HIVE_NODETYPE" == "full" ]; then
-    FLAGS="$FLAGS --syncmode snap"
+    FLAGS="$FLAGS --syncmode full"
```
lightclient, line 84: This is probably wrong to do, but is there a recommended way for client to full sync but not be in archive mode?

simulators/eth2/testnet/main.go
lightclient: Done, thanks.

clients/merge-go-ethereum/geth.sh
fjl: Why do you need to force full sync here?

lightclient: @fjl when I start a client in fast sync and begin propagating blocks to it, it discards them as it considers itself "syncing". I'm not sure how to resolve this, other than by forcing the node to be in full sync?

fjl: I see. We don't have this problem in other tests, and I think it's because we initialize the nodes with at least one block. So what I can recommend is, just create a chain of one block using `cmd/hivechain` and initialize the eth1 node with it.

lightclient: Ah, okay. I will give that a try. Thanks.

lightclient:
> S ince this modifies the eth2 testnet simulator, does that mean there will no longer be a way to run stand-alone eth2 testnets without the merge?

I've updated the eth2 testnet simulator to again be able to run non-merge testnets.

> It would be nice to get the eth2 client spec in separately

Removed the doc updates out of this PR, will submit a separate one for them.

clients/merge-go-ethereum/geth.sh
lightclient: Hmm, unfortunately I realized that because the genesis is dynamic (`timestamp` need to be close to current time to be interoperable with the beacon chain). I suppose this could be worked around, but it seems like it may open up many edge cases if the network is on a different time than actual.

fjl: What we could also do is, just launch another client to produce blocks, and have it propagate blocks to the merge client that is being tested. In geth, once a block has been imported via p2p, snap sync is considered done. Setting up an actual client to produce blocks might also remove the need to inject blocks from within the test code.

This reminds me, when I tried the eth2 simulator last time, it did PoW mining which made everything extremely slow. It would be so much better to use clique, and we do that in all other tests. Not sure if it is possible here because of TTD.

fjl: Oh, very cool, I see you already kind of implemented my suggestion in https://github.com/ethereum/hive/pull/495/commits/661822d706f4386bf8464ac7bf15740c51503105.

lightclient: Yes, decided that the mock functionality isn't really being utilized yet so it makes more sense to simplify the PR. Clique would definitely be simpler - the main reason I've opted to do PoW was to as closely mimic the real transition as possible. Now that I'm not using the mock module, it seems the eth1 client just mines on 1 thread so maybe this light enough to be more acceptable? I also adjusted the parameters accordingly. If you still think Clique is preferable, I can update to using that instead.

I'll need to test if the snap sync is working now, but I'm not sure why it would've been different with the mock module because it also announced the block over devp2p. Maybe because the geth client had outstanding snap requests to the mock client, which didn't know how to respond to them?

lightclient: Okay switching to snap worked. Going to mark this as resolved - please reopen if you'd like me to using Clique instead of PoW.

fjl:
Rebased, and I have played with this a bit. One issue is, the test gets stuck somehow about half the time. I think this happens when the machine is too slow and doesn't manage to start CPU mining fast enough.

fjl:
We need to change this simulator to use clique instead of ethash PoW, it will run WAY faster.

PR #496, opened 2021-12-16
simulators/ethereum: create Engine API simulator with Consensus Layer mock

marioevz:
This PR adds a new simulator ethereum/engine which tests the new merge Engine API by the use of a mocking tool which mimics the Consensus Layer in order to send testing API calls to the Execution Clients.

The simulator at the moment depends on MariusVanDerWijden/go-ethereum@merge-devnet-3 due to required structures to format the Engine API calls in "github.com/ethereum/go-ethereum/eth/catalyst", which are unavailable on the master branch.
In the case that this is not acceptable, it can be sorted out by simply copying the definitions into engineclient.go in this PR.

Currently, none of the clients will succeed in running the Engine API tests since this functionality is not included in the master branches of the clients, but the simulator can be used for local testing of the clients with the appropriate modifications to each of the clients' Dockerfile.

A full description of the simulator, along with the test cases currently implemented, is included in the Readme file.


simulators/ethereum/engine/main.go
```diff
@@ -0,0 +1,177 @@
...
+			defer s.put()
+			t.Run(hivesim.TestSpec{
+				Name:        fmt.Sprintf("%s (%s)", test.Name, c.Type),
+				Description: test.About,
+				Run: func(t *hivesim.T) {
+					runTest(test.Name, t, c, ec, vault, clMocker, &m, &st, test.Run)
+				},
+			})
+		}()
+	}
```
holiman, line 152: This is pretty unusual way of doing this, what's more common is to use a `sync.WaitGroup`. 

```suggestion
	var wg sync.WaitGroup
	for _, test := range tests {
		wg.Add(l1)
		test := test
		go func() {
			defer wg.Done()
			t.Run(hivesim.TestSpec{
				Name:        fmt.Sprintf("%s (%s)", test.Name, c.Type),
				Description: test.About,
				Run: func(t *hivesim.T) {
					runTest(test.Name, t, c, ec, vault, clMocker, &m, &st, test.Run)
				},
			})
		}()
	}
```
and later on, `wg.Wait()` instead of `s.drain()`.

simulators/ethereum/engine/clmock.go
```diff
@@ -0,0 +1,351 @@
...
+
+	// Merge related
+	FirstPoSBlockNumber         *big.Int
+	TTDReached                  bool
+	PoSBlockProductionActivated bool
+	NextFeeRecipient            common.Address
+
+	// PoS Status Mutexes: Set-Reset lock is used to guarantee each test finds the environment as expected,
+	// and not as previously modified by another test.
+	PayloadBuildMutex                SetResetLock
```
holiman, line 34: Not sure I understand the SetResetLock concept. What are the requirements on this mutex? What semantics do you want to achieve?

simulators/ethereum/engine/main.go
marioevz: Worked perfectly, thanks!

simulators/ethereum/engine/clmock.go
marioevz: This is an attempt to provide a clean slate for all test cases to start from.

The test cases request a lock to "wake up" at a specific point during the PoS block production procedure and, when the CL mocker has reached that point, it releases the lock to let the test cases' procedures do their testing.

The problem that came up when using a regular lock was that most of the test cases send API calls to test some functionality, and in doing so, the state at which the next test case finds the client is different from what the CL mocker left, and from what the test case expected. This resulted in tests failing.

The idea behind the set-reset is that only one test case runs each time the CL mocker reaches a certain point in its block production.

The CL mocker 'resets' the lock to signal that nobody else has gotten the lock since the last time it was released, and therefore the state of the client is exactly as expected.

If the test case gets the lock but it is already 'set', it knows that another test case got to run before them, and it can pass on getting the lock this time and try again until another block is produced and it gets the lock in a 'reset' state.

simulators/ethereum/engine/clmock.go
```diff
@@ -0,0 +1,365 @@
...
+// This function must NOT be executed after we have reached TTD.
+func (cl *CLMocker) checkTTD() {
+	if len(cl.EngineClients) == 0 {
+		// We have no clients running yet, we have not reached TTD
+		time.AfterFunc(tTDCheckPeriod, cl.checkTTD)
+		return
+	}
+
+	// Pick a random client to get the total difficulty of its head
+	ec := cl.EngineClients[rand.Intn(len(cl.EngineClients))]
```
lightclient, line 122: What is the purpose of choosing a random client?

simulators/ethereum/engine/clmock.go
```diff
@@ -0,0 +1,365 @@
...
+
+	lastBlockNumber, err := ec.Eth.BlockNumber(ec.Ctx())
```
lightclient, line 124: Is there a reason to calculate this iteratively? You should be able to get a block with `HeaderByNumber` and use the `latest` tag to get the head block, which includes the `totalDifficulty` in the response.

simulators/ethereum/engine/clmock.go
```diff
@@ -0,0 +1,365 @@
...
+
+// Check whether a block number is a PoS block
+func (cl *CLMocker) isBlockPoS(bn *big.Int) bool {
+	if cl.FirstPoSBlockNumber == nil {
+		return false
+	}
+	if cl.FirstPoSBlockNumber.Cmp(bn) <= 0 {
+		return true
+	}
+	return false
```
lightclient, line 169:
```suggestion
	if cl.FirstPoSBlockNumber == nil || cl.FirstPoSBlockNumber.Cmp(bn) > 0 {
		return false
	}
        return true
```

simulators/ethereum/engine/clmock.go
```diff
@@ -0,0 +1,365 @@
...
+
+// Sets the fee recipient for the next block and returns the number where it will be included.
+func (cl *CLMocker) setNextFeeRecipient(feeRecipient common.Address, ec *EngineClient) *big.Int {
+	for {
+		cl.PayloadBuildMutex.LockSet()
+		if ec == nil || cl.NextBlockProducer == ec {
+			cl.NextFeeRecipient = feeRecipient
+			defer cl.PayloadBuildMutex.Unlock()
+			return big.NewInt(cl.LatestFinalizedNumber.Int64() + 1)
+
```
lightclient, line 180:
```suggestion
```

simulators/ethereum/engine/clmock.go
marioevz: The idea was to cover testing of a client receiving an ExecutionPayload both created by itself and created by a different client.

Originally I thought of running more than 2 clients at the same time and it made sense to have a client picked randomly here, but perhaps we could simply round-robin.

simulators/ethereum/engine/clmock.go
marioevz: For some reason, totalDifficulty is not available in the struct returned by HeaderByNumber:
https://github.com/ethereum/go-ethereum/blob/1884f37f2c772d6f1d567432e76cbd7b90d442a9/core/types/block.go#L69

Neither does the Block struct, so instead of directly calling the RPC and then parsing the response, I calculated the difficulty manually.

lightclient: Ah I see, didn't realize there was no type `RPCHeader`. I think you could do something like this though?
```go
type td struct {
        TotalDifficulty *hexutil.Big `json:"totalDifficulty"`
}
var td *TotalDifficulty
err := ec.c.CallContext(ctx, &td, "eth_getBlockByNumber", "latest", false)
if err == nil && td == nil {
	panic("not found")
}
```

marioevz: Added in latest commit, thanks!

holiman:
> This is an attempt to provide a clean slate for all test cases to start from.

> The test cases request a lock to "wake up" at a specific point during the PoS block production procedure and, when the CL mocker has reached that point, it releases the lock to let the test cases' procedures do their testing.

I may be missing something, but it sounds to me like this goes a bit against the, kind of, Way To Do It in Hive. 

What I mean is: Hive is made to create a certain precondition, by spinning up nodes, feeding blocks etc, from scratch, and then execute a test. In some cases, like graphql, we don't need to do it all from scratch: as luck would have it, the tests can all run on the same instance, since they don't modify the state of the client. 

But in e..g the consensus tests, each test instantiates everything from scratch, guaranteeing a clean slate and reproducible results. 

Without having actually looked very deep into how these tests work under the hood, it _sounds_ like you are using the one-instance-for-all-tests approach, but use a custom locking scheme to make it work as if it were separate instances. So what I'm wondering is, wouldn't it make more sense to split up the tests to simply use standalone instances? 

As I said, I don't have a full picture of the nature of the tests, so I might totally have missed some crucial aspect.

marioevz:
Thanks for the comment @holiman, this simulator is indeed attempting to do a one-instance-for-all-tests approach, I followed this approach from the ethereum/rpc simulator:
https://github.com/ethereum/hive/blob/c2251f19b65520895bf17e7b999025ed93ecfae2/simulators/ethereum/rpc/main.go#L111

I had not considered spinning up different instances for each test case, the main issue would be that we would still have to halt the Consensus Layer mocker at some point for all instances in order to perform some of the test cases, because the test cases rely on the mocker halting their payload broadcast to guarantee that the instance only receives the test case's payload.

Nevertheless, I think it would be really interesting to implement the approach you suggest for two reasons:
- We would no longer require the special locking mechanism.
- The simulator would better test the consensus of multiple execution clients after PoS transition

One issue though would be that, if we have 15 test cases in the simulator, we would need 15 instances running simultaneously.
Is there a limit on how many instances should a hive simulator have?

holiman:
I think this is an improvement. I'll run the tests locally a bit and try them out.

marioevz:
> I think this is an improvement. I'll run the tests locally a bit and try them out.

@holiman you can use branch https://github.com/marioevz/hive/tree/engine-api-simulator-geth-merge which contains this simulator's changes plus client "go-ethereum-merge", which points to Marius' Kintsugi changes, and finally run using: `./hive --client go-ethereum-merge --sim engine`

marioevz:
Hi @holiman, just a heads up that I pushed a small fix for the CLMocker because it would panic and abort if any of the test cases resulted in a payload getting rejected by the Execution Client.

This happens almost every time during test case `Invalid ParentHash ExecutePayload`, and according to @MariusVanDerWijden this seems to be a known issue in the kintsugi branch.

The CLMocker now tries with each client and only panics if all of them fail.

The fix is already in `engine-api-simulator-geth-merge` if you'd like to run the tests locally.

simulators/ethereum/engine/engineclient.go
```diff
@@ -0,0 +1,126 @@
...
+			t:     t,
+			hc:    hc,
+			inner: http.DefaultTransport,
+		},
+	}
+	// Prepare HTTP Client
+	rpcHttpClient, _ := rpc.DialHTTPWithClient(fmt.Sprintf("http://%v:%v/", hc.IP, EnginePortHTTP), client)
+	// Prepare WS Client
+	ctx, done := context.WithTimeout(context.Background(), 5*time.Second)
+	rpcWsClient, err := rpc.DialWebsocket(ctx, fmt.Sprintf("ws://%v:%v/", hc.IP, EnginePortWS), "")
```
MariusVanDerWijden, line 52: You would usually `defer done()` before the `DialWebsocket` call here, otherwise the dial will not timeout properly

simulators/ethereum/engine/engineclient.go
```diff
@@ -0,0 +1,126 @@
...
+		ec.lastCancel()
+	}
+	ec.lastCtx, ec.lastCancel = context.WithTimeout(context.Background(), rpcTimeout)
+	return ec.lastCtx
+}
+
+// Engine API Call Methods
+func (ec *EngineClient) EngineForkchoiceUpdatedV1(ctx context.Context, fcState *catalyst.ForkchoiceStateV1, pAttributes *catalyst.PayloadAttributesV1) (catalyst.ForkChoiceResponse, error) {
+	var result catalyst.ForkChoiceResponse
+	err := ec.http_c.CallContext(ctx, &result, "engine_forkchoiceUpdatedV1", fcState, pAttributes)
```
MariusVanDerWijden, line 112: You have this really complicated switching logic to switch between http and ws and in the end you hardcode http here?
I think you should delete most of the switching logic and use ec.c here

marioevz: The protocol switching was removed all together.

simulators/ethereum/engine/engineclient.go
marioevz: This part has been removed altogether .

simulators/ethereum/engine/clmock.go
```diff
@@ -0,0 +1,424 @@
...
+		}
+
+	}
+
+	cl.PayloadBuildMutex.Unlock()
+	cl.PayloadBuildMutex.Lock()
+
+	// Generate a random value for the Random field
+	nextRandom := common.Hash{}
+	for i := 0; i < common.HashLength; i++ {
```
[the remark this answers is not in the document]
marioevz: Updated, thanks.

simulators/ethereum/engine/clmock.go
```diff
@@ -0,0 +1,424 @@
...
+	NewGetPayloadMutex               sync.Mutex
+	NewExecutePayloadMutex           sync.Mutex
+	NewHeadBlockForkchoiceMutex      sync.Mutex
+	NewSafeBlockForkchoiceMutex      sync.Mutex
+	NewFinalizedBlockForkchoiceMutex sync.Mutex
+}
+
+func NewCLMocker(t *hivesim.T) *CLMocker {
+	// Init random seed for different purposes
+	rand.Seed(time.Now().Unix())
```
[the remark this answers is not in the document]
marioevz: Updated, thanks.

simulators/ethereum/engine/clmock.go
```diff
@@ -0,0 +1,424 @@
...
+		}
+		// Unlock and keep trying to get the requested Engine Client
+		cl.PayloadBuildMutex.Unlock()
+		time.Sleep(PoSBlockProductionPeriod)
+	}
+}
+
+// Unlock all locks in the given CLMocker instance
+func unlockAll(cl *CLMocker) {
+	cl.PayloadBuildMutex.Unlock()
```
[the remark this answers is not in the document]
marioevz: I have updated the mechanism to use a channel-based semaphore, which also is closed when the block production ceases completely.

Let me know if this looks better.

simulators/ethereum/engine/clmock.go
```diff
@@ -0,0 +1,424 @@
...
+	cl.NewSafeBlockForkchoiceMutex.Unlock()
+	cl.NewFinalizedBlockForkchoiceMutex.Unlock()
+}
+
+// Mine a PoS block by using the Engine API
+func (cl *CLMocker) minePOSBlock() {
+	cl.EngineClientsLock.Lock()
+	defer cl.EngineClientsLock.Unlock()
+	if cl.BlockProductionMustStop {
+		unlockAll(cl)
```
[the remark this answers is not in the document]
marioevz: I have updated minePOSBlock to remove the time.AfterFunc() usage and rely in a single loop which sleeps between blocks being produced, so it runs in a single thread now, and I think it looks much cleaner/safe now.

EngineClientsLock is the only sync.Mutex I have not replaced with a channel-based semaphore.

Let me know if this is ok.

marioevz:
Hi @holiman, I addressed Marius comments in the latest commits, please let me know if the changes are ok.

There is also an upcoming update to the Execution API, but I think the updates/new test cases for those changes should be addressed in a separate PR after this one is merged.

Thanks.

simulators/ethereum/engine/enginetests.go
```diff
@@ -0,0 +1,837 @@
...
+	// Wait until TTD is reached by this client
+	if !t.WaitForPoSSync() {
+		t.Fatalf("FAIL (%v): Timeout on PoS sync", t.TestName)
+	}
+
+	// Wait for GetPayload
+	if closed := t.CLMock.OnGetPayload.Wait(); closed {
+		t.Fatalf("FAIL (%v): CLMocker stopped producing blocks", t.TestName)
+	}
+	defer t.CLMock.OnGetPayload.Yield()
```
holiman, line 246: The wait/yield looks pretty nice enough, but it's not what I'd call go-idiomatic - more pythonic maybe? 

We typically do it something like this: 
```suggestion
        select{ 
            case payload := < t.CLMock.payloadCh:
               // got payload ok
             case time.After(1 * time.Second) : 
               // time-out
               	t.Fatalf("FAIL (%v): CLMocker timed out", t.TestName)
              case shutdownCh:
               // shutdown channel closed
               	t.Fatalf("FAIL (%v): CLMocker stopped producing blocks", t.TestName)
       }
```
Having a `select` means we can act on the chan we're interested in (in this case `payloadCh`), but we can also have other cases, to handle timeouts (typically good to have in case there's  some deadlock somewhere) and/or if the whole thing is shutting down. 
Then for the payloadCh, you'd only ever deliver payloads over it. If/when you want to shutdown, you just close the `shutdownCh`. Any/all listeners currently on such a `select` will _all_ react to the `shutdownCh` close, and they can all exit nicely.

holiman:
I pushed some small changes on top. 
I kind of see what you're doing now with those Wait/Yield -- so essentially a test wants to let the Mocker go through the requisite steps, until it reaches the point where the test is concerned. At that point, the CL should wait until the test says that it can continue. Is that correct?

But if each test runs on one instance, couldn't just the test tell CL to go through a certain point and then do it's thing? Or do some tests need to do things on several points of interest? It seems most of them do the `Yield` on `defer` -- which I _think_ means that they tell the CL to continue progressing after they are done -- but doesn't that mean that the test is concluded and the CL can just shut down?


I tested the PR with normal go-ethereum (not merge-enabled), which caused it to just never complete (on the second test). I think we need to make it so that there are some high-level timeouts - because what I suppose happens is that TTD is never reached, so it just goes on sealing blocks forever.

simulators/ethereum/engine/helper.go
```diff
@@ -24,7 +24,7 @@ import (
 // default timeout for RPC calls
 var rpcTimeout = 10 * time.Second
 
-// TestClient is the environment of a single test.
+// TestClient is 00the environment of a single test.
```
MariusVanDerWijden, line 27: 00? are you using i3?

marioevz:
Hi @holiman, thanks for all the feedback, I've updated the simulator with the following changes:
- All tests now run using a single client, and a new instance of the CLMocker each. We could start a second client if necessary to test PoW -> PoS sync for example.
- The CLMocker has a shutdown channel which is closed once the block production ends/fails.
- All tests use select to listen to the CLMocker shutdown, or timeout, or the events from the CLMocker, and properly exit or continue depending on the received channel signal.

Thanks again, and let me know what you think.

simulators/ethereum/engine/helper.go
holiman: That was my bad.. And yes, I use i3. But how is that typo attributable to i3???

marioevz:
Update:
- Rebase to ethereum/master
- Updated CLMocker and all existing tests to new [Execution API spec v1.0.0-alpha.6](https://github.com/ethereum/execution-apis/blob/v1.0.0-alpha.6/src/engine/specification.md)
- Refactored CLMocker routines to introduce more timeout checks, use shutdown and exit channels

fjl:
I have committed the simulator API changes spec'd in #502 to the master branch. To ensure your simulator will keep working, you must upgrade the hivesim dependency next time you rebase this PR.

To do this, run

```
cd simulators/ethereum/engine
go get -d github.com/ethereum/hive@master
go mod tidy
```

holiman (approved):
LGTM

holiman:
Nice piece of work @marioevz! Let's see it on prod!

PR #499, opened 2022-01-18
hivesim: add get client enode in specific network / clients: add symlink to enode.sh in hive-bin

marioevz:
Add API HTTP endpoint to fetch the eth client's enode/IP on a custom docker network.

This addition will be useful to create test cases where a secondary network is used as the main connection between clients, and can be disconnected on demand to simulate network partitioning.

### DIFF
--- a/clients/aleth/Dockerfile
+++ b/clients/aleth/Dockerfile
@@ -15,6 +15,8 @@ ADD genesis.json /genesis.json
 # Add the enode URL retriever script.
 ADD enode.sh /enode.sh
 RUN chmod +x /enode.sh
+RUN mkdir /hive-bin
+RUN ln -s /enode.sh /hive-bin/enode.sh
 
 # Copy buildinfo into version.json, and also remove the 'bool' field is_prerelease.
 RUN jq -r '.version' /usr/share/aleth/buildinfo.json > /version.txt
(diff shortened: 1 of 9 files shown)


fjl:
I think this can be done on the client side.

The API endpoint to get the enode URL is semi-deprecated, we would rather move the enode functionality [into a script](https://github.com/ethereum/hive/blob/master/docs/simulators.md#running-client-scripts).

marioevz:
Going through the current simulators, it seems like most of them use the `(c *Client) EnodeURL()` to obtain a client's Enode, which in turn uses the API endpoint.

Is it now the recommended approach to use `(c *Client) Exec(command ...string)` to call `enode.sh` in the client's container, and then manually append the result of `(sim *Simulation) ContainerNetworkIP` for a given client's container?

fjl:
No, it's not yet recommended, just wanted to mention this because its an idea for long-term. The server-side replacing of the IP was there before we introduced ContainerNetworkIP. Now that the simulator can get the IP, we could change hivesim to do this transparently.

It can be done like this:

- Add the script for all clients.
- Change hivesim to call the script.
- Update hive dependency in simulators.
- Remove the enode API endpoint.

fjl:
I have performed all necessary simulator upgrades and the old endpoint is now removed on the master branch.

PR #501, opened 2022-01-26
internal/libhive: add client option for initially connected networks

marioevz:
Adds an option to the StartClient API to include a set of networks in the request to which the new client must already be connected at startup.

The added "NETWORKS" form value must be an array of strings containing the names of existing networks to which the new client will be connected before it starts to run.

Example application of this new feature includes Consensus client Lighthouse where enr-address is passed as command line parameter for the client to generate a valid ENR in a different network than the default.

### DIFF
--- a/internal/libhive/api.go
+++ b/internal/libhive/api.go
@@ -215,14 +215,16 @@ func (api *simAPI) startClient(w http.ResponseWriter, r *http.Request) {
 	// Get the client name.
 	clientDef, err := api.checkClient(r)
 	if err != nil {
-		http.Error(w, fmt.Sprintf("could not parse node request: %s", err.Error()), http.StatusBadRequest)
+		log15.Error("API: " + err.Error())
+		http.Error(w, err.Error(), http.StatusBadRequest)
 		return
 	}
 
-	// Get the network names, if any, for the container to be connected to at start
-	networks, err := api.checkClientNetworks(r)
+	// Get the network names, if any, for the container to be connected to at start.
+	networks, err := api.checkClientNetworks(r, suiteID)
 	if err != nil {
-		http.Error(w, fmt.Sprintf("could not parse node request: %s", err.Error()), http.StatusBadRequest)
+		log15.Error("API: " + err.Error())
+		http.Error(w, err.Error(), http.StatusBadRequest)
 		return
 	}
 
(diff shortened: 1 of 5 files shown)


fjl:
This is a good idea. Parsing the new option is very annoying because of the silly JSON-in-urlencoding API format, but that's not your fault. We really need to fix the API to use just JSON.

fjl:
Aww, damn, I've checked it and since this is the client startup request that also handles uploaded files, we can't easily convert this to use JSON.

marioevz:
Hi @fjl, thanks for checking this, would using a comma-separated string with the names of the networks be better alternative at the moment ?

fjl:
We should also add a StartOption for initial networks in hivesim, i.e.

```
func WithInitialNetworks(networks []string) StartOption
```

internal/libhive/api.go
```diff
@@ -218,6 +218,13 @@ func (api *simAPI) startClient(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	// Get the network names, if any, for the container to be connected to at start
+	networks, err := api.checkClientNetworks(r, w)
+	if err != nil {
+		// We return here so we don't even create the container in bad requests
+		return
```
fjl, line 225: You need to return something to the client here, otherwise it will get status code 200 and no response body. We usually also log something in this case. See line 195 for an example.

fjl: Oh wait, I see the function `checkClientNetworks` already sets the response.

internal/libhive/api.go
```diff
@@ -315,6 +333,41 @@ func (api *simAPI) checkClient(r *http.Request, w http.ResponseWriter) (*ClientD
...
+	// Check that each network indeed exists
+	for _, network := range networks {
+		netExists, err := api.tm.NetworkExists(suiteID, network)
+		if err != nil {
+			http.Error(w, err.Error(), http.StatusBadRequest)
+			return nil, err
+		}
+		if !netExists {
+			http.Error(w, fmt.Sprintf("network '%v' not found", network), http.StatusBadRequest)
+			return nil, errors.New("Unknown network requested")
```
fjl, line 364: Error strings should not be capitalized.

internal/libhive/api.go
fjl: I would prefer to handle / log the error here, i.e. don't pass the response to `checkClientNetworks`.

marioevz: Yes, I agree, it looks confusing having a simple return here. Will change it no problem.

marioevz: Should I take the opportunity to also change lines 216-219 to use the same pattern ?

fjl: Why not!

fjl:
Here's my plan for further cleaning up of the API: https://github.com/ethereum/hive/issues/502

I'm happy to make those changes next week, just wanted to let you know.

PR #507, opened 2022-02-24
simulators/ethereum/engine: Update Engine API tests to be compatible with Kiln public testnet

marioevz:
This PR updates the tests in the Engine API simulator to match the requirements of the first Kiln public testnet.

The changes included are: 
- Renaming of the `RANDOM` opcode to `PREVRANDAO`, including the fields of `ExecutionPayloadV1` and `PayloadAttributesV1` (`json: "random"` -> `json: "prevRandao"`).
- Engine API HTTP port changed to 8550, which includes the JWT in the header, which is part of kiln v2, and not required for the public testnet, however this is not a breaking change (Kiln v2 will have 8551 as the single authenticated port).
- Engine API go data structures were copied from Marius' branch to be able to make the prevRandao rename.
- Added `HIVE_JWT_AUTH_TOKEN_SECRET` which is expected to be parsed by the clients on Kiln v2 when JWT authentication is required.
- Added test cases for invalid transactions in the `NewPayload` directive.
- Added sanity transaction tracing for `PREVRANDAO` transaction tests for Geth and Nethermind.
- Added ExchangeTransitionConfiguration data structures (unused until Kiln v2).


fjl:
Regarding the secret, it would be simpler to just hard-code the secret in all clients.

marioevz:
Added a new test with an invalid PrevRandao in NewPayload, and addressed comment by @fjl on using a default JWT secret (removing HIVE_JWT_AUTH_TOKEN_SECRET).

fjl:
I tried running this today, but there are two problems. In recent go-ethereum commits, the flag names were changed again and configuration is now done using `--authrpc.port` and `--authrpc.jwtsecret`. But even when I change the flags, it doesn't work because the API server only listens on 127.0.0.1, so it's not reachable when running in Docker. I will fix this in go-ethereum upstream.

PR #516, opened 2022-03-31
simulators/ethereum/engine: add auth tests, fix re-org tests

marioevz:
This PR adds JWT authentication to all supported clients and also adds tests for the authentication as defined in the spec.
Basic tests for the ExchangeTransitionConfiguration but these are at the moment failing -- need to ask client devs regarding the expected outcome of these tests.
Existing re-org tests are now modified so the forkchoiceUpdated directive does not re-org back into the canonical chain, but always attempts to re-org to a new payload (sidechain).


marioevz:
I will be creating several and more succinct PRs that are easier to review than this big PR.

PR #520, opened 2022-04-06
simulators/engine: Fix re-org tests

marioevz:
This PR fixes Engine API tests that perform a re-org, in order to ensure that the forkchoiceUpdated directive always points to a side-chain and does not attempt to jump back and forth to previous blocks on the canonical chain.

However, a simple sanity test is added which performs this behavior, but in this case the test simply expects that the client does not fail and is capable of progressing after such event.


holiman (approved):
LGTM

PR #521, opened 2022-04-07
simulators/ethereum/engine: Increase bad hash tests coverage

marioevz:
Adds first batch of new tests based on the coverage document by @mkalinin ([link](https://hackmd.io/vjgC9hV_TrK1ZuftVm1rZA)).

Tests added:
- EL Not SYNCING: Invalid hash payload extending side chain
- EL SYNCING: Invalid hash payload extending canonical chain
- EL SYNCING: Invalid hash payload extending side chain chain

Follow up PRs will fill remaining gaps. Please advice if this format is fine: small number of new tests per PR, or a big PR with all test cases is better.


marioevz:
Hey @fjl @holiman, let me know if these changes are ok.

I changed some string constants for an Enum, which I think looks cleaner.

Also I added a comment with a diagram to better exemplify what the test case is trying to cover, but let me know if these kinds of diagrams are frowned upon :eyes:

marioevz:
I'm converting this to draft since I would like to merge a refactor that much improves readability of the test cases before this one.

fjl:
IMHO the enum is fine, but you can also just define constants for the string values.

Diagrams also fine.

PR #523, opened 2022-04-14
clients: add nimbus-eth1 client

jangko:


fjl:
I have tested this, and there were some minor issues we need to figure out later. One thing is, the client container takes a long time to build because it recompiles everything from scratch. In other clients, we just use a release binary.

Another issue is that the rpc simulator just hangs when running against nimbus. This could also be an issue with the simulator though.

PR #540, opened 2022-05-13
simulators/eth2/engine: Implement EL mock to test CLs

MariusVanDerWijden:
This PR implements a new class of simulators used to test CL implementations. We start a proxy between EL and CL which allows us to modify requests and responses to and from the EL.


simulators/eth2/engine/proxies.go
```diff
@@ -0,0 +1,77 @@
...
+
+func (p *Proxy) UserRPCAddress() (string, error) {
+	return fmt.Sprintf("http://%v:%d", p.IP, p.port), nil
+}
+
+func (p *Proxy) EngineRPCAddress() (string, error) {
+	return fmt.Sprintf("http://%v:%d", p.IP, p.port), nil
+}
+
+func (p *Proxy) AddRequest(spoof *proxy.Spoof) {
```
marioevz, line 58: I think it would be great to have the option to add a callback, something like `func (p *Proxy) AddRequestCallback(callback func(interface{}) *proxy.Spoof)` , and the callback function receives the request and you can respond on a request-by-request basis whether to spoof or not.

I think at the moment `AddRequests` spoofs every single request it receives from the CL which is ok, but being able to decide on which requests to spoof would let us fine-tune the test cases. E.g. a test case where only the transition payload is corrupted, but every payload after that is ok.

marioevz (commented):
This is amazing! I only have one suggestion on the changes which I think would help us greatly to design the test cases more deterministically.

simulators/eth2/engine/proxies.go
MariusVanDerWijden: Yes, I thought about this too, but unfortunately raul's proxy does not support that. I can try to implement it in the proxy so we can use it in hive eventually

marioevz: I've added a PR to raul's repo to add this functionality: https://github.com/rauljordan/engine-proxy/pull/4
I haven't tested the changes but will do tomorrow.

simulators/eth2/engine/prepared_testnet.go
```diff
@@ -0,0 +1,260 @@
...
+		if index < 0 || index >= len(testnet.eth1) {
+			testnet.t.Fatalf("only have %d eth1 nodes, cannot find index %d for BN", len(testnet.eth1), index)
+		}
+	}
+
+	var addrs []string
+	var engineAddrs []string
+	for _, index := range eth1Endpoints {
+		eth1Node := testnet.proxies[index]
+		userRPC, err := eth1Node.UserRPCAddress()
```
marioevz, line 206: I had trouble using the `UserRPCAddress` of the proxy, I kept getting "the method net_version does not exist/is not available" errors.
I used `testnet.eth1[index].UserRPCAddress` for the `userRPC` and `testnet.proxies[index]` for `engineRPC`.

marioevz:
This looks good to me.

I tested the callbacks from my PR in the proxy and it's working ok.

There are a couple of changes required for the callbacks to work in this simulator, but we have to wait for my PR to be merged in Raul's engine-proxy.

I can do these required changes in a later PR, or we can wait and incorporate them here, however you decide.

holiman (approved):
It's fine for me to merge this, if you think it's ready?

MariusVanDerWijden:
I think it's okay

PR #543, opened 2022-05-17
clients/besu: add support for engine API tests

jflo:
Besu now supports both http/websockets as well as jwt auth/unauth on a single port. Defaults to 8551 like all of its friends.


clients/besu/besu.sh
```diff
@@ -124,25 +124,31 @@ else
...
 if [ "$HIVE_GRAPHQL_ENABLED" == "" ]; then
     RPCFLAGS="$RPCFLAGS --rpc-http-enabled --rpc-http-api=ETH,NET,WEB3,ADMIN --rpc-http-host=0.0.0.0"
 else
-    RPCFLAGS="$RPCFLAGS --rpc-http-port=8550" # work around duplicate port error
     RPCFLAGS="$RPCFLAGS --graphql-http-enabled --graphql-http-host=0.0.0.0 --graphql-http-port=8545"
 fi
 
 # Enable WebSocket.
-RPCFLAGS="$RPCFLAGS --rpc-ws-enabled --rpc-ws-api=ETH,NET,WEB3,ADMIN --rpc-ws-host=0.0.0.0"
+#RPCFLAGS="$RPCFLAGS --rpc-ws-enabled --rpc-ws-api=ETH,NET,WEB3,ADMIN --rpc-ws-host=0.0.0.0"
```
fjl, line 145: No WebSocket anymore?

jflo: Websocket is still in place for the "main" stack. Only the engine api is on a combined http stack that upgrades from http to websockets.

holiman (approved):

clients/besu/besu.sh
fjl: Ahh, so `--rpc-ws-enabled` is not needed anymore to enable it? I was just wondering why you commented out this flag here.

fjl: I have tested this, and WebSocket is no longer enabled on besu with this change.

PR #548, opened 2022-05-18
clients/erigon: add support for clique private key

Giulio2002:


fjl:
It doesn't seem to work: 

```
[a665f74b] [EROR] [05-19|10:50:43.544] Erigon startup                           err="signer missing: %!w(<nil>)"
```

fjl:
I checked again, and erigon starts now, but blocks aren't always produced. This prevents the rpc test suite from running correctly, for example. Merging anyway, because this is an improvement now.

PR #550, opened 2022-05-21
clients/nimbus: add nimbus beacon node and validator client

marioevz:
Adds `nimbus-bn` and `nimbus-vc`, which are compiled from source from branch https://github.com/status-im/nimbus-eth2/tree/kiln-dev-auth, with a custom setting `SECONDS_PER_SLOT=3`, to allow quicker test runtimes.

### DIFF
--- a/clients/nimbus-bn/nimbus_bn.sh
+++ b/clients/nimbus-bn/nimbus_bn.sh
@@ -24,6 +24,7 @@ sed -i 's/TERMINAL_BLOCK_HASH: "\(0x[[:xdigit:]]\+\)"/TERMINAL_BLOCK_HASH: \1/' 
 sed -i 's/GENESIS_FORK_VERSION: "\(0x[[:xdigit:]]\+\)"/GENESIS_FORK_VERSION: \1/' /data/testnet_setup/config.yaml
 sed -i 's/ALTAIR_FORK_VERSION: "\(0x[[:xdigit:]]\+\)"/ALTAIR_FORK_VERSION: \1/' /data/testnet_setup/config.yaml
 sed -i 's/BELLATRIX_FORK_VERSION: "\(0x[[:xdigit:]]\+\)"/BELLATRIX_FORK_VERSION: \1/' /data/testnet_setup/config.yaml
+sed -i 's/CAPELLA_FORK_VERSION: "\(0x[[:xdigit:]]\+\)"/CAPELLA_FORK_VERSION: \1/' /data/testnet_setup/config.yaml
 sed -i 's/SHARDING_FORK_VERSION: "\(0x[[:xdigit:]]\+\)"/SHARDING_FORK_VERSION: \1/' /data/testnet_setup/config.yaml
 echo Using config.yaml
 cat /data/testnet_setup/config.yaml
--- a/clients/nimbus-bn/nimbus_bn.sh
+++ b/clients/nimbus-bn/nimbus_bn.sh
@@ -1,0 +1,74 @@
+#!/bin/bash
+
+# Immediately abort the script on any error encountered
+set -e
+
+if [ ! -f "/hive/input/genesis.ssz" ]; then
+    if [ -z "$HIVE_ETH2_ETH1_RPC_ADDRS" ]; then
+      echo "genesis.ssz file is missing, and no Eth1 RPC addr was provided for building genesis from scratch."
+      # TODO: alternative to start from weak-subjectivity-state
+      exit 1
+    fi
+fi
+
+mkdir -p /data/testnet_setup
+
+cp /hive/input/genesis.ssz /data/testnet_setup
+cp /hive/input/config.yaml /data/testnet_setup
+
+# znrt encodes the values of these items between double quotes (""), which is against the spec:
+# https://github.com/ethereum/consensus-specs/blob/v1.1.10/configs/mainnet.yaml
+# Nimbus is the only client that complains about this.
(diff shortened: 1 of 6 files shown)


clients/nimbus-bn/nimbus_bn.sh
```diff
@@ -0,0 +1,74 @@
...
+cat /data/testnet_setup/config.yaml
+echo ""
+
+echo "${HIVE_ETH2_CONFIG_DEPOSIT_CONTRACT_ADDRESS:-0x1111111111111111111111111111111111111111}" > /data/testnet_setup/deposit_contract.txt
+echo "${HIVE_ETH2_DEPOSIT_DEPLOY_BLOCK_NUMBER:-0}" > /data/testnet_setup/deploy_block.txt
+
+mkdir -p /data/beacon
+chmod 0700 /data/beacon
+
+LOG=INFO
```
tersec, line 38: Not sure if people are expected to look at the log output in any meaningful way here, but if the disk resources are available, it's useful to have debug logs.

fjl: The log level can be configured by hive, see below.

fjl:
@marioevz is this ready for merging?

marioevz:
> @marioevz is this ready for merging?

After the latest commit fixing the capella fork parsing, nimbus is still failing to start with this error:
```
ERR 2022-06-23 16:28:38.758+00:00 State from database does not match network, check --network parameter genesis=2e3c320f:0 tail=2e3c320f:0 headRef=2e3c320f:0 stateFork="(previous_version: 02000000, current_version: 02000000, epoch: 0)" configFork="(previous_version: 01000000, current_version: 02000000, epoch: 0)"
```
The error is caused by `ALTAIR_FORK_EPOCH == BELLATRIX_FORK_EPOCH == 0`.

We could workaround this issue by setting `ALTAIR_FORK_EPOCH == 1`, `BELLATRIX_FORK_EPOCH == 2` in the simulator,
but I think it would be better for this issue to be fixed directly in nimbus (accepting all forks to be equal to 0).

Either way, I don't think more changes are required for the files in this PR, so it can be merged.

tersec:
> ERR 2022-06-23 16:28:38.758+00:00 State from database does not match network, check --network parameter genesis=2e3c320f:0 tail=2e3c320f:0 headRef=2e3c320f:0 stateFork="(previous_version: 02000000, current_version: 02000000, epoch: 0)" configFork="(previous_version: 01000000, current_version: 02000000, epoch: 0)"
> ```
> 
> The error is caused by `ALTAIR_FORK_EPOCH == BELLATRIX_FORK_EPOCH == 0`.
> 
> We could workaround this issue by setting `ALTAIR_FORK_EPOCH == 1`, `BELLATRIX_FORK_EPOCH == 2` in the simulator, but I think it would be better for this issue to be fixed directly in nimbus (accepting all forks to be equal to 0).

Nimbus does allow `ALTAIR_FORK_EPOCH == BELLATRIX_FORK_EPOCH == 0`; the check it's performing in `checkForkConsistency` is
```nim
    doAssert distinctBase(firstForkEpoch) < distinctBase(secondForkEpoch) or
             firstForkEpoch in [GENESIS_EPOCH, FAR_FUTURE_EPOCH]
```
This allows for multiple forks defined with `FORK_EPOCH`s of 0 or "never".

What goes wrong is in trying to do this while providing a phase 0 genesis state, where it performs checks such as
```nim
      try:
        genesisState = newClone(readSszForkedHashedBeaconState(
          cfg,
          genesisStateContents.toOpenArrayByte(0, genesisStateContents.high())))
      except CatchableError as err:
        raiseAssert "Invalid baked-in state: " & err.msg
```
If one runs `launch_local_testnet.sh`, this is where it dies.

This is because
```nim

  var metadata = config.loadEth2Network()

  if config.terminalTotalDifficultyOverride.isSome:
    metadata.cfg.TERMINAL_TOTAL_DIFFICULTY =
      parse(config.terminalTotalDifficultyOverride.get, UInt256, 10)

  # Updating the config based on the metadata certainly is not beautiful but it
  # works
  for node in metadata.bootstrapNodes:
    config.bootstrapNodes.add node

  template applyConfigDefault(field: untyped): untyped =
    if config.`field`.isNone:
      if not metadata.configDefaults.`field`.isZeroMemory:
        info "Applying network config default",
          eth2Network = config.eth2Network,
          `field` = metadata.configDefaults.`field`
      config.`field` = some metadata.configDefaults.`field`

  applyConfigDefault(lightClientEnable)
  applyConfigDefault(lightClientDataServe)
  applyConfigDefault(lightClientDataImportMode)

  let node = BeaconNode.init(
    metadata.cfg,
    rng,
    config,
    metadata.depositContractDeployedAt,
    metadata.eth1Network,
    metadata.genesisData,
    metadata.genesisDepositsSnapshot)
```
here, it's looking for which preset to use, and the only presets it knows about start from phase 0. That's not a requirement though, and one can point it to other network metadata instead.

This is because it's trying to `readSszForkedHashedBeaconState` a phase 0 genesis state when its fork schedule tolls it that it's in the Bellatrix fork. One solution is to make those two things consistent, and in particular, set up a network configuration (which one can pass to Nimbus using `--network=${PATH_TO_CONFIG}` such as those in https://github.com/eth-clients/merge-testnets or https://github.com/eth-clients/eth2-networks from which Nimbus derives its network presets to begin with.

I'm not sure whether it's worth increasing the Nimbus binary size by another 15MB or so, a typical bump, for this purpose, since for Nimbus CI purposes, I prefer that it runs through phase0 and altair at least somewhat, to detect regressions. As forks pile on, and once Capella needs to be tested, I might re-evaluate this, but not in the immediate timeframe.

PR #551, opened 2022-05-23
clients/besu: remove --Xmerge-support and --engine-rpc-enabled options since no more needed

fab-10:
There is no more the need to manually enable merge support, since it is automatically enabled when TTD is present in the genesis file


fab-10:
PR ready

PR #554, opened 2022-05-25
simulators/ethereum/engine: Terminal Block and Transition Payload checks

marioevz:
This PR adds checks on the transition tests, specifically to verify the return values of the Engine API endpoints on:
- `newPayload(TP)` where `TP` is a valid transition payload built on top of a valid terminal block.
- `forkchoiceUpdated(TB)` where `TB` is a valid terminal block.
- `newPayload(INV_TP)` where `INV_TP` is an invalid payload built on top of a terminal block where `TB.Parent.TD >= TTD`
- `newPayload(INV_TP)` where `INV_TP` is an invalid payload built on top of a terminal block where `TB.TD < TTD`
- `forkchoiceUpdated(INV_TB)` where `INV_TB.Parent.TD >= TTD`
- `forkchoiceUpdated(INV_TB)` where `INV_TB.TD < TTD`

These changes improve on previous tests where the only passing requirement was that the client didn't switch to the invalid chain. Here the `INVALID` response to the CL must be verified.

cc @mkalinin


mkalinin (approved):
LGTM! 👍

PR #557, opened 2022-05-28
simulators/ethereum/engine: fix Invalid Number NewPayload (Syncing)

MariusVanDerWijden:
Fixes `Invalid Number NewPayload (Syncing)`
An invalid number can force us to start a sync cycle, since NewPayload allows to return `Syncing`, this should confirm to the spec


fjl:
@marioevz let me know if you think this was OK to add

PR #558, opened 2022-05-30
cmd/hivechain: implement generate pos command

MariusVanDerWijden:


fjl:
Seems to me that the `-pos` option should be applied in `generateAndSave` rather than cloning this function with only minor changes.

MariusVanDerWijden:
yes

marioevz:
Looks great, using this could we generate pos blocks with uncles just for testing purposes?
Would the process to create the invalid blocks is to first create the chain using hivechain and then manually modify the rlp file externally ?

PR #560, opened 2022-06-01
simulators/ethereum/engine: fix Reorg using Sync tests

MariusVanDerWijden:
The tests used to start two node `n1` and `n2`, and create a chain `p1...p10` with `p9` (usually) being the bad block. `p9` and `p10` were fed to node `n2`, while `p1...p8` were fed to `n1`. The test then sends `forkChoiceUpdates` to block `p10` to `n2` and expects `{INVALID, LVH:p8}`. However this requires `n2` to cache the un-validated payloads `p9` and `p10` and execute them later on. This caching is not required by the spec though.

The new tests use a slightly hacked version of geth as `n1` which allows for importing invalid payloads, so we feed `p1...p9` to `n1` and then send `forkChoiceUpdates` to `n2` for block `p10`. `n2` should sync from `n1`, detect the invalid chain and return `{INVALID, LVH:p8}`.

I also refactored the `Reorg using Sync` and `Reorg using NewPayload` out for better readability. 

The only test that still fails is `"Invalid Ancestor Chain Re-Org, Invalid Number, Invalid P9', Reveal using sync (go-ethereum)"`. Since the header chain is not continuous, because the number is invalid, we don't have the parent saved in our database therefore we will panic. I might just remove this test for cleanliness as it doesn't really test that much


simulators/ethereum/engine/enginetests.go
```diff
@@ -1217,108 +1197,222 @@ func invalidMissingAncestorReOrgGen(invalid_index int, payloadField InvalidPaylo
...
-								SafeBlockHash:      altChainPayloads[i].BlockHash,
-								FinalizedBlockHash: common.Hash{},
-							}, nil)
-							t.Logf("INFO (%s): Response from main client fcu: %v", t.TestName, s.Response.PayloadStatus)
-						}
+					r := t.TestEngine.TestEngineNewPayloadV1(altChainPayloads[i])
+					p := t.TestEngine.TestEngineForkchoiceUpdatedV1(&ForkchoiceStateV1{
+						HeadBlockHash:      altChainPayloads[i].BlockHash,
+						SafeBlockHash:      altChainPayloads[i].BlockHash,
+						FinalizedBlockHash: altChainPayloads[i].BlockHash,
```
marioevz, line 1204:
```suggestion
						FinalizedBlockHash: common.Hash{},
```

simulators/ethereum/engine/enginetests.go
```diff
@@ -226,108 +227,103 @@ var engineTests = []TestSpec{
...
+		Name:             "Invalid Ancestor Chain Re-Org, Invalid GasLimit, Invalid P8', Reveal using sync",
 		SlotsToFinalized: big.NewInt(20),
-		Run:              invalidMissingAncestorReOrgGen(9, InvalidGasLimit, true, false),
+		TimeoutSeconds:   15,
+		Run:              invalidMissingAncestorReOrgGenSync(8, InvalidGasLimit, false),
 	},
 	{
-		Name:             "Invalid Ancestor Chain Re-Org, Invalid GasUsed, Invalid P9', Reveal using sync",
-		TimeoutSeconds:   15,
+		Name:             "Invalid Ancestor Chain Re-Org, Invalid GasUsed, Invalid P8', Reveal using sync",
```
marioevz, line 275: Why are some `P8'` instead of `P9'`?

simulators/ethereum/engine/enginetests.go
```diff
@@ -1217,108 +1197,222 @@ func invalidMissingAncestorReOrgGen(invalid_index int, payloadField InvalidPaylo
...
-							if err != nil {
-								t.Fatalf("FAIL (%s): Unable to get latest block: %v", t.TestName, err)
-							}
+				// If we are syncing through p2p, we need to keep polling until the client syncs the missing payloads
+				for {
+					r := t.TestEngine.TestEngineNewPayloadV1(altChainPayloads[n])
+					t.Logf("INFO (%s): Response from main client: %v", t.TestName, r.Status)
+					s := t.TestEngine.TestEngineForkchoiceUpdatedV1(&ForkchoiceStateV1{
+						HeadBlockHash:      altChainPayloads[n].BlockHash,
+						SafeBlockHash:      altChainPayloads[n].BlockHash,
```
marioevz, line 1379:
```suggestion
						SafeBlockHash:      cA.BlockHash,
```

simulators/ethereum/engine/enginetests.go
```diff
@@ -1217,108 +1197,222 @@ func invalidMissingAncestorReOrgGen(invalid_index int, payloadField InvalidPaylo
...
+						FinalizedBlockHash: altChainPayloads[n].BlockHash,
```
marioevz, line 1380:
```suggestion
						FinalizedBlockHash: cA.BlockHash,
```

simulators/ethereum/engine/testenv.go
```diff
@@ -115,7 +115,7 @@ func RunTest(testName string, ttd *big.Int, slotsToSafe *big.Int, slotsToFinaliz
 	}()
 
 	// Setup timeouts
-	env.Timeout = time.After(timeout)
+	env.Timeout = time.After(2 * timeout)
```
marioevz, line 118: I think it would be better to increment the timeout only for the tests that require this double value.

marioevz: Also, this timeout was just recently updated in https://github.com/ethereum/hive/commit/7ef78bf4723fc1199d6b276b30b231394aaae4b9

marioevz (commented):
Looks great overall, I think this is more in line with what would happen in real case scenario.
I ran the tests and only Erigon seems to pass these tests. Is a change in geth necessary for them to pass ?

simulators/ethereum/engine/enginetests.go
MariusVanDerWijden: Because I am unable to mine invalid blocks on top of an invalid block on a node. The next block is send via NewPayload. See also description

fjl:
Should we just merge this?

marioevz:
> Should we just merge this?

I think so, these new issues look like genuine issues in some of the clients, I just wanted to share the logs with the clients' devs so they get a sense of what is being changed.

PR #562, opened 2022-06-03
clients/nethermind: add merge support in client configuration

LukaszRozmej:
Removes "merge-nethermind" client configuration.
Makes "nethermind" client configuration merge-compliant and uses latest image.


fjl:
Yes, and I have changed/redeployed hivetests2 to use 'nethermind' instead of 'merge-nethermind'.

PR #567, opened 2022-06-15
simulators/ethereum/engine: Fix Finality On Invalid Payload Attributes Test

marioevz:
On the "ForkchoiceUpdated Invalid Payload Attributes" tests, the forkchoice was set to instant finality and then rolled back on the subsequent ForkchoiceUpdated call, which produced unintended behavior on some clients, which is something the test did not intend to verify.
Fix is to send the same hash of the latest forkchoice for `SafeBlockHead` and `FinalizedBlockHead`.


MariusVanDerWijden (approved):
LGTM

PR #569, opened 2022-06-16
simulators/eth2/engine: Update Proxy, Add Test Cases

marioevz:
This PR is WIP, adds three EL-mock-based testnet test cases:
- Invalid Transition Payload: Validator produces an incorrect (invalid stateRoot) transition payload and generates an invalid beacon block with it, which shall be rejected by the rest of the validators.
- Invalid Payload BlockHash: Validator produces payload with an invalid hash and generates an invalid beacon block with it, which shall be rejected by the rest of the validators.
- Invalid Header PrevRandao: Validator produces a valid payload but with an incorrect PrevRandao value and generates an invalid beacon block with it, which shall be rejected by the rest of the validators.

All three test cases use a new functionality of the proxy that implements callbacks and allows to introduce logic before modifying/corrupting a payload.


marioevz:
Update:
- Added "SYNCING with invalid chain" test case
- Added parameter `--data-storage-mode=ARCHIVE` to Teku, which allows queries on finalized states
- Rebased to master

marioevz:
Update, fixed "syncing-with-invalid-chain" by spoofing `forkchoiceUpdated` to also return `SYNCING`. Test is now passing in at least one client.

marioevz:
Added invalid timestamp test case.

clients/teku-bn/teku_bn.sh
```diff
@@ -51,4 +53,5 @@ fi
     --p2p-udp-port="${HIVE_ETH2_P2P_UDP_PORT:-9000}" \
     --p2p-advertised-ip="${CONTAINER_IP}" \
     --p2p-peer-lower-bound="${HIVE_ETH2_P2P_TARGET_PEERS:-10}" \
-    --rest-api-enabled=true --rest-api-interface=0.0.0.0 --rest-api-port="${HIVE_ETH2_BN_API_PORT:-4000}" --rest-api-host-allowlist="*"
+    --rest-api-enabled=true --rest-api-interface=0.0.0.0 --rest-api-port="${HIVE_ETH2_BN_API_PORT:-4000}" --rest-api-host-allowlist="*" \
+    --data-storage-mode=ARCHIVE
```
zilm13, line 57:
```suggestion
    --data-storage-mode=ARCHIVE \
    --Xstartup-target-peer-count=0    
```

simulators/eth2/engine/running_testnet.go
```diff
@@ -126,12 +199,14 @@ func (t *Testnet) WaitForFinality(ctx context.Context) (common.Checkpoint, error
 					}
 
 					var checkpoints eth2api.FinalityCheckpoints
-					if exists, err := beaconapi.FinalityCheckpoints(ctx, b.API, eth2api.StateIdRoot(headInfo.Header.Message.StateRoot), &checkpoints); err != nil {
-						ch <- res{err: fmt.Errorf("beacon %d: failed to poll finality checkpoint: %v", i, err)}
-						return
-					} else if !exists {
-						ch <- res{err: fmt.Errorf("beacon %d: Expected state for head block", i)}
-						return
+					if exists, err := beaconapi.FinalityCheckpoints(ctx, b.API, eth2api.StateIdRoot(headInfo.Header.Message.StateRoot), &checkpoints); err != nil || !exists {
```
zilm13, line 202: Could we add these changes and few blocks below to `simulators/eth2/testnet/running_testnet.go` so we could pass `eth2/testnet` too? I don't want to just copy and make PR from it but if you want, I can

marioevz: I think it would be better to create another PR to do this to not mix different simulators on the same PR.

marioevz:
@zilm13 I think it would be better to not launch the validator if it has nothing to validate, so I will try this change directly in the simulator.

The reason to include beacon nodes without validators is to verify that they don't follow a chain if their configuration mismatches (e.g. Terminal Total Difficulty), even if the chain has 66%+ participation.

marioevz:
Hi @fjl, I think this PR should be ready to merge with a few considerations:
- Nimbus will fail to start until https://github.com/protolambda/zrnt/pull/38 is merged and we update.
- All test cases require https://github.com/rauljordan/engine-proxy/pull/5, therefore `engine-proxy` currently points to my branch at https://github.com/marioevz/engine-proxy/tree/request-bytes-callback until the PR is merged (otherwise none of the test cases would work).
- Lodestar shows networking issues, but I think this needs to be addressed in https://github.com/ethereum/hive/pull/577

I think these changes can be part of new PRs to keep this PR from further growing.
Let me know what you think.

PR #572, opened 2022-06-18
rpc-compat: new simulator

lightclient:
Not many tests in this simulator yet, but I think it is ready for review. 

The way the new `rpc-compat` simulator works is it clones https://github.com/lightclient/rpctests into the simulator container, walks the `tests` directory  loading tests, executes the test by sending a request (denoted `<<` in the test file) to the hive client, and finally, verifies that the response matches the test response (denoted `>>` in the test file).

It does a round trip unmarshal/marshal of the response value (both received and expected) in order to eliminate any padding or sorting differences.


fjl:
I think this is a very solid start. Will deploy on the prod instance soon!

PR #573, opened 2022-06-18
all: remove openethereum

lightclient:
Openethereum has been [deprecated](https://twitter.com/OpenEthereumOrg/status/1529147048758595585?s=20&t=QvbUmyLilOYBmGkeFGBmeQ) and is no longer maintained. Officially, London was the last hard fork it supported but because Arrow Glacier was just a difficulty adjustment it continued on a bit longer.


fjl:
I would prefer to keep this client definition for now. It is working still.

PR #577, opened 2022-06-22
clients/lodestar: add Lodestar Beacon Node and Validator Client

marioevz:
Adds Lodestar client for hive consensus layer tests.

ToDos:
- Intermittently, the beacon nodes are unable to peer with each other.
- Method `VerifyProposers` uses `/eth/v1/beacon/states/{slot}/validators/{index}` to get the validator for a certain slot, but this method fails.


marioevz:
Superseded by #680

PR #581, opened 2022-06-28
clients/prysm: add Prysm Beacon Node and Validator Client 

marioevz:
Adds Prysm beacon node and validator client images.

A few considerations:
- Simulator requires that Altair and Bellatrix fork on different epochs (e.g. Altair on epoch 1, and bellatrix on epoch 2) which results in long test case runtimes.
- Requires changes to all FORK_VERSION configurations because Prysm does not admit testnet identifiers to have the same versions as mainnet.
- Prysm's REST API endpoint `/eth/v2/debug/beacon/states/` returns error 404. This endpoint is used by method `getHealth` and therefore it fails, which might require a workaround.

These considerations aside, the client is functional and most of the test cases in https://github.com/ethereum/hive/pull/569 are passing.

### DIFF
--- a/clients/prysm-bn/Dockerfile
+++ b/clients/prysm-bn/Dockerfile
@@ -1,9 +1,12 @@
-ARG branch=latest-alpine
-FROM gcr.io/prysmaticlabs/prysm/beacon-chain:$branch
+ARG branch=latest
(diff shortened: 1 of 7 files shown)


marioevz:
@fjl I think this is ready for merge, I've been running tests on the `eth2/engine` sim and most are passing.
Will merge once I finish running the latest version of the sim locally.

PR #582, opened 2022-06-29
clients/erigon: add -experimental.overlay flag

Giulio2002:


marioevz (approved):

fjl:
Got the answer on Discord:

> it adds support to erigon for side forks 
> also it optimizes the json rpc
> ...
> Specifically, it's an experimental in-memory block execution that reduces the number of cases when Erigon returns ACCEPTED.

PR #583, opened 2022-06-30
clients/besu: update sync mode

daniellehrner:
Signed-off-by: Daniel Lehrner <daniel.lehrner@consensys.net>

The `rpc` test suite for Besu is failing a lot of tests, because the client start with fast sync instead of full sync and gets stuck. This PR fixes the configuration of the sync mode for all existing test suites.

We cannot only rely on `HIVE_NODETYPE` for it, but need to take `HIVE_TERMINAL_TOTAL_DIFFICULTY` into consideration as well, because the `engine` test suite needs full sync (at least for Besu), but `HIVE_NODETYPE` is empty.

### DIFF
--- a/clients/besu/besu.sh
+++ b/clients/besu/besu.sh
@@ -124,13 +124,15 @@ else
     FLAGS="$FLAGS --network-id=1337"
 fi
 
-# Sync mode, must only be changed for non merge related tests.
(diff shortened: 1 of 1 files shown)


fjl:
Thanks! I will deploy this change soon.

PR #586, opened 2022-07-04
clients/go-ethereum: set syncmode=full by default

MariusVanDerWijden:
We changed something in geth that makes a lot of tests fail in snap sync mode.
Basically you can not import blocks one by one in snap sync mode, the snap sync has to finish before importing blocks one by one, otherwise we might overwrite data that is already in the database. 
The tests depend on this one-by-one importing though, which is only available to a full synced node in geth. This changes the default sync mode to full.


holiman:
IIUC, this PR made it so that our sync-test now uses a full-sync for testing. So we now have no test that actually tests whether geth can default-sync (snap) from geth.

PR #588, opened 2022-07-04
M1 mac support

mandrigin:
Support launching hive tests on macOS (both Intel and M1).

For M1 platforms, the Docker image of a client should have arm64 binaries.

E.g. for Erigon, you build it locally on M1 mac

(1) `docker build . -t thorax/erigon:local`

(2) `go run . --client erigon_local --sim ethereum/engine`


mandrigin:
will keep that in my own fork for now

PR #594, opened 2022-07-05
Fix reorgBack: send back FCU only after payload verification

yperbasis:
This improves "Re-Org Back into Canonical Chain" test. The FCU pointing to the previous block sent by `reorgBack` should be sent after the payload verification in `produceSingleBlock`. When the back FCU is sent too early, it might render the newly created block unaccessible by `eth_getBlockByNumber("latest")`, thus leading to "CLMocker: None of the clients accepted the newly constructed payload".

### DIFF
--- a/simulators/ethereum/engine/clmock.go
+++ b/simulators/ethereum/engine/clmock.go
@@ -341,7 +341,7 @@ type BlockProcessCallbacks struct {
 	OnForkchoiceBroadcast     func()
 	OnSafeBlockChange         func()
 	OnFinalizedBlockChange    func()
-	OnVerificationEnd         func()
+	OnBlockProductionEnd      func()
 }
 
 func (cl *CLMocker) produceSingleBlock(callbacks BlockProcessCallbacks) {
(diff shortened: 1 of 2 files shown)


simulators/ethereum/engine/clmock.go
```diff
@@ -341,6 +341,7 @@ type BlockProcessCallbacks struct {
 	OnForkchoiceBroadcast     func()
 	OnSafeBlockChange         func()
 	OnFinalizedBlockChange    func()
+	OnVerificationEnd         func()
```
marioevz, line 344: This addition seems fine to me, but perhaps the name is not accurate. This is more like OnBlockProductionEnd.

marioevz (approved):

yperbasis:
I've realised that my previous reasoning was wrong because of the following rule from the [Engine API spec](https://github.com/ethereum/execution-apis/blob/main/src/engine/specification.md):

Client software **MAY** skip an update of the forkchoice state and **MUST NOT** begin a payload build process if `forkchoiceState.headBlockHash` references an ancestor of the head of canonical chain. In the case of such an event, client software **MUST** return `{payloadStatus: {status: VALID, latestValidHash: forkchoiceState.headBlockHash, validationError: null}, payloadId: null}`.

PR #595, opened 2022-07-06
clients/besu: implement retry mechanism to get enode

### LINKED ISSUE #593: Unable to obtain bootnode: invalid URL scheme, want "enode"
At least for Besu, there are some tests that do not start and the simulator log this error:

```
FAIL (Invalid Ancestor Chain Re-Org, Invalid PrevRandao, Invalid P9', Reveal using sync): Unable to obtain bootnode: invalid URL scheme, want "enode"
```
Source: https://hivetests2.ethdevops.io/?page=v-pills-results-tab&suite=1657032249-2075290de9cb4eff7492fe284a5f5951.json

This seems to be related to the simulator asking something and Besu not always ready to respond, any idea if it is possible to tune it in the Besu client script, or it is something that needs to be changed in the test suite?

fab-10:
Implement a simple retry strategy when getting the enode URL from Besu, since it could not be yet ready to serve that information on the first call.
 
fixes #593

### DIFF
--- a/clients/besu/enode.sh
+++ b/clients/besu/enode.sh
@@ -13,6 +13,7 @@ data='{"jsonrpc":"2.0","method":"admin_nodeInfo","params":[],"id":1}'
 while [[ "$TARGET_RESPONSE" != *enode* ]]
 do
     TARGET_RESPONSE=$(curl -s -X POST  -H "Content-Type: application/json"  --data $data "localhost:8545" )
+    ((c++)) && ((c==50)) && break
     sleep 0.1
 done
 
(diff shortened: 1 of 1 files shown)


clients/besu/enode.sh
```diff
@@ -10,7 +10,11 @@
 set -e
 
 data='{"jsonrpc":"2.0","method":"admin_nodeInfo","params":[],"id":1}'
-TARGET_RESPONSE=$(curl -s -X POST  -H "Content-Type: application/json"  --data $data "localhost:8545" )
+while [[ "$TARGET_RESPONSE" != *enode* ]]
+do
+    TARGET_RESPONSE=$(curl -s -X POST  -H "Content-Type: application/json"  --data $data "localhost:8545" )
+    sleep 0.1
```
marioevz, line 16:
```suggestion
    ((c++)) && ((c==50)) && break
    sleep 0.1
```
We could add a max retry count here, perhaps 5 seconds ?

PR #604, opened 2022-07-13
clients/lighthouse-bn: Add Optimistic Sync Options + Trusted Peers with Static Keys

marioevz:
Adds client support for `HIVE_ETH2_SAFE_SLOTS_TO_IMPORT_OPTIMISTICALLY` setting:
- `SAFE_SLOTS_TO_IMPORT_OPTIMISTICALLY` is set in `config.yaml`
- `--safe-slots-to-import-optimistically` is set as parameter to the same value

### DIFF
--- a/clients/lighthouse-bn/Dockerfile
+++ b/clients/lighthouse-bn/Dockerfile
@@ -8,6 +8,8 @@ RUN chmod +x /lighthouse_bn.sh
 ADD lighthouse_version.sh /lighthouse_version.sh
 RUN chmod +x /lighthouse_version.sh
 
+RUN apt-get update -y && apt-get install -y xxd
+
 RUN /lighthouse_version.sh > /version.txt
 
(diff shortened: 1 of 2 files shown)


marioevz:
Hi @pawanjay176, I added static keys for at least 5 nodes.
Let me know if this looks ok, thanks!

pawanjay176:
@marioevz we are very close to merging a PR (https://github.com/sigp/lighthouse/pull/3428) which might make the `trusted-peers` hack redundant. I'm trying to test out the optimistic sync hive test cases with that PR but the tests are ending prematurely for some reason.

I'll try to fix it up locally and let you know if the static peers is indeed required.

marioevz:
Hey @pawanjay176, I guess it wouldn't hurt to have this fix merged and then remove it in a subsequent PR if necessary.
Could we go ahead and merge since this is required for another PR (#637) ?
Thanks.

pawanjay176:
Yep, sgtm.

PR #605, opened 2022-07-13
clients/teku-bn: Add SAFE_SLOTS_TO_IMPORT_OPTIMISTICALLY

marioevz:
Adds client support for `HIVE_ETH2_SAFE_SLOTS_TO_IMPORT_OPTIMISTICALLY` setting:
- `SAFE_SLOTS_TO_IMPORT_OPTIMISTICALLY` is set in `config.yaml`
- `--Xnetwork-safe-slots-to-import-optimistically` is set as parameter to the same value


marioevz:
Confirmed to work by running #606. Merging.

PR #607, opened 2022-07-15
clients/lighthouse-bn,clients/lighthouse-vc: add version info

michaelsproul:
Add a script to pull out the Lighthouse version and store it in `version.txt`


fjl:
Thank you!

PR #610, opened 2022-07-15
libhive, libdocker: Add client multiple port check

marioevz:
`HIVE_CHECK_LIVE_PORT` can now be parsed as a comma-separated string of multiple ports to check for liveness of a client.


fjl:
Not entirely sure about this. The port check is mostly for hive to verify that the client has started. If you need a check for another port, you could just do it from the simulator instead.

marioevz:
@fjl Is it possible to perform a check-live-port from inside of the simulator?

marioevz:
Will fix directly in the simulator.

PR #613, opened 2022-07-19
simulators/ethereum/engine: Major Refactoring, Gossip Test Cases

marioevz:
## Changes Included
- Refactor client implementation to be able to plug the Hacked Geth Node into the CL Mocker and introduce the module into the transition test cases
- Refactored the entire simulator into a multiple package go project, to make it more readable and maintainable.
- Changed engine api `Out of Order Payloads` test case into `Consecutive Payload Execution` because not all clients cache received new payloads, and sending retroactive payloads is a behavior the CL clients don't do.
- Added PoW gossip transition test cases with the aid of the Hacked Geth Node module
- Updated README to better reflect behavior of existing test cases
- DockerFile now produces PoW DAG using geth in order to mine blocks during runtime for some of the new test cases
- CL Mocker can now be configured to mimic behavior of the `SAFE_SLOTS_TO_IMPORT_OPTIMISTICALLY` of some CL clients
- Transition/Merge test cases now differentiate between a case where the client is expected to immediately return a `VALID` or `INVALID` value to a `newPayload` or the client is expected to first return `SYNCING` until a resolution is reached

This PR was previously meant to finish TestingTheMerge missing test cases, but these will be incorporated into a new PR due to the already big size of this PR, and the fact that includes new test cases breaking in some of the clients which are necessary for debugging and should be merged as soon as possible.

Total EL not passing test cases:
- Geth total 0, new 0
- Besu total 1, new 1:
  - Build Payload After Multiple Terminal blocks via gossip
- Erigon total 6, new 4:
  - Consecutive Payload Execution
  - Two Block PoW Re-org to Lower-Height Chain, Transaction Overwrite
  - Stop processing gossiped Post-TTD PoW blocks
  - Build Payload After Multiple Terminal blocks via gossip
- Nethermind total 18, new 0


marioevz:
I ran all test suites with all clients, and it seems like the previous version of the `sync` tests was incorrect: The tests were incorrectly passing even when the second client had not synced.

@MarekM25 it seems like nethermind was directly affected by this.

Will be merging asap to have the tests correct version running.

PR #621, opened 2022-08-01
Do not set finalized blockhash on invalid missing ancestor reorg via sync test

asdacap:
- Nethermind removes peers that does not have the finalized blockhash.
- This causes this test to fail because the secondary client does not have the final block which is also set as the finalized blockhash.
- Finalized blockhash are also suppose to be at least 32 block behind head AFAIK.

### DIFF
--- a/simulators/ethereum/engine/suites/engine/tests.go
+++ b/simulators/ethereum/engine/suites/engine/tests.go
@@ -1479,7 +1479,7 @@ func invalidMissingAncestorReOrgGenSync(invalid_index int, payloadField helper.I
 						status2, err := secondaryClient.ForkchoiceUpdatedV1(ctx, &api.ForkchoiceStateV1{
 							HeadBlockHash:      altChainPayloads[i].BlockHash,
 							SafeBlockHash:      cA.BlockHash,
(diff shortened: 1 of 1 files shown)


mkalinin:
> Finalized blockhash are also suppose to be at least 32 block behind head AFAIK.

Strictly speaking it is 64-95 blocks behind the head in the best case. Finalized block is always the first block of the finalized epoch if the first slot is non-empty, finalized epoch is `curr_epoch - 2` in the best case.

> Nethermind removes peers that does not have the finalized blockhash

@MarekM25 Is this only for the merge transition case? I am not sure if EL should do anything with the finalized block hash beyond serving it to a user. We have been touching this topic by a "should EL reject newPayloads with payloads that are conflicting with finalzied payload?", and the answer for this question is strictly no.

I am not sure about networking though. Could it cause any peering issues in the case when finalized checkpoing was updated but for some reason a remote peer is yet behind it?

LukaszRozmej:
@mkalinin we do it only during **our** sync, so we don't have potentially useless peers for sync (either from different forks or some that are far behind themselves). This helps us a lot on shadow-fork sync, but should be useful even later.

marioevz (approved):

PR #622, opened 2022-08-01
simulators/ethereum/engine: Fix Gossip Test Verification

marioevz:
### Changes Included

- Change TTD values of gossip tests to make the height of the TTD-reaching block more consistent
- Fix verification of the gossiped block count
- Add tests where the height difference between the common ancestor and the re-org head is greater than 1
- Verify that the node actually mines the expected number of terminal blocks
- Directly use ethash to mine the blocks instead of relying on the miner module; This is done in order to not be affected by some clients that send back a newly mined block to the client where they originally received it from, which caused an internal re-org that messed up with the mining of several terminal blocks

### Outstanding issues

~Some clients are sending back the newly mined block as NewBlockMsg to the `PoW Producer`.
This results in the `PoW Producer` thinking it's out of sync after we re-org back to a lower height block, and therefore abort mining.
Therefore, the `PoW Producer` never mines the expected amount of terminal blocks and the test is invalid.
Filed https://github.com/ethereum/hive/issues/623 for this.~

### DIFF
--- a/simulators/ethereum/engine/Dockerfile
+++ b/simulators/ethereum/engine/Dockerfile
@@ -2,8 +2,10 @@
 # Use a static version because this will never need to be updated.
 FROM ethereum/client-go:v1.10.20 AS geth
 RUN \
- /usr/local/bin/geth makecache  1 /ethash && \
- /usr/local/bin/geth makedag    1 /ethash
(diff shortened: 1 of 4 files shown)


marioevz:
Will merge after running the simulator with all clients and verifying that no test issues are outstanding.

PR #627, opened 2022-08-05
simulators/ethereum/engine: Remaining Tests for Execution Client

marioevz:
### Changes Included

- Introduce tests where the client is directed (via `newPayload`) to sync to a chain which contains a payload with uncles, which is invalid for post-Merge:
  - On the transition payload, the uncle is another terminal PoW block
  - On a post-transition payload, the uncle is another post-transition payload
- Introduce tests where the client is directed (via `newPayload`) to sync to a chain which contains an invalid transition payload


mkalinin (approved):
LGTM 👍

PR #630, opened 2022-08-09
rpc-compat: add special test for `debug_getBadBlocks`

lightclient:
Since `debug_getBadBlocks` relies on the client seeing a bad block over p2p, I've [added support](https://github.com/ethereum/go-ethereum/pull/25212) in geth to send arbitrary new block announcements and have sealed a "bad" block in the execution-apis repo, along side the normal `chain.rlp`.

This PR takes the provided bad block and sends it over the wire to the client. The client should then reject the block as invalid and record it somewhere to later be retrieved by `debug_getBadBlocks`.

### DIFF
--- a/clients/go-ethereum/Dockerfile
+++ b/clients/go-ethereum/Dockerfile
@@ -4,29 +4,29 @@
 # produce a very minimalistic container that can be reused many times without
 # needing to constantly rebuild.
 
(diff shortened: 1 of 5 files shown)


fjl:
I'm closing this because the p2p protocol no longer allows injecting block announcements.

PR #633, opened 2022-08-11
simulators/ethereum/engine: Randomize Coinbase on Mined Terminal Blocks

marioevz:
### Changes Included

- Tests that mine PoW blocks during runtime now will get a different coinbase each time to be able to test multiple terminal block with different stateRoots


mkalinin:
Won't the state root for PoW blocks be different if the `coinbase` address remain the same?

But that would not help for the PoS blocks. There should be some transaction included in every block if we want their state root to be changed

PR #634, opened 2022-08-11
simulator/ethereum/engine: Fix Transaction Already Known Issue

marioevz:
### Changes Included

- Fixes error where a transaction included in a pre-mined chain could be re-org'd back into the txPool, and when sent again by the simulator it resulted in "already known" error on `eth_sendRawTransaction`


marioevz:
Hi @Giulio2002, this proposed change fixes the test where we were getting the `already known` for the transaction that was unwound back into the txPool. Let me know if this is ok.

PR #636, opened 2022-08-17
clients/lighthouse-bn,teku-bn,prysm-bn,nimbus-bn: Add Shutdown Scripts

marioevz:
### Changes Included
- Adds shutdown scripts for each supported beacon client, which send `SIGINT` to the running client in order to shutdown a node mid-test
- Script `shutdown.sh` can be called using `hivesim.Client.Exec` at any point during the test


fjl:
I don't understand why you'd want to run this script. If you want to terminate a client, you can [just use StopClient](https://pkg.go.dev/github.com/ethereum/hive/hivesim#Simulation.StopClient).

marioevz:
Yes, that should totally do it. I thought I might be missing something, I was looking for this functionality in the hiveclient construct. Thanks!

PR #638, opened 2022-08-18
simulators/ethereum/engine: Syncing on an Invalid Terminal Block (Invalid Execution)

marioevz:
## Changes Included
- Introduce tests `Syncing on an Invalid Terminal Execution`, described here: https://github.com/txrx-research/TestingTheMerge/blob/main/tests/transition.md (`Syncing with the chain where terminal block is invalid with respect to EE`), where the client is delivered a `newPayload(P1)` and `forkchoiceUpdated(P1)`, and `P1` is built on top of an unknown terminal block that reaches the Terminal Total Difficulty but it's an invalid block due to one of these characteristics
  - `difficulty` number in the block header is incorrect according to the ethash consensus mechanism
  - Sealed `mixHash` in the block header is incorrect according to the ethash consensus mechanism
  - Sealed `nonce` in the block header is incorrect according to the ethash consensus mechanism

Please suggest further modes of invalidation for the terminal block.

cc @mkalinin

### DIFF
--- a/simulators/ethereum/engine/client/node/node.go
+++ b/simulators/ethereum/engine/client/node/node.go
@@ -433,7 +433,7 @@ func (n *GethNode) PoWMiningLoop() {
 
 		// Modify the block before sealing
 		if n.config.BlockModifier != nil {
-			b, err = n.config.BlockModifier.ModifyUnsealedBlock(b)
+			b, err = n.config.BlockModifier.ModifyUnsealedBlock(n.eth.BlockChain(), s, b)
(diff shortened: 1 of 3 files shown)


mkalinin:
May invalid `stateRoot` work as well?

mkalinin (approved):

marioevz:
> May invalid `stateRoot` work as well?

Yes, the last commit added such test by increasing the balance of a random account to produce and invalid stateRoot.

PR #641, opened 2022-08-25
simulators/eth2/testnet: Add bn multiaddrs in $HIVE_ETH2_STATIC_PEERS

zilm13:
It's actually 
`simulators/eth2/testnet: Add bn multiaddrs in $HIVE_ETH2_STATIC_PEERS`
and 
`clients/teku: Use $HIVE_ETH2_STATIC_PEERS for persistent BN connection`
I could split it to 2 separate PRs if you approve API changes. And it would be good to get same changes in `engine` too.
Btw, I'm not sure that first address from `P2PAddresses` is always good, maybe we should push them all?

### DIFF
--- a/clients/lighthouse-bn/Dockerfile
+++ b/clients/lighthouse-bn/Dockerfile
@@ -8,6 +8,8 @@ RUN chmod +x /lighthouse_bn.sh
 ADD lighthouse_version.sh /lighthouse_version.sh
 RUN chmod +x /lighthouse_version.sh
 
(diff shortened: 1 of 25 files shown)


marioevz:
LGTM, I added a PR on your branch to also bring the fix to the `eth2/engine`, because I think this fixes `no-viable-head-due-to-optimistic-sync` for Teku on that sim.

marioevz (approved):
LGTM

PR #645, opened 2022-08-31
simulators/ethereum/sync: test all sync modes

MariusVanDerWijden:
Adds tests for all sync modes


fjl:
This works, but it will run a lot of pointless tests because most clients do not support all those syncmodes.

PR #648, opened 2022-09-04
simulators/ethereum/engine: Invalid Transactions Test

marioevz:
## Changes Included
- Add test case to verify that invalid transactions received using `eth_sendRawTransaction` are not included in payloads built


fjl:
@marioevz was this change also a part of #659?

PR #650, opened 2022-09-09
Added debug module in enabled modules

DemuirGos:
Changes Included
Add Enabled Modules explicitly to include previously disabled modules like the Debug Module (for Nethermind)


holiman (commented):
LGTM

PR #652, opened 2022-09-09
simulators/eth2/engine: Fix Test Issues

marioevz:
## Changes Included
- Add subnets of execution clients: Tests will now be able to limit some execution clients from seeing each other. This will place some of the clients in `SYNCING` mode until, e.g., an `INVALID` response is mocked by the engine proxy.
- Test RPC Error test now expects a lower participation, and also waits for an execution payload to be finalized before actually starting the test (Fixes the test for prysm).
- Re-Org Sync test has been reworked to have a simpler and more deterministic approach:
  - Two client pairs are created, each containing a builder and an importer, and their execution clients within the pair will be able to communicate with each other, but not with the execution clients of the other pair.
  - Each pair has a different starting PoW chain, and since they cannot exchange information to sync with the other clients, one pair will eventually and certainly end up in optimistic sync mode.
  - The test then detects which pair ended up in optimistic sync mode (`execution_optimistic==true`), which means that this pair did not produce the transition payload, and hence is not able to sync the chain to keep building and producing blocks.
  - The response is then mocked for the optimistic sync pair to `INVALID`, which should result in the pair building its own separate chain.


holiman:
@marioevz sorry this seems to have been missed -- do you still want this in?

marioevz:
> @marioevz sorry this seems to have been missed -- do you still want this in?

Yes! I forgot about it too, but it would be ok to merge it :+1:

marioevz:
Updated to latest branch and reran. Works ok, merging.

PR #658, opened 2022-11-15
clients/erigon: Add externalcl flag

yperbasis:
After https://github.com/ledgerwatch/erigon/pull/5813 Erigon uses embedded light-client CL by default. That mode is not compatible with Hive tests and erigon should be switched back to using external CL.


marioevz (approved):
LGTM

PR #659, opened 2022-11-19
simulators/ethereum/engine: Withdrawals Test Suite

marioevz:
Draft of withdrawals tests.

At the moment it simply makes the required changes to the CLMocker to be able to send the new versions of the Engine API methods that support withdrawals.

More test cases need to be added.


marioevz:
Build for `cmd/hivechain` seems to be failing because `simulators/ethereum/engine`'s `go.mod` replaces `github.com/ethereum/go-ethereum` with `github.com/lightclient/go-ethereum@withdrawals-timestamp` due to the new execution engine payload format, and also `types.Block` and `types.Header` contain withdrawals.

This doesn't seem right, I'm not sure why a replace within a submodule could affect the root module in a workspace, do you know if this behavior is expected @fjl ? Any pointers would be appreciated, since I've tried many workarounds to this error and I could not get it fixed.

fjl:
This was merged in #676

PR #660, opened 2022-11-23
internal/libhive: fix assignment to entry in nil map bug

ogenev:
When the "environment" field is not included in the client launch configuration,  `env := clientConfig.Environment`(line 224) returns nil and a panic  `assignment to entry in nil map` is raised when we do  `env["HIVE_LOGLEVEL"] = strconv.Itoa(api.env.SimLogLevel)` later (line 236).


fjl:
Thanks!

PR #661, opened 2022-11-27
clients/besu: update config to address rpc and send-legacy-transaction rpc-compat failure

mark-terry:
Signed-off-by: mark-terry <mark.terry@consensys.net>

### DIFF
--- a/clients/besu/Dockerfile
+++ b/clients/besu/Dockerfile
@@ -20,7 +20,10 @@ RUN chmod +x /hive-bin/enode.sh
 
 RUN ./bin/besu --version > /version.txt
 
-# Export the usual networking ports to allow outside access to the node
-EXPOSE 5005 8545 8546 8551 30303 30303/udp
+# Expose JVM debugger port.
+EXPOSE 5005
+
+# Expose the usual networking ports to allow outside access to the node.
+EXPOSE 8545 8546 8551 30303 30303/udp
 
 ENTRYPOINT ["/opt/besu/bin/besu-hive.sh"]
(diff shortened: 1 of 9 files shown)


fjl:
Thanks for finding the fix. I think we should not add a new environment variable for this option, and instead simply set this flag for besu unconditionally.

mark-terry:
> Thanks for finding the fix. I think we should not add a new environment variable for this option, and instead simply set this flag for besu unconditionally.

My only hesitation with that approach is that it'll prevent testing any base-fee related behaviour; which didn't seem to be a concern with rpc-compat, but might be relevant in other simulations. Any feedback on that?

fjl:
I'm just wondering how the RPC tests can pass in other clients with no extra configuration. Could you explain a little bit what this flag does?

mark-terry:
> I'm just wondering how the RPC tests can pass in other clients with no extra configuration. Could you explain a little bit what this flag does in besu?

Certainly. These params were updated:

- `--min-gas-price=1` - the transaction in the simulation has a `gasPrice` of 1, so the previous `min-gas-price=16` value was causing the transaction to be rejected.
- `--tx-pool-limit-by-account-percentage=1"`- this option increases the number of transactions permitted in the transaction pool from a single sender; which is appropriate for a test suite.

Genesis config:
`zeroBaseFee` - this option configures free gas on networks with London+ forks and removes the 7 wei base fee; which is higher than the transaction's `gasPrice`.

mark-terry:
@fjl Were you waiting on anything from my side?

fjl:
Sorry, I was on vacation and didn't get to merge it before that.

fjl:
I have resolved the issue with baseFee being too low by updating the tests instead: https://github.com/ethereum/execution-apis/pull/347

fjl:
I will merge this now. The legacy tx test will start passing when tests are updated upstream.

PR #662, opened 2022-11-27
internal/libdocker: add authentication options for docker client

mrod502:
Add config options to:
- specify authentication type through `docker.authentication` (currently only empty (none) and cred-helper are supported)
- specify which docker registries to load credentials for

This allows users to run tests against private docker images in cases where they may not want a pre-prod image to be exposed publicly.

Example usage (in a circleci build pipeline):
```yaml
steps:
      - run:
          name: authenticate with GCP
          command: |
            gcloud auth configure-docker us-central1-docker.pkg.dev
      - run:
          command: |
            ./hive \
            -sim=<<parameters.sim>> \
            -sim.loglevel=5 \
            -docker.pull=true \
            -docker.cred-helper=true \

 -client=go-ethereum,private-image_<<parameters.version>> |& tee /tmp/hive.log || echo "failed."
```

Github issue: https://github.com/ethereum/hive/issues/663


internal/libdocker/builder.go
```diff
@@ -57,15 +63,26 @@ func (b *Builder) BuildClientImage(ctx context.Context, name string) (string, er
...
+	err := b.buildImage(ctx, dir, "Dockerfile", branch, tag)
 	return tag, err
 }
 
 // BuildSimulatorImage builds a docker image of a simulator.
 func (b *Builder) BuildSimulatorImage(ctx context.Context, name string) (string, error) {
 	dir := b.config.Inventory.SimulatorDirectory(name)
+	buildContextPath := dir
+	buildDockerfile := "Dockerfile"
+	// build context dir of simulator can be overridden with "context.txt" file containing the desired build path
```
fjl, line 75: What's the purpose of this feature?

mrod502: This is out of scope for this PR, will remove.

fjl: In scope or not, I'm still curious what this is about :)

mrod502: So initially, I was working on this change on a fork of the `ethereum-optimism/hive` fork of this repo, where this change exists - see commit [here](https://github.com/ethereum-optimism/hive/commit/656d759a1f302425274389b712316a0eec7f5995).
From what I can tell, this allows the user to add a custom relative path to use for build context. It seems to be useful when you want to build a simulator from a `Dockerfile` in an arbitrary dir, or when you have a `Dockerfile.<module>` naming convention.

hive.go
```diff
@@ -69,10 +71,14 @@ func main() {
...
+	dockerRegistries := strings.Split(strings.Trim(*dockerRegistriesRaw, " "), ",")
+
 	// Create the docker backends.
 	dockerConfig := &libdocker.Config{
-		Inventory:   inv,
-		PullEnabled: *dockerPull,
+		Inventory:        inv,
+		PullEnabled:      *dockerPull,
+		DockerRegistries: dockerRegistries,
+		AuthType:         libdocker.AuthType(*dockerAuthType),
```
fjl, line 81: Introducing the `AuthType` ends up creating a lot of boilerplate code. What you want to achieve is using the `cred-helper` if it is defined.

Let's just simplify this whole setting into a boolean `UseCredHelper` (and flag `--docker.cred-helper`) which will configure the authentication for all registries when set.

fjl (changes requested):

internal/libdocker/authenticator.go
```diff
@@ -0,0 +1,97 @@
...
+			return docker.AuthConfigurations{
+				Configs: authConfigurations,
+			}, err
+		}
+		authConfigurations[registry] = *authConfig
+	}
+	return docker.AuthConfigurations{Configs: authConfigurations}, nil
+}
+
+func loadCredsHelpers() (c CredHelpers, err error) {
```
mrod502, line 78: Based on the `go-dockerclient` code, it appears this is the only way to load cred helper configs without having a command-line option for registry.

fjl (approved):

PR #664, opened 2022-12-10
all: Go workspace support and dependency sync

protolambda:
This PR is focused at making cross-module development in Hive more enjoyable: we can add new simulators for EIPs and hive-testnets much more easily if we can share setup code & run a debugger throughout the shared code and simulators with ease.

[Go workspaces](https://go.dev/blog/get-familiar-with-workspaces) fixed this in Go 1.18 to make development across multiple related go modules much easier. This PR adds a `go.work` to the repo that links all modules together, to make development between them easier.

With this, we can follow up with PRs that deduplicate the eth2 testnet code of `simulators/eth2/engine` and `simulators/eth2/testnet` and more future simulators :).

To optionally use the go workspace linking within docker too, a `hive_context.txt` can now be used to change the build context of a simulator. See `docs/simulators.md#simulators-with-shared-modules` for docs and example of how this works.
Note that the docker builds will still use the exact `go.mod` dependencies if not using a changed build context & go.work. If we ever need to pin a simulator to an older version of Hive or other modules, we can do so by just using the `go.mod` in isolation.

Additionally, to make cross-module development easier, I updated *all* dependencies to the latest version. I found a Go 1.13 simulator, different old versions of geth, and other dependency issues like the older conflicting btcd versions. And updated all Go modules to Go 1.18. To start sharing more code between these simulators, syncing the dependencies is a good start. And with `go work sync` it'll be easy to keep in sync going forward.

Note for reviewing:
- the first commit changes the workspace/build context functionality.
- the second commit updates the go mod dependencies. +2000 -8000 in sum-file updates, a much needed tidy&sync!


fjl:
I think `github.com/influxdata/influxdb/client` was upgraded a bit too much in this PR. I have trouble building the new version because it has an additional C library dependency. There is also a Go compile error related to generics.

PR #667, opened 2022-12-19
simulators/eth2: Common Library

marioevz:
Coalesce common code between `eth2` simulators into a library that allows code reuse.


charlestonxorg:
Fuckin send it

marioevz:
Updated to use latest geth version, which requires `go.mod replace`.

Changing to draft because merging would break other simulators until either:
- Withdrawals' changes are merged
- We use `go.mod replace` on the modules/simulators where the replace is breaking stuff.

fjl:
These changes were merged as part of #681

PR #674, opened 2023-01-19
clients/besu: add Shanghai fork and fix issue with duplicated node keys

gfukushima:
This fixes the tests where hive spins up a second instance of besu and sync from the bootnode. Currently node 1 and 2 have the same enode address which stops them from peering with each other 

Linked to [#4747](https://github.com/hyperledger/besu/issues/4747)

### DIFF
--- a/clients/besu/besu.sh
+++ b/clients/besu/besu.sh
@@ -99,14 +99,15 @@ if [ -d /blocks ]; then
     IMPORTFLAGS="$IMPORTFLAGS $blocks"
 fi
 
-
-# Configure mining.
-if [ "$HIVE_MINER" != "" ]; then
-    FLAGS="$FLAGS --miner-enabled --miner-coinbase=$HIVE_MINER"
-    # For clique mining, besu uses the node key as the block signing key.
-    if [ "$HIVE_CLIQUE_PRIVATEKEY" != "" ]; then
-        echo "Importing clique signing key as node key..."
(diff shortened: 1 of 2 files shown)


fjl:
Is the nodekey deterministic in besu?

gfukushima:
If you place a key in the data directory besu will read that and use it to generate the node id. If you use the same key for the bootnode and node 2 (usually used in hive to test the sync) they will have the same node address and node id. Which eventually will prevent these nodes from peering with each other.
https://besu.hyperledger.org/en/stable/public-networks/concepts/node-keys/#node-public-key

fjl:
I understand. AFAIK, hive does not place a key in the data directory, and besu should generate a random key by itself.
If hive configures the same node key for all instances, that's a bug we should fix.

gfukushima:
@fjl my understanding is that this is exactly what this line is doing 
https://github.com/ethereum/hive/blob/6152200ada55a957f95955351738453b4101b6d8/clients/besu/besu.sh#L109

gfukushima:
@fjl  is there anything this PR is missing to be accepted? Right now everyone who runs the hive withdrawals suite need to add those changes manually for Besu

fjl:
Sorry, I just wanted to confirm why the nodekey is used in clique mining mode. But it seems this is not easy to fix in besu, so we can apply this workaround for now.

PR #682, opened 2023-01-23
feat(erigon): pass shanghai override

revitteth:
image

With teku-bn,teku-vc,erigon.


fjl:
Can't we pass this via the genesis config instead? Using fork override flags is discouraged unless there really is no other way of configuring it.

fjl:
Hmm, we did add it to the genesis in https://github.com/ethereum/hive/commit/108450632418738088c26decbc3b46e84e67c226, but maybe it's not recognized?

revitteth:
@fjl I have provided a subsequent update - which I think should be ok (tests pass for erigon).

PR #695, opened 2023-01-26
Add support for custom genesis delay in start

nazarhussain:
During the testing of withdrawals it's observed that a few slots skipped during the very start (at least locally on dev machine). Setting up a genesis delay makes the tests more stable don't fail because a few slots skipped earlier.

### DIFF
--- a/simulators/eth2/withdrawals/specs.go
+++ b/simulators/eth2/withdrawals/specs.go
@@ -49,7 +49,7 @@ var (
 		AltairForkEpoch:         common.Big0,
 		BellatrixForkEpoch:      common.Big0,
 		CapellaForkEpoch:        common.Big1,
-		GenesisDelaySlots:       big.NewInt(10), // To avoid skipped slots in the start
+		GenesisDelaySlots:       big.NewInt(5), // To avoid skipped slots in the start
 		Eth1Consensus:           &el.ExecutionCliqueConsensus{},
 	}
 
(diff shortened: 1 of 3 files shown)


nazarhussain:
One other observation is that below code does not wait for exact number of lots. Tried increasing it but the tests does not wait for that number of slots before timeout. Is it matter of understanding or this code does not respect the genesis time? 

https://github.com/ethereum/hive/blob/d8dfa0ddd8a64e897eae66f74235ea472ec4bf7a/simulators/eth2/withdrawals/scenarios.go#L190-L191

marioevz:
@nazarhussain which client combinations require this at the moment ?

nazarhussain:
> @nazarhussain which client combinations require this at the moment ?

I was running with Lodestar and Nethermind, and on local machine the first few slots (3 to 4) always missed on for the `lodestar_bn`. With the above change it works fine to start.

simulators/eth2/withdrawals/specs.go
```diff
@@ -49,6 +49,7 @@ var (
 		AltairForkEpoch:         common.Big0,
 		BellatrixForkEpoch:      common.Big0,
 		CapellaForkEpoch:        common.Big1,
+		GenesisDelaySlots:       big.NewInt(10), // To avoid skipped slots in the start
```
marioevz, line 52: I think 10 slots might be too much, would 5 slots accomplish the same outcome?

This would add 1 minute to every test case run for each combination unfortunately.

nazarhussain: That actually depends upon the system load and use case. For this one, I believe `5` will work as well.

fjl:
@marioevz do you still want to add this?

marioevz:
It's nice to have the extra option, but forcing the delay as a default for the withdrawals suite seems a bit unnecessary in my opinion.

PR #706, opened 2023-02-07
simulators/ethereum/engine: Test sending `engine_forkchoiceUpdatedV1`, `engine_newPayloadV1` post-shanghai

marioevz:
Adds a test in `simulators/ethereum/engine` to verify that sending `engine_forkchoiceUpdatedV1` or `engine_newPayloadV1` post-Shanghai results in `InvalidParams` error (-32602).

Although this is not explicitly stated in spec, it caused an issue on nimbus as geth started building a transition payload with empty withdrawals because of this.

Setting this as draft because there is nothing explicit in the spec to describe this error, but I think it should be laid out.

@mkalinin @MariusVanDerWijden

### DIFF
--- a/simulators/ethereum/engine/suites/withdrawals/tests.go
+++ b/simulators/ethereum/engine/suites/withdrawals/tests.go
@@ -1171,7 +1171,7 @@ func (ws *WithdrawalsBaseSpec) Execute(t *test.Env) {
 			if !ws.SkipBaseVerifications {
 				// Try to send a PayloadAttributesV1 with null withdrawals after
 				// Shanghai
-				r := t.TestEngine.TestEngineForkchoiceUpdatedV2(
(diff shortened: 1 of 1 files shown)


mkalinin:
We would have to update all `V1` methods to make them return an error if they are called with post-Shanghai payload or payload attributes, moreover, the next fork would imply the same update for `V2` methods and so forth.

Though, we could do that I wonder if there is another approach to address this problem as if we require CL clients to use `V2` methods since Shanghai. Would we be able to test this kind of requirement in Hive?

marioevz:
Yes, this is addressed in #712

fjl:
@marioevz still want this in?

marioevz:
> @marioevz still want this in?

I think this is way too outdated, will close for now.

PR #711, opened 2023-02-09
simulators/ethereum/engine: Combinatorial Withdrawal Sync Tests

marioevz:
Creates suite `engine-withdrawals-sync`, which contains all syncing withdrawals tests from `engine-withdrawals` (thus tests are removed from this suite), and allows running the tests with all possible combinations.
For example, using
```
hive --client go-ethereum,nethermind,erigon --sim ethereum/engine --sim.limit "engine-withdrawals-sync/"
```
Will run the following sync tests:
- Nethermind syncing from Go-ethereum
- Erigon syncing from Go-ethereum
- Go-ethereum syncing from Nethermind
- Erigon syncing from Nethermind
- Go-ethereum syncing from Erigon
- Nethermind syncing from Erigon

Also adds test case where the client has to sync blocks with empty withdrawals list and empty transactions.

### DIFF
--- a/simulators/ethereum/engine/client/engine.go
+++ b/simulators/ethereum/engine/client/engine.go
@@ -69,7 +69,13 @@ type EngineClient interface {
 }
 
 type EngineStarter interface {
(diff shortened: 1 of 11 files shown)


fjl:
We now have withdrawals in the ethereum/sync test suite, so closing.

PR #712, opened 2023-02-10
simulators/eth2/common,withdrawals: Verify Engine API Calls from Beacon Client

marioevz:
Adds a library to detect Engine API calls coming from the beacon client using expired endpoints.

E.g. `engine_newPayloadV1` or `engine_forkchoiceUpdatedV1` containing `timestamp >= ShanghaiTime`.

Code can be easily reused for future upgrades.

Also is added the verification of these expired endpoints on all withdrawals testnet tests, and correctly detects nimbus sending an expired endpoint:
```
FAIL: received directive using expired endpoint engine_forkchoiceUpdatedV1: timestamp 1676055587 >= 1676055587
```
@mkalinin

### DIFF
--- a/simulators/eth2/common/spoofing/beacon/beacon_verification.go
+++ b/simulators/eth2/common/spoofing/beacon/beacon_verification.go
@@ -1,0 +1,121 @@
+package beacon
+
+import (
+	"fmt"
+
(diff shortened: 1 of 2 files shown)


mkalinin (approved):
Awesome 👍 I will attempt to constrain `V1` method usage on the spec level

PR #715, opened 2023-02-16
simulators/eth2/withdrawals: add builder tests

marioevz:
## Changes Included
- Adds a mock builder to the `eth2/common` library that is able to relay payload requests to a local Execution client to simulate a working builder
  - Client receives and registers validators to produce payloads when requested
  - When requested, the builder queries the CL for its state to obtain the prevrandao value to be able to construct the correct payload attributes
  -  The payload attributes are relayed to the EL and then it waits for the payload to be built
  - The builder constructs and signs a builder bid which is returned to the CL
  - When the signed blinded beacon block is received, the full payload (transactions + withdrawals) is sent as response
- Changes all Beacon clients to include builder parameters

Clients passing tests:
- Nimbus
- Lighthouse
- Prysm
- Teku
- Lodestar

~Lodestar test is currently failing because it stops requesting headers from the builder at the Capella fork.~ Fixed by https://github.com/ChainSafe/lodestar/pull/5179

### DIFF
--- a/clients/lighthouse-bn/lighthouse_bn.sh
+++ b/clients/lighthouse-bn/lighthouse_bn.sh
@@ -63,13 +63,14 @@ esac
 echo "bootnodes: ${HIVE_ETH2_BOOTNODE_ENRS}"
 
 CONTAINER_IP=`hostname -i | awk '{print $1;}'`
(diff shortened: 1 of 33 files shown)


marioevz:
@fjl let me know if it's ok to include client's changes in this same PR or I can split it up in a PR for each client (adds Builder API parameters to all consensus clients).

PR #717, opened 2023-02-17
clients/reth: adding reth client definition

onbjerg:
This is a work-in-progress PR to add support for [reth].

reth is still early stages, and we will gradually add support for more simulators. Support is being tracked in https://github.com/paradigmxyz/reth/issues/851

Before moving out of draft I will clean up the commit history in this pull request.

[reth]: https://github.com/paradigmxyz/reth

### DIFF
--- a/clients/reth/Dockerfile
+++ b/clients/reth/Dockerfile
@@ -1,0 +1,26 @@
+ARG branch=main
+FROM paradigmxyz/reth:$branch
+
+# Install script tools.
+RUN apt-get update -y
+RUN apt-get install -y bash curl jq
+
+# Add genesis mapper script.
+ADD genesis.json /genesis.json
+ADD mapper.jq /mapper.jq
+
+# Add the startup script.
+ADD reth.sh /reth.sh
+RUN chmod +x /reth.sh
+
+# Add the enode URL retriever script.
+ADD enode.sh /hive-bin/enode.sh
+RUN chmod +x /hive-bin/enode.sh
+
+# Create version.txt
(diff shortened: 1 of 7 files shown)


fjl:
Can we get this merged? It's OK to have clients that do not pass all tests in main branch. Having reth in would make it easier for us to apply ongoing maintenance to the client definition.

For example, we just now renamed the `branch` build argument to `tag`. This change also needs to be applied to clients/reth, but we can't do that if its in a Draft status PR.

Rjected:
> Can we get this merged? It's OK to have clients that do not pass all tests in main branch. Having reth in would make it easier for us to apply ongoing maintenance to the client definition.

Hi @fjl ! Sorry it's been a _while_, but I'm trying to prepare this so it can be merged. I just changed from `branch` to `tag`, to stay up to date, but noticed now that there are `.git` and `.local` dockerfiles. After adding those, what else would we need to for this to get merged?

fjl:
Alternative Dockerfiles are not mandatory. They are only needed to test the non-default branch of a client. I think the client definition is OK as is.

fjl:
The thing about error checks in the `rpc-compat` suite is on my TODO list, btw!

Rjected:
> The thing about error checks in the rpc-compat suite is on my TODO list, btw!

awesome! I removed those changes from this PR, since I'd like to just add the client definitions in this PR

holiman:
Does `paradigmxyz/reth` exist? Please see https://github.com/ethereum/hive/issues/970

PR #732, opened 2023-02-23
simulators/ethereum/pyspec: Add simulator for python spec tests (ethereum/exec-spec-tests)

spencer-tb:
Adds a new simulator for the tests in the execution-spec-tests repo: https://github.com/ethereum/execution-spec-tests

**For each test fixture:**
- Loads genesis fields into client,
- Blocks are fed into client one by one using EngineAPI,
- Storage, Nonce, Balance is extracted from last valid block using EngineAPI, and compared against expected fixture values.

Only for fixtures with forks >= Merge.


simulators/ethereum/pyspec/runner.go
```diff
@@ -0,0 +1,196 @@
...
+		if err != nil {
+			t.Fatalf("HIVE FATAL ---> Unable to send payload[%v] in test %s: %v ", blockNumber, tc.name, err)
+		}
+	}
+	t2 := time.Now()
+
+	// update head of "beacon" chain with the latest valid response
+	latestBlockHash := *engineClient.LatestNewPayloadResponse().LatestValidHash
+	fcState := &api.ForkchoiceStateV1{HeadBlockHash: latestBlockHash}
+	engineClient.ForkchoiceUpdatedV2(ctx, fcState, nil)
```
marioevz, line 134: We should check the result here, for `VALID` or `INVALID`, depending on whether the test contains a valid or invalid block.
And also check for the `err` in the response, in case the client did not respond correctly.

simulators/ethereum/pyspec/runner.go
```diff
@@ -0,0 +1,196 @@
...
+	t3 := time.Now()
+
+	// check nonce, balance & storage of accounts in final block against fixture values
+	latestBlockNumber := new(big.Int).SetUint64(engineClient.LatestNewPayloadSent().Number)
```
marioevz, line 138: It can be the case that the latest payload was `INVALID`, because the test is verifying that an invalid block is rejected.
In that case, `engineClient.LatestNewPayloadSent().Number` will be an unknown block to the client.
It should be ok to simply use `nil` instead of `latestBlockNumber`, to always check the head of the chain after the blocks were executed.

simulators/ethereum/pyspec/runner.go
```diff
@@ -0,0 +1,196 @@
...
+			t.Errorf("HIVE ERROR ---> Unable to call balance from account: %v, in test %s: %v", account, tc.name, errB)
+		} 
+
+		// check final nonce & balance matches expected in fixture 
+		if genesisAccount.Nonce != gotNonce {
+			t.Errorf(`HIVE ERROR ---> Nonce recieved from account %v doesn't match expected from fixture in test %s:
+			Recieved from block: %v
+			Expected in fixture: %v`, account, tc.name, gotNonce, genesisAccount.Nonce)
+		}
+		if common.BigToHash(genesisAccount.Balance) != common.BigToHash(gotBalance) {
```
marioevz, line 156:
```suggestion
		if genesisAccount.Balance.Cmp(gotBalance) != 0 {
```

simulators/ethereum/pyspec/runner.go
```diff
@@ -0,0 +1,196 @@
...
+			t.Errorf(`HIVE ERROR ---> Balance recieved from account %v doesn't match expected from fixture in test %s:
+			Recieved from block: %v
+			Expected in fixture: %v`, account, tc.name, gotBalance, genesisAccount.Balance)
+		}
+	
+		// check values in storage match those in fixture
+		for stPosition, stValue := range genesisAccount.Storage {
+			// get value in storage position stPosition from last block
+			gotStorage, errS := engineClient.StorageAt(ctx, account, common.BytesToHash(stPosition.Bytes()), latestBlockNumber)
```
marioevz, line 165:
```suggestion
			gotStorage, errS := engineClient.StorageAt(ctx, account, stPosition, latestBlockNumber)
```

simulators/ethereum/pyspec/runner.go
```diff
@@ -0,0 +1,196 @@
...
+			if errS != nil {
+				t.Errorf("HIVE ERROR ---> Unable to call storage bytes from account address %v, storage position %v, in test %s: %v", account, stPosition, tc.name, errS)
+			}
+			// check recieved value against fixture value
+			if stValue != common.BytesToHash(gotStorage) {
```
marioevz, line 170: It could be worth it to convert this block into a batch request like in https://github.com/ethereum/go-ethereum/blob/ee530c0d5aa70d2c00ab5691a89ab431b73f8165/ethclient/ethclient.go#L143-L162

simulators/ethereum/pyspec/types.go
```diff
@@ -0,0 +1,280 @@
...
+	ExtraData        []byte           `json:"extraData"`
+	MixHash          common.Hash      `json:"mixHash"`
+	Nonce            types.BlockNonce `json:"nonce"`
+	BaseFee          *big.Int         `json:"baseFeePerGas"`
+	Hash             common.Hash      `json:"hash"`
+	WithdrawalsRoot  common.Hash	  `json:"withdrawalsRoot`
+}
+
+func (b *blockHeader) UnmarshalJSON(input []byte) error {
+	type blockHeader struct {
```
marioevz, line 78: All `UnmarshalJSON` procedures in this file can be replaced by automatically-generated files by placing a comment like this:
```
//go:generate go run github.com/fjl/gencodec -type blockHeader -field-override blockHeaderMarshaling -out gen_bh.go
```
(where `blockHeaderMarshaling` is the structure contained in this procedure) somewhere within this file.

Then running `go generate` in the same folder as this file.

Make sure to use `go get github.com/fjl/gencodec` before running `go generate` for it to work properly.

marioevz (changes requested):
Looks very promising!

A few comments.

simulators/ethereum/pyspec/runner.go
```diff
@@ -0,0 +1,196 @@
...
+		for name, fixture := range fixtureTests {
+			network := fixture.json.Network
+			// skip networks post merge
+			for _, skip := range []string{"Istanbul", "Berlin", "London"} {
+				if strings.Contains(network, skip) {
+					continue
+				}
+			}
+			// check network is valid 
+			if _, exist := envForks[network]; !exist {
```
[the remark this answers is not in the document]
spencer-tb: How is the following?

```go

network := fixture.json.Network
// skip networks post merge or not supported
for _, skip := range []string{"Istanbul", "Berlin", "London"} {
	if strings.Contains(network, skip) || envForks[network] == nil {
		t.Logf("skipping test for network %s", network)
			continue
	}
}

```

simulators/ethereum/pyspec/types.go
spencer-tb: Changes here:
https://github.com/ethereum/hive/pull/732/commits/21148677b2cbfd86057d4c83ebdf704eef92650f#diff-5a6a26a4292c855637c91a0f97c3fae500808507cc0919c6172da75c8e748121

fjl:
I have removed the very detailed documentation in README.md. There is certainly no need to duplicate parts of the code into the README.

simulators/ethereum/pyspec/runner.go
spencer-tb: I have something like this. My assumption is that by using the batch request if will speed up the overall simulation, but at the cost of extra code?

```go
// create a batch of RPC calls for each storage position
if len(genesisAccount.Storage) > 0 {
	stValues := make(map[common.Hash]common.Hash, len(genesisAccount.Storage))
	reqs := make(map[common.Hash]rpc.BatchElem, len(genesisAccount.Storage))
	// create request for every storage element
	for stPosition, _ := range genesisAccount.Storage {
		reqs[stPosition] = rpc.BatchElem{
			Method: "eth_getStorageAt",
			Args:   []interface{}{account, stPosition, "latest"},
			Result: stValues[stPosition],
		}
	}
	// batch call for each storage element
	engineClient.BatchCallContext(ctx, reqs)
	// check each request was successful and that each storage value matches expected 
	for stPosition, expValue := range genesisAccount.Storage {
		if reqs[stPosition].Error != nil {
			fmt.Errorf(`HIVE ERROR ---> Unable to call storage bytes from account address %v, storage position %v, in test %s:
				%v`, account, stPosition, tc.name, reqs[stPosition].Error)
		}
		if stValues[stPosition] != expValue {
			fmt.Errorf(`HIVE ERROR ---> Storage received from account %v doesn't match expected from fixture in test %s:
			From storage address: %v
			Received from block: %v
			Expected in fixture: %v`, account, tc.name, stPosition, stValues[stPosition], expValue)
		}
	}
}
```

simulators/ethereum/pyspec/runner.go
spencer-tb: It seems like the `err` in the response: `plStatus.ValidationError` is returning `null` for any invalid response.
```
{"jsonrpc":"2.0","id":1,"result":{"status":"INVALID","latestValidHash":null,"validationError":null}
```

I have committed the following nonetheless. 

```go
// send payloads and check response
latestValidHash := common.Hash{}
for blockNumber, payload := range tc.payloads {
	plException := tc.fixture.json.Blocks[blockNumber].Exception
	expectedStatus := "VALID"
	if plException != "" {
		expectedStatus = "INVALID"
	}
	// execute fixture block payload
	plStatus, plErr := engineClient.NewPayloadV2(context.Background(), payload)
	if plErr != nil {
		t.Fatalf("HIVE FATAL ---> ...`)
	}
	// update latest valid block hash
	if plStatus.Status == "VALID" {
		latestValidHash = *plStatus.LatestValidHash
	}
	// check payload status is expected from fixture
	if expectedStatus != plStatus.Status {
		t.Errorf(`HIVE ERROR ---> ...`)
	}
	// check error message if invalid
	if expectedStatus == "INVALID" && *plStatus.ValidationError != plException {
		t.Errorf(`HIVE ERROR ---> ...`)
	}
}
```

marioevz: Yes, I think it's an issue of the lack of error unification, most clients print different error strings here but it would be very difficult for us to match a string for an expected error on every client.

For now we could simply check for `INVALID`.

simulators/ethereum/pyspec/runner.go
marioevz: It packages everything in a single RPC call, which then the client should optimize and fetch all at the same time from its internal state DB.

So I definitely think it's worth it for bigger test cases where we check a ton of storage keys from different accounts.

simulators/ethereum/pyspec/runner.go
```diff
@@ -0,0 +1,207 @@
...
+		"HIVE_FORK_DAO_VOTE": "1",
+		"HIVE_CHAIN_ID":      "1",
+		"HIVE_SKIP_POW":      "1",
+		"HIVE_NODETYPE":      "full",
+	}
+	tc.updateEnv(env)
+	t0 := time.Now()
+
+	// start client (also creates an engine RPC client internally)
+	t.Log("HIVE LOG --> Starting client with Engine API.")
```
fjl, line 100: Please remove these `HIVE LOG -->`, `HIVE FATAL -->` etc. prefixes. It is sufficient to prefix the error message in `t.Fatalf()` calls with simply `Error:`

spencer-tb: Sure, happy to remove! I added it to help distinguish where the log/error messages were coming from (i.e hive or geth). But I've only just realised now that the differing containers achieve this :)

fjl: BTW, you can also run the web interface locally to check your results.  See here for more information: https://github.com/ethereum/hive/blob/master/docs/commandline.md#viewing-simulation-results-hiveview

simulators/ethereum/pyspec/README.md
```diff
@@ -0,0 +1,26 @@
...
+
+The `pyspec simulator` uses the latest test fixtures from the
+most recent execution-spec-tests [release](https://github.com/ethereum/execution-spec-tests/releases).
+
+
+### Running Specific Test Fixtures
+
+To run a specific or set of test fixtures `testCategory := ""`, within the third code section of `main.go/fixtureRunner()`, can be updated:
+- To only run the `large_withdrawals.json` test fixture: `testCategory := "withdrawals/withdrawals/large_withdrawals.json"`.
+- To run all the withdrawals test fixtures: `testCategory := "withdrawals"`.
```
fjl, line 20: Not sure if you are aware, but the hive tool has an option for this: `hive --sim.limit <pattern>` will select tests based on a regular expression. The format of the pattern is `<suite>/<test>` and the suite/test components are matched individually. To run all suites, but only tests containing `foo` within them, run it hive like this `hive --sim.limit /foo`.

spencer-tb: I wasn't aware I could use this here! Thanks for the tip - I've addressed the latter the latest commit.

spencer-tb:
To help aid the use of `--sim-limit` in my previous commit, I am linking this PR: https://github.com/ethereum/execution-spec-tests/pull/69.

simulators/ethereum/pyspec/main.go
```diff
@@ -40,9 +38,7 @@ func main() {
 	hivesim.MustRunSuite(hivesim.New(), suite)
 }
 
-// ------------------------------------------------------------------------- //
-// fixtureRunner() loads the pyspec test files and spawns the client tests.  //
-// ------------------------------------------------------------------------- //
+// fixtureRunner() loads the pyspec test files and spawns the client tests.
```
fjl, line 41: Parentheses denoting a function are typically omitted in Go documentation comments.

simulators/ethereum/pyspec/types.go
```diff
@@ -0,0 +1,181 @@
...
+	GasUsed    math.HexOrDecimal64   `json:"gasUsed"`
+	Timestamp  *math.HexOrDecimal256 `json:"timestamp"`
+	ExtraData  hexutil.Bytes         `json:"extraData"`
+	BaseFee    *math.HexOrDecimal256 `json:"baseFeePerGas"`
+}
+
+//go:generate go run github.com/fjl/gencodec -type transactions -field-override transactionsUnmarshaling -out gen_txs.go
+type transactions struct {
+	Nonce     uint64         `json:"nonce"`
+	To        common.Address `json:"to.omitempty"`
```
marioevz, line 99:
```suggestion
	To        common.Address `json:"to,omitempty"`
```

simulators/ethereum/pyspec/types.go
```diff
@@ -0,0 +1,181 @@
...
+	Number     *math.HexOrDecimal256 `json:"number"`
+	GasLimit   math.HexOrDecimal64   `json:"gasLimit"`
+	GasUsed    math.HexOrDecimal64   `json:"gasUsed"`
+	Timestamp  *math.HexOrDecimal256 `json:"timestamp"`
+	ExtraData  hexutil.Bytes         `json:"extraData"`
+	BaseFee    *math.HexOrDecimal256 `json:"baseFeePerGas"`
+}
+
+//go:generate go run github.com/fjl/gencodec -type transactions -field-override transactionsUnmarshaling -out gen_txs.go
+type transactions struct {
```
marioevz, line 97: Type name should be `transaction`, because normally the plural is reserved for a type that overrides an array.

simulators/ethereum/pyspec/types.go
```diff
@@ -0,0 +1,181 @@
...
+	Nonce     uint64         `json:"nonce"`
+	To        common.Address `json:"to.omitempty"`
+	Value     *big.Int       `json:"value"`
+	Data      []byte         `json:"data"`
+	GasLimit  uint64         `json:"gasLimit"`
+	GasUsed   uint64         `json:"gasUsed"`
+	SecretKey common.Hash    `json:"secretKey"`
+}
```
marioevz, line 105: There are some missing fields that should be included from here:
https://github.com/ethereum/execution-spec-tests/blob/2dcfb503f4c50d368958c34a99b39e1aa3f62238/src/ethereum_test_tools/common/types.py#L912-L940

simulators/ethereum/pyspec/types.go
```diff
@@ -0,0 +1,181 @@
...
+	UncleHash           common.Hash      `json:"sha3Uncles"`
+	UncleHashAlt        common.Hash      `json:"uncleHash"` // name in fixtures
+	Coinbase            common.Address   `json:"coinbase"`
+	CoinbaseAlt         common.Address   `json:"author"` // nethermind/parity/oe name
+	CoinbaseAlt2        common.Address   `json:"miner"`  // geth/besu name
+	StateRoot           common.Hash      `json:"stateRoot"`
+	TransactionsTrie    common.Hash      `json:"transactionsRoot"`
+	TransactionsTrieAlt common.Hash      `json:"transactionsTrie"` // name in fixtures
+	ReceiptTrie         common.Hash      `json:"receiptsRoot"`
+	ReceiptTrieAlt      common.Hash      `json:"receiptTrie"` // name in fixturse
```
marioevz, line 71:
```suggestion
	ReceiptTrieAlt      common.Hash      `json:"receiptTrie"` // name in fixtures
```

marioevz (commented):
Some more comments.

marioevz (approved):

PR #738, opened 2023-03-03
simulators/eth2/common: create client interfaces, update mock builder

marioevz:
This PR introduces the interfaces `Client` and `ManagedClient` which are now used by the `ExecutionClient`, `BeaconClient` and `ValidatorClient` structures to communicate with the clients.

Previously, the structures directly used `hivesim.Client`, which required a client launched and managed by hive.

`hivesim.Client` is now abstracted through `ManagedClient` to allow instantiation of `ExecutionClient`, `BeaconClient` or `ValidatorClient` and use an external client.

With this change, modules like the mock builder can now be imported by other non-hive modules, and even compiled separately as standalone commands.

### Testnet Clients
- `Client`/`ManagedClient` interfaces
- Execution and Beacon clients now have separate configuration structures, `ExecutionClientConfig` and `BeaconClientConfig` respectively
- Beacon client structure can now automatically fetch the beacon spec from the client itself
- Beacon client subnets are now possible: If a beacon client is started with a specific subnet, their ENR will only be passed to the clients of the same subnet (useful to test forks and re-orgs)
- Testnet creation now uses configuration structs

### Withdrawals Test Suite
-  Tests now use the mock builder functionality to invalidate payloads

### Mock builder
- Support to invalidate payloads, used in testnets to verify the circuit breaker behavior
- Support to create invalid outdated (from a previous fork) payloads.
- Mock builder behavior can now be altered while running by using the REST API (see builder/mock/README.md)


holgerd77:
Hi @marioevz, thanks for all these extensive restructuring/refactoring works and efforts, really appreciated by "the users"!! 🤗 🤩 ❤️

marioevz:
We could first merge #741 and then I could update the `ManagedClient` interface in this PR to include pause/unpause feature, which would allow to construct beacon chain tests which pause the beacon and validator clients. But let me know what you think @fjl

marioevz:
I've been running these changes locally for quite some time now, and they are required to introduce more tests into the branch, so I will be merging now.

PR #746, opened 2023-03-15
graphql: consistent rule for Long outputs and args

s1na:
Pending on https://github.com/ethereum/execution-apis/pull/389


s1na:
@fjl spec is merged, PTAL

PR #754, opened 2023-04-07
simulators/ethereum/engine: Fix Geth Node Closing

marioevz:
## Changes Included
- Properly unsubscribe from events so the Geth node can close
- Close secondary clients that were never added to the CLMocker


marioevz:
Ran entire set of `ethereum/engine` simulator tests, and they are all passing. Merging.

@fjl we might need to update the hivetests.

PR #757, opened 2023-04-18
simulators/ethereum/engine: Unique Payload ID Tests

marioevz:
Implements test for https://github.com/ethereum/execution-apis/pull/401

cc @mkalinin

### DIFF
--- a/simulators/ethereum/engine/client/engine.go
+++ b/simulators/ethereum/engine/client/engine.go
@@ -4,8 +4,8 @@ import (
 	"context"
 	"math/big"
 
-	"github.com/ethereum/go-ethereum/core"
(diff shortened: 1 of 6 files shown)


mkalinin:
It might worth adding the case when a new withdrawal is appended to the list, this case is artificial but should be pretty easy to add. It seems like fuzzing for this kind of things should work well too in addition to regular tests.

marioevz:
> It seems like fuzzing for this kind of things should work well too in addition to regular tests.

I'd really like to add fuzzing to hive, and my main deterrent is that hive tests are already flaky on their own, so adding fuzzing will make this worse, but it's something I'll keep in mind.

I added the suggested test case @mkalinin

PR #760, opened 2023-04-21
clients/erigon: --externalcl flag is not supported anymore

yperbasis:
After https://github.com/ledgerwatch/erigon/pull/7349 the `--externalcl` flag is not supported anymore (the external CL mode is the default now)


marioevz (approved):
LGTM

lemenkov:
The changes like this one should be mentioned in ChangeLog since it's a breaking one (Erigon refuses to start until you remove --externalcl)

PR #761, opened 2023-04-25
clients/go-ethereum: Cancun support

marioevz:
Adds support for `HIVE_CANCUN_TIMESTAMP` environment variable.

Also add commented lines for building from git in the Dockerfile.


clients/go-ethereum/Dockerfile
```diff
@@ -17,6 +17,32 @@ FROM ethereum/client-go:$branch as builder
...
+#   echo "{}"                                                      \
+#   | jq ".+ {\"repo\":\"$(git config --get remote.origin.url)\"}" \
+#   | jq ".+ {\"branch\":\"$(git rev-parse --abbrev-ref HEAD)\"}"  \
+#   | jq ".+ {\"commit\":\"$(git rev-parse HEAD)\"}"               \
+#   > /version.json)                                            && \
+#   cp go-ethereum/build/bin/geth /usr/local/bin/geth           && \
+#   apk del go git make gcc musl-dev linux-headers              && \
+#   rm -rf /go-ethereum && rm -rf /var/cache/apk/*
+
+## ----
```
fjl, line 44: Wondering why you are adding these lines, since we already have the equivalent steps just above?

marioevz: This one makes it just a tiny bit easier to build from a git branch, instead of having to manually clone inside the folder, and having to checkout to switch branches to compile from a different source.

I've been going to the commit where this was removed and copying this excerpt to do my local testing, I just find it easier to test locally with this code.

holiman: > since we already have the equivalent steps just above?

Yeah, no the steps above execute within the `ethereum/client-go` container as a builder, and is not suitable if you want to clone some non-official repository. LGTM

holiman (approved):
LGTM

PR #765, opened 2023-04-27
simulators/eth/pyspec/eip4844: Update the pyspec simulator for eip4844

spencer-tb:
WIP branch for running the 4844 generated fixtures from execution spec tests.


fjl:
Can we get this branch cleaned up? Support for build_args is in now, so a lot of changes can be removed in this branch.

fjl:
go-ethereum upstream now also has support for blob transaction types, so replacing the version might not be necessary anymore.

PR #767, opened 2023-04-27
libhive: client build args and alternative dockerfiles

marioevz:
Allows specifying a `--client-file` parameter which shall be read in YAML or JSON format and must contain the client build modifiers.

Supported modifiers:
- client: Name of the client, eg: besu, go-ethereum, etc
- dockerfile: Dockerfile to use. E.g. using `dockerfile: git` will build using `Dockerfile.git` instead of the default `Dockerfile`
- build_args: Build arguments passed to the docker build engine to use when building the image

Supported docker image build arguments depend on the client and the docker image being used, but common client build arguments are:

 - tag: Git branch or docker tag name to use during build process.
 - github: The repository containing the client source code (for git-based Dockerfiles)
 - baseimage: The client image name on DockerHub.

Also adds example dockerfiles for go-ethereum: `Dockerfile.git`, to build from a git branch, and `Dockerfile.local` to build from a local source path.

This change should allow clients to introduce different Dockerfiles to test using hive in many different environments, such as the CI.

### DIFF
--- a/clients/go-ethereum/Dockerfile
+++ b/clients/go-ethereum/Dockerfile
@@ -5,32 +5,6 @@ ARG repo=client-go
 ARG branch=latest
 FROM $user/$repo:$branch as builder
 
-## ----
-## Uncomment the steps below (and comment out the steps above!) to build
-## go-ethereum from a git branch.
-
-# FROM alpine:latest as builder
-# ARG user=ethereum
-# ARG repo=go-ethereum
-# ARG branch=master
-# RUN \
-#   apk add --update bash curl jq go git make gcc musl-dev         \
-#         ca-certificates linux-headers                         && \
-#   git clone --depth 1 --branch $branch                           \
-#       https://github.com/$user/$repo                          && \
-#   (cd go-ethereum && make geth)                               && \
-#   (cd go-ethereum                                             && \
-#   echo "{}"                                                      \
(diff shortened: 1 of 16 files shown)


marioevz:
@fjl Another option for the yaml/json files is to have a `build_params` key which could basically be a dictionary that contains any build parameter passed to the docker build:
```yaml
- name: go-ethereum
  dockerfile: git
- name: nethermind
  build_params:
    user: nethermindeth
    repo: hive
    branch: latest
- name: go-ethereum
  dockerfile: local
  build_params:
    local_path: ./path/to/local/code/base/
```
The problem here would be how to tag each image built.

Right now we simply append the branch to the client's name to form the tag (`go-ethereum-master`), but with an arbitrary amount of parameters this could become cumbersome.

One possibility could be to create the tag using the source client config yaml/json file name and index of the client in the file.

E.g. for a `clients.yaml`:
```yaml
- name: go-ethereum
  dockerfile: git
- name: nethermind
  build_params:
    user: nethermindeth
    repo: hive
    branch: latest
- name: go-ethereum
  dockerfile: local
  build_params:
    local_path: ./path/to/local/code/base/
```
the clients could be named `go-ethereum-clients-1`, `nethermind-clients-2`, `go-ethereum-clients-3`.

PR #770, opened 2023-05-02
simulators/eth2: Extract Builder and Clients

marioevz:
This PR extracts the following components into their own repository:
`./simulators/eth2/common/builder` -> https://github.com/marioevz/mock-builder
`./simulators/eth2/common/clients` -> https://github.com/marioevz/eth-clients

Reason is being able to use `mock-builder` as an external command, and this requires the structs and interfaces declared in `eth-clients`.

This results in removing lot of code from `./simulators/eth2/common` to make it more reusable.

Other fixes included:
- Slot time configuration is now part of the common library and automatically configured for clients that don't support 6s slots
- Incremental forks are removed (trick to start from phase0 up to bellatrix/capella in order for some clients to work, no longer necessary)
- Fix TTD > 0 setup in `eth2/engine`


marioevz:
Rebased to main to fix geth not working due to PoW.

marioevz:
I'll finish up running the simulators with the change and merge in a few hours unless objections :+1: @fjl

PR #773, opened 2023-05-03
clients/go-ethereum: remove deprecated PoW

s1na:
Geth PoW mining and real verification, and as a result many of the utilities used in tests such as makecache makedag etc.

The `--fakepow` flag is also removed. Now a merge-ready network is required and in that case a PoW fake verifier will be available. To trigger this we set `"terminalTotalDifficultyPassed": true`.

https://github.com/ethereum/go-ethereum/pull/27178

### DIFF
--- a/simulators/ethereum/engine/Dockerfile
+++ b/simulators/ethereum/engine/Dockerfile
@@ -1,7 +1,3 @@
-# Generate the ethash verification caches.
-# Use a static version because this will never need to be updated.
-FROM ethereum/client-go:v1.10.20 AS geth
-
 # This simulation runs Engine API tests.
 FROM golang:1-alpine as builder
 RUN apk add --update gcc musl-dev linux-headers
@@ -16,5 +12,4 @@ FROM alpine:latest
 ADD . /source
 WORKDIR /source
 COPY --from=builder /source/engine .
-COPY --from=geth    /ethash /ethash
 ENTRYPOINT ["./engine"]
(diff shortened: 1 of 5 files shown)


simulators/ethereum/engine/Dockerfile
```diff
@@ -1,11 +1,6 @@
 # Generate the ethash verification caches.
 # Use a static version because this will never need to be updated.
 FROM ethereum/client-go:v1.10.20 AS geth
-RUN \
- /usr/local/bin/geth makecache  1     /ethash && \
- /usr/local/bin/geth makecache  30000 /ethash && \
- /usr/local/bin/geth makedag    1     /ethash && \
- /usr/local/bin/geth makedag    30000 /ethash
```
holiman, line 8: I believe we can skip the line above too, and also L24 `COPY --from=geth /ethash /ethash`

holiman (approved):
LGTM, it can't be any less functional than current master. Might aswell get it merged IMO. Un-draft @s1na ?

holiman:
> The --fakepow flag is also removed. Now a merge-ready network is required and in that case a PoW fake verifier will be available. To trigger this we set "terminalTotalDifficultyPassed": true.

An alternative solution is to re-add the `--fakepow` flag. For example, we might have cases with invalid chains, due to invalid difficulty. We can still support that, because the fakepow mechanism still exists internally in geth, we just removed the flag. 

Fakepow is basically ethash but without verifying the PoW `nonce`. 

This PR will instead accept "invalid difficulty chain" due to the `terminalTotalDifficultyPassed` flag. **Although**, one might argue that failing those tests are _correct_, because geth is no longer able to function correctly without external guidance on canon header.

s1na:
I still hadn't a chance to look at every simulation but un-drafted for now. This will at least fix CI for other PRs.

Issue #774, opened 2023-05-05
pyspec simulator failure


holiman:
See https://hivetests.ethdevops.io/suite.html?suiteid=1683213361-92616ab5a25698357b8ccca7d9b4195f.json&suitename=pyspec#test-74 

The cause is `unable to extract fixture fields: failed to decode block: transaction type not supported`. 
I'm looking into it

PR #775, opened 2023-05-05
sim/pyspec: bubble error

holiman:
I think this fixex https://github.com/ethereum/hive/issues/774 , at least I don't hit the error locally any longer. That said, for some reason it doesn't report the test as failing either. 

Ideally, we'd not have  a panic, but instead a failing test which contains info on what went wrong.


holiman:
To be reopened in a while...

PR #766, opened 2023-04-27
clients/erigon: explicit --db.size.limit value

yperbasis:
After https://github.com/ledgerwatch/erigon/pull/7325 the default DB limit doesn't work with Hive and results in
```
[452f4e7d] [EROR] [04-27|13:51:09.753] Erigon startup                           err="mdbx_env_open: MDBX_TOO_LARGE: Database is too large for current system, e.g. could NOT be mapped into RAM, label: chaindata, trace: [kv_mdbx.go:266 node.go:323 node.go:326 backend.go:205 node.go:106 main.go:59 make_app.go:41 command.go:274 app.go:332 app.go:309 main.go:36 proc.go:250 asm_amd64.s:1598]"
[452f4e7d] mdbx_env_open: MDBX_TOO_LARGE: Database is too large for current system, e.g. could NOT be mapped into RAM, label: chaindata, trace: [kv_mdbx.go:266 node.go:323 node.go:326 backend.go:205 node.go:106 main.go:59 make_app.go:41 command.go:274 app.go:332 app.go:309 main.go:36 proc.go:250 asm_amd64.s:1598]
```

This PR fixes Erigon invocation under Hive.

PR #776, opened 2023-05-05
clients/erigon: change p2p version priorities to fix several Erigon devp2p tests

cryptoscruffy:
Use 68 version as default. This fixes 5 failed devp2p tests for Erigon

### DIFF
--- a/clients/erigon/erigon.sh
+++ b/clients/erigon/erigon.sh
@@ -55,6 +55,8 @@ fi
 mkdir /erigon-hive-datadir
 FLAGS="$FLAGS --datadir /erigon-hive-datadir"
 FLAGS="$FLAGS --db.size.limit 2GB"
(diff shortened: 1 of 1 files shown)


holiman:
Hm... so is the end-user supposed to know this, and set those flags explicitly?

fjl:
Yeah, I think the same thing, why are these flags not default? cc. @yperbasis

cryptoscruffy:
As far as I know end users use --p2p.protocol setting often enough,  but I don't sure it is a correct way to resolve the issue. I've tried to copy accepted https://github.com/ethereum/hive/pull/766 PR, reject this PR please if you think it is wrong approach.

cryptoscruffy:
Looks like 67 version is default now just because it was not updated in Erigon,  so the correct way to fix this issue is to update Erigon's defaults. I'll try to make PR to Erigon's repo instead of this one.

PR #779, opened 2023-05-05
clients/erigon: Dockerfile support to build Erigon from git or locally.

spencer-tb:
Requires: https://github.com/ethereum/hive/pull/767.

Adds clients/erigon/Dockerfile.git to build Erigon from any git branch.
Adds clients/erigon/Dockerfile.local to build Erigon from a local directory.

### DIFF
--- a/clients/erigon/Dockerfile
+++ b/clients/erigon/Dockerfile
@@ -1,29 +1,31 @@
+## Build Erigon Via Pre-Built Image:
+ARG user=thorax
+ARG repo=erigon
 ARG branch=devel
-FROM thorax/erigon:$branch
+FROM $user/$repo:$branch
 
 # The upstream erigon container uses a non-root user, but we need
 # to install additional commands, so switch back to root.
 USER root
 
-# Install script tools.
-RUN apk add --update bash curl jq
-
-# Add genesis mapper script.
-ADD genesis.json /genesis.json
-ADD mapper.jq /mapper.jq
-
-# Add the startup script.
-ADD erigon.sh /erigon.sh
-RUN chmod +x /erigon.sh
-
-# Add the enode URL retriever script.
-ADD enode.sh /hive-bin/enode.sh
-RUN chmod +x /hive-bin/enode.sh
+RUN apk add --no-cache \
+    bash \
+    curl \
+    jq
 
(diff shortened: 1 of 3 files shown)


clients/erigon/Dockerfile
```diff
@@ -1,29 +1,31 @@
...
 # The upstream erigon container uses a non-root user, but we need
 # to install additional commands, so switch back to root.
 USER root
 
-# Install script tools.
-RUN apk add --update bash curl jq
+RUN apk add --no-cache \
+    bash \
+    curl \
+    jq
```
holiman, line 14:
```suggestion
RUN apk add --no-cache bash curl jq
```

> The --no-cache option allows to not cache the index locally, which is useful for keeping containers small.
> Literally it equals apk update in the beginning and rm -rf /var/cache/apk/* in the end.

Nice, today I learned. Prefer one line though

clients/erigon/Dockerfile.git
```diff
@@ -1,45 +1,54 @@
...
+COPY --from=builder /go/erigon/build/bin/* /usr/local/bin/
 
-# Create version.txt
-RUN /usr/local/bin/erigon --version | sed -e 's/erigon version \(.*\)/\1/' > /version.txt
+RUN apk add --no-cache \
+    bash \
+    curl \
+    jq \
+    libstdc++ \
+    libgcc
```
holiman, line 37: Why are thes (libstdc++, libgcc) needed in the "not-builder"?

holiman (commented):
Looks mostly good to me

clients/erigon/Dockerfile.git
spencer-tb: Ahh so I believe these are already included in the default Dockerfile base image with a pre-built Erigon. 

As we are building from source here (‘.git’) and also in ‘.local’ : in the “not-builder” stage where we are running the client, these dependencies are needed to run the client successfully

clients/erigon/Dockerfile
spencer-tb: Made some small changes! 

This led down the path of trying to reduce the size of the builder stage only to realize it was redundant for a multi stage docker build so I also learned something :D

PR #780, opened 2023-05-12
Enable DEBUG API used by rcp-compat suite

gfukushima:
Currently besu fails on some of the rcp-compat tests because the JSON RPC DEBUG API is not enabled


marioevz (approved):
LGTM.

There are some subtle differences still that are making the rest of the failing test not pass.

PR #786, opened 2023-05-25
simulators/ethereum/engine: Reduce max withdrawals per block

marioevz:
Reduces the maximum number of withdrawals per test case to the more coherent value of 16, which is the maximum for the beacon chain in mainnet.

The single exception is `Withdraw many accounts`, which deliberately attempts to send more withdrawals for testing purposes (which are still applied because the EL has no hard-coded limit).


stdevMac (approved):
LGTM

PR #790, opened 2023-06-08
simulators/ethereum/pyspec: update for EIP-4844 tests

spencer-tb:
### simulators/ethereum/pyspec
:heavy_plus_sign: Adds changes required to run EIP-4844 execution-spec-tests:
https://github.com/ethereum/execution-spec-tests/tree/main/tests/eips/eip4844

:heavy_plus_sign: Now using `fixtures-cancun.tar.gz` from:
https://github.com/ethereum/execution-spec-tests/releases/tag/v1.0.0

### Remaining tasks
- [ ] All clients passing on -> https://hivecancun.ethdevops.io/
- [x] Replace RLP decoder with custom function that extracts fields from fixtures for EngineAPI payloads
- [x] Remove go.mod replace in both `pyspec` and `engine` simulators (requires Cancun changes to be merged


clients/besu/besu.sh
```diff
@@ -153,6 +153,11 @@ if [ "$HIVE_TERMINAL_TOTAL_DIFFICULTY" != "" ]; then
     RPCFLAGS="$RPCFLAGS --engine-host-allowlist=* --engine-jwt-enabled --engine-jwt-secret /jwtsecret"
 fi
 
+# Enable kzg trusted setup post Cancun
+if [ "$HIVE_CANCUN_TIMESTAMP" != "" ]; then
+    FLAGS="$FLAGS --kzg-trusted-setup=/kzg.txt"
+fi
+
```
fjl, line 160: I wonder, doesn't besu ship with the default trusted setup like all other clients?

simulators/ethereum/pyspec/go.mod
```diff
@@ -50,10 +55,14 @@ require (
...
+	golang.org/x/exp v0.0.0-20230515195305-f3d0a9c9a5cc // indirect
+	golang.org/x/sync v0.2.0 // indirect
+	golang.org/x/sys v0.8.0 // indirect
+	golang.org/x/text v0.9.0 // indirect
 	google.golang.org/protobuf v1.28.1 // indirect
 	gopkg.in/natefinch/npipe.v2 v2.0.0-20160621034901-c1b8fa8bdcce // indirect
+	rsc.io/tmplfunc v0.0.3 // indirect
 )
+
+replace github.com/ethereum/go-ethereum v1.12.0 => github.com/marioevz/go-ethereum v1.10.14-0.20230607235252-075f62d72ccc
```
fjl, line 68: Why is this replacement needed?

fjl:
I'm not a big fan of putting the yaml config files in `clients/`. Maybe we can create a new directory for this, like `inv/4844/`, where these will be stored. 

I understand that client configs with a custom branch are needed while the clients don't have the feature in their default branch yet. And we definitely need a way to share these configs.

clients/besu/besu.sh
spencer-tb: Not yet at least - we get thrown this error without including it as a flag:
```
WARN  | Besu | Failed to start Besu...
[6efa3097] picocli.CommandLine$ParameterException:
 --kzg-trusted-setup is mandatory when providing a custom genesis that support data blobs
```

I'm assuming it will have the default setup for mainnet so I will dig deeper and ask around!

simulators/ethereum/pyspec/go.mod
spencer-tb: Currently its the only branch (I believe) with the dataGasUsed and excessDataGas fields added the `*core.Genesis` struct which we are using in `types.go`

This is the required type here: https://github.com/ethereum/hive/blob/master/simulators/ethereum/engine/client/hive_rpc/hive_rpc.go#L40
that we use to start a clients with a genesis file alongside the EngineAPI

Maybe there is a better way to do this, that allows us to use a custom genesis struct?

Nonetheless once 4844 is merged no replacement will be required!

fjl: I see. Please please, if you need certain fields added in go-ethereum, send a PR to go-ethereum!

clients/besu/besu.sh
Gabriel-Trintinalia: I believe it ships:
```
# Copy KZG-trusted-setup
COPY --from=builder /besu/config/src/main/resources/kzg-trusted-setups/mainnet.txt /kzg-trusted-setup.txt
```

jflo: It ships, but if you specify your own genesis you need to also provide a corresponding setup if that genesis supports datagas.

fjl (approved):

Issue #284, opened 2020-05-05
Permalink functionality (viewer)


holiman:
Right now, when we load a testsuite, the suite get's "permalinked" by adding it to the location via history api. 
A better solution would be to have a "permalink" button, which would, 
- put the testsuite into the url
- put any datatable filters into the url, 
- put datatable sort-column/order into the url, 

Loading page would then apply those parameters. It would be easy to show someone a specific result. 
Current viewer is here: https://github.com/holiman/testreport_template

fjl:
This is improved in #348. The datatable filters etc. are not tracked, but basic navigation and linking works.

Anything beyond the changes in #348 will be a lot of work and will mess up the app code a lot. I think it would be good to split up the 'tabs' in the UI into individual HTML pages with exactly one purpose per page. It would simplify the app, but I don't want to spend more time on it right now.

fjl:
Linking to tests is now possible. Closing this now because I think it's good enough. 

The situation of wanting to share a specific view of the tests table with search+sort order is pretty rare, and I think it'd be kind of confusing to have a button that creates the link.

PR #811, opened 2023-08-04
cmd/hiveview: Add Column Filters

holgerd77:
This adds more differentiated column filters to `hiveview` for easier selection of a test suite or a client and generally a better overview what is available. This is not meant to replace but complement the existing global search functionality.

grafik

Following additions:

- `Date` filter: Filters for "Today", "Yesterday", "x days ago"
- `Suite` filter: Collects the different suites for choice
- `Client` filter: Collects the single clients for choice (I have decided to *not* include combined versions for better overview, this can still be done by fulltext search though)
- `Status` filter: Filter for "SUCCESS", "FAIL" and "TIMEOUT"

Side note on the date filter: it would have been nice to also be able to filter by a range of dates, e.g. a period of a whole week. However this is substantially more complex to implement and I have therefore decided to leave out.

### DIFF
--- a/cmd/hiveview/assets/index.html
+++ b/cmd/hiveview/assets/index.html
@@ -30,7 +30,8 @@
           <div id="loading" class="spinner-border text-secondary" role="status" style="width: 26px; height: 26px; display: none;"></div>
         </h2>
         <p id="page-text" style="display: none;">These test suites are available, and can be loaded. Click on 'Load' to load a certain suite.</p>
-        <table id="filetable" class="table table-bordered"></table>
+        <table id="filetable" class="table table-bordered">
+        </table>
       </div>
     </main>
   </body>
--- a/cmd/hiveview/assets/index.html
+++ b/cmd/hiveview/assets/index.html
@@ -30,8 +30,7 @@
           <div id="loading" class="spinner-border text-secondary" role="status" style="width: 26px; height: 26px; display: none;"></div>
         </h2>
         <p id="page-text" style="display: none;">These test suites are available, and can be loaded. Click on 'Load' to load a certain suite.</p>
-        <table id="filetable" class="table table-bordered">
-        </table>
+        <table id="filetable" class="table table-bordered"></table>
       </div>
     </main>
   </body>
--- a/cmd/hiveview/assets/lib/app-index.js
+++ b/cmd/hiveview/assets/lib/app-index.js
@@ -181,7 +181,6 @@ function selectWithOptions(api, colIdx) {
                 select.append( $('<option value="'+d+'">'+d+'</option>') );
             }
         } );
-    
 }
 
 function statusSelect(api, colIdx) {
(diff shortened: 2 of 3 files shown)


fjl:
Implementation looks very clean! I'll check it later.

holgerd77:
Took me 10 rounds or so to get this into a "clean" state. 😋 Thanks!

holgerd77:
Did you have a chance to take a quick look?

holgerd77:
I see all PRs getting merged around this, except this one. 😋 Is there anything blocking this?

To re-iterate with some additional notes:

So this is not throwing any additional errors when the additional JavaScript is loaded, see following browser console output:

grafik

So there should be no side effect on additional UI functionality (which can happen in JavaScript if one adds some functionality which breaks in between and then follow up initializations from previous code parts get interrupted).

I have also tested using all filters and then going back to "Select all" as well as using pagination back and forth with filtering, all with no problems.

Would be glad if this can get merged soon, relatively eager to use this myself! 🙂

fjl:
@holgerd77 being impatient on here is not recommended. PRs are not always merged in the order they were opened. I agree this is a nice improvement, and it will be merged as soon as I find the time to try it out.

holgerd77:
I'll try.

holgerd77:
Maybe you just need more reviewers for this repository? 🤔

holgerd77:
So aren't @marioevz or @spencer-tb reviewers here as well anyhow? Maybe they can take on this if you do not have spare ressources?

fjl:
So I tried it, and it kind of works, but there is also no real benefit over using the search field directly. I kind of expected that, if I select `rpc` in the suite column, that it would match on `rpc` exactly. But right now, it also shows `rpc-compat`.

fjl:
Pushed an attempt to fix it. I also think we should save these filters to the URL hash segment. It's not really a useful feature if I have to click the filter every time I reload the page.

fjl:
The date filter has questionable utility, so we should consider removing it, or turn it into something else. Our public instances are always running tests, but runs aren't meaningfully attributable to days because the duration of one test cycle depends on what else is in the cycle. 

Client versions update at the beginning of each cycle, but the start time of the suite is anywhere within the cycle duration. So if you filter for 'yesterday', you will get the suites that 'started yesterday', but that has nothing to do with the update time of the client. Instead of clicking through the 'days' you would be much better off just searching in the full list of runs.

What would be way more useful, is a filter on client versions. It requires some logic in the hiveview command to create an additional index, but should be totally doable.

fjl:
What I meant with the URL, is that it would enable bookmarking a query. It reminds me of this issue: https://github.com/ethereum/hive/issues/284

holgerd77:
Ah, yes, agree that would be nice. Nevertheless still holds that this would be too much work for me to be able to add atm.

Might look like a small thing, but I would estimate at the end that is 3+ hours of work.

holgerd77:
🙏

PR #812, opened 2023-08-04
clients: merge dockerfile

joohhnnn:
It seems that consolidating multiple files cases as comments within a single file might be more user-friendly for developers unfamiliar with Docker. (Embarrassingly, I modified Dockerfile.git and was puzzled as to why it didn't work. It took me two hours to finally understand why it wasn't functioning. LOL


fjl:
These dockerfiles are a part of the client build arguments feature. So they will actually be used and can't be deleted!
See here for more info about client build configuration: https://github.com/ethereum/hive/blob/master/docs/commandline.md#client-build-parameters

PR #822, opened 2023-08-09
simulators/ethereum/pyspec: Add Engine API error code checks for erroneous test cases

spencer-tb:
Fixes current false positive and false negative cases for fork transition tests. 

Engine API error codes have been added to the fixtures such they can be parsed within the simulator:
https://github.com/ethereum/execution-spec-tests/pull/248

If an error code exists within a test case the simulator will now check that it matches the expected response from the Engine API.


marioevz (approved):
LGTM!

PR #831, opened 2023-08-16
simulators/ethereum/engine: Coalesce fork configs

marioevz:
Brings together the configuration elements of the different forks into a single method.

Fixes an issue when running `ExchangeCapabilities` tests of an improperly configured genesis.


marioevz:
Ran `engine-withdrawals`, `engine-cancun`, `engine-exchange-capabilities` on geth and nethermind, and results look ok, merging.

PR #832, opened 2023-08-16
clients/erigon: Specify golang versioned image

marioevz:
Fixes an issue with Erigon not building properly in hivecancun host due to usage of the latest go version.


marioevz:
Tested to be working on hivecancun server and deployed. Merging.

Issue #810, opened 2023-08-03
Remaining Engine API Cancun Tests


marioevz:
Extracted list of remaining tasks for Cancun Engine API from https://github.com/ethereum/hive/pull/759
```[tasklist]
### Tasks
- [ ] Implement tests for https://github.com/ethereum/execution-apis/pull/426
- [ ] Field checking of new engine api methods, pre and post fork
- [ ] Sync Tests
- [ ] Parallelism Tests
- [ ] Auto-generate README.md from the information in the test case list
```

jochem-brouwer:
This seems currently untested:

- [ ] `forkchoiceUpdatedV2` throws `INVALID_PARAMS` if `parentBlockBeaconRoot` is provided
- [ ] `forkchoiceUpdatedV3` throws `UNSUPPORTED_FORK` if it is called at `< Cancun` time
- [ ] `getPayloadV3` throws `UNSUPPORTED_FORK` is a pre-Cancun payload is requested

PR #834, opened 2023-08-18
simulators/ethereum/engine: Add incorrect engine versions tests

marioevz:
Addresses missing GetPayloadV2/V3 and ForkchoiceUpdatedV2/V3 tests at fork time, described in comment: https://github.com/ethereum/hive/issues/810#issuecomment-1672182427

### DIFF
--- a/simulators/ethereum/engine/suites/cancun/tests.go
+++ b/simulators/ethereum/engine/suites/cancun/tests.go
@@ -725,7 +725,7 @@ var Tests = []test.SpecInterface{
 			`,
 		},
 
-		CancunForkHeight: 1,
+		CancunForkHeight: 0,
 
 		TestSequence: TestSequence{
 			SendBlobTransactions{
--- a/simulators/ethereum/engine/suites/cancun/tests.go
+++ b/simulators/ethereum/engine/suites/cancun/tests.go
@@ -537,6 +537,235 @@ var Tests = []test.SpecInterface{
 		},
 	},
 
+	// ForkchoiceUpdatedV3 before cancun
+	&CancunBaseSpec{
+		Spec: test.Spec{
+			Name: "ForkchoiceUpdatedV3 Set Head to Shanghai Payload, Nil Payload Attributes",
+			About: `
+			Test sending ForkchoiceUpdatedV3 to set the head of the chain to a Shanghai payload:
+			- Send NewPayloadV2 with Shanghai payload on block 1
+			- Use ForkchoiceUpdatedV3 to set the head to the payload, with nil payload attributes
+
+			Verify that client returns no error.
+			`,
+		},
+
+		CancunForkHeight: 2,
+
+		TestSequence: TestSequence{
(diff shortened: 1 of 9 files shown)


marioevz:
Hi @jochem-brouwer @g11tech, I tried to add the test described in the ACD on commit https://github.com/ethereum/hive/pull/834/commits/0b0390dd6450c8101c957ee8f3fb429443b531b8, could you let me know if this addresses the edge case? Thanks!

g11tech:
> Hi @jochem-brouwer @g11tech, I tried to add the test described in the ACD on commit https://github.com/ethereum/hive/pull/834/commits/0b0390dd6450c8101c957ee8f3fb429443b531b8, could you let me know if this addresses the edge case? Thanks!

Will check and update 👍

simulators/ethereum/engine/suites/cancun/tests.go
```diff
@@ -537,6 +537,235 @@ var Tests = []test.SpecInterface{
...
+	},
+
+	// ForkchoiceUpdatedV2 after cancun
+	&CancunBaseSpec{
+		Spec: test.Spec{
+			Name: "ForkchoiceUpdatedV2 To Request Cancun Payload, Zero Beacon Root",
+			About: `
+			Test sending ForkchoiceUpdatedV2 to request a Cancun payload:
+			- Payload Attributes uses Cancun timestamp
+			- Payload Attributes' Beacon Root zero
```
g11tech, line 665: would like to avoid interpreting beacon root (check against zero) to return error. not essential imo but also no strong opinions here

g11tech (changes requested):
@marioevz mostly looks amazing

the test case which i was requesting on the call was:

1. send fcUv3 with payload attributes with a parent beacon block root `x`
2. resenf fcUv3  with payload attributes with updated parent beacon block root `y` (lets say CL changed its head) but rest everything same

checks:

1. the payload id for 1 & 2 should be different,
2.  fetch the payload for new payload id recieved in 2nd fcUv3 request 
3. new payload v3 request with the new payload and the updated beacon block root `y` should return `VALID`

simulators/ethereum/engine/suites/cancun/tests.go
marioevz: My intention was to test that the client responds "Invalid Params" when the beacon root is provided for FcUv2, which should not contain a beacon root, trying to follow this line on v3's spec: "provided set of parameters and their fields strictly matches the expected one".

The zero beacon root was just to represent a non-nil value, any non-nil value should (in my opinion) result in the same error.

marioevz:
@g11tech I merged the new test and verified before and after EthJS fix and it correctly detected the issue/passes the test on fix!

simulators/ethereum/engine/suites/cancun/tests.go
g11tech: ahh ok thanks for clarification

PR #835, opened 2023-08-22
simulators/devp2p: enable discv5 test suite

fjl:
This adds support for running the discv5 test suite from go-ethereum's cmd/devp2p against EL and CL clients.


holgerd77:
For context: is this more/only for CL clients? EL clients are not supposed to support discv5 in any way, or do they? 🤔

fjl:
It's not mandatory, but we're definitely moving in the direction of it being required.

holgerd77:
Oh, that's interesting to hear. 🤯 Is there already an option to run Go Ethereum with discV5? And does it work if one does (on mainnet, e.g.)?

fjl:
The option is there, but it isn't enabled for peer discovery. We do want to make that work though.

holgerd77:
Would this be a topic for next ACD-E call (or P2P breakout) what your guys' ideas or rough plans are for this? Or is this still too early?

PR #837, opened 2023-08-30
simulators/ethereum/engine: Fix ForkID Tests

marioevz:
Fixes a false-positive result on the ForkID tests where Geth accepts the connection but disconnects within a second due to a mismatch in their local ForkID value.

The fix includes waiting for a second before sending a `ping` message and waiting for a `pong` response, where in the failing case, Geth responds with `disconnect` instead.

EthJS and Nethermind seem to be passing all tests.


lightclient (approved):
Geth is passing this after apply the patch from https://github.com/ethereum/go-ethereum/pull/28034

PR #838, opened 2023-09-04
[WIP] simulators/eth2/dencun: Dencun Interop Simulator and Builder Tests

marioevz:
This is a work in progress but the plan is to have a simulator that spawns a local testnet with one or many client combinations and include the following tests:
- Interop sanity tests between client combinations
- Interop syncing tests (Test blob syncing)
- Builder API tests using the mock-builder


marioevz:
Changed back to draft because it requires changes in #898, and they might not even be included in master until we find a proper solution.

marioevz:
I rebased and squashed most of the commits to be able to merge using a merge commit, so we preserve the multiple commits in the master branch.

I'm running locally tonight and should be able to merge tomorrow if all is good.

PR #842, opened 2023-09-05
clients/reth: add cancun fork in mapper

Rjected:
Adds cancun fork in mapper and documents shanghai / cancun timestamps in `reth.sh`


marioevz (approved):
LGTM :+1:

PR #861, opened 2023-09-08
clients/erigon: Fix `Dockerfile.git`

marioevz:
Fixes the `Dockerfile.git` on some tests where it needed to fetch the enode because curl was not installed.


marioevz:
Verified fix locally, merging now

PR #862, opened 2023-09-12
hiveview: Client URL based filtering

spencer-tb:
Adds a small change to filter all the simulations within `index.html` based on the client provided within the url.

For example, now `/?client=go-ethereum` will display only simulations by geth. This acknowledges that client names within simulation logs usually include extra metadata such as `go-ethereum_cancun-git`, hence the value specified in the url is matched against the string before the first `_`.

### DIFF
--- a/cmd/hiveview/assets/lib/app-index.js
+++ b/cmd/hiveview/assets/lib/app-index.js
@@ -59,6 +59,12 @@ function showFileListing(data) {
         suites.push(suite);
     });
 
+    // Filter based on client before rendering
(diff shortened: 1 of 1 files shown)


fjl:
I've implemented a more complete version of this in #864. Please try it. Main advantage of my thing is that you get a way of creating the URL by changing the select, whereas with this PR, you just sort of have to know that it can be filtered by URL.

spencer-tb:
Closing in favor of https://github.com/ethereum/hive/pull/864!

PR #863, opened 2023-09-14
clients/besu: sh - Remove legacy option from startup flags

Gabriel-Trintinalia:
Besu has implemented a new txpool format and will not start if any legacy flag is present.


marioevz (approved):
LGTM

PR #867, opened 2023-09-17
simulators/ethereum/engine: Add London non-zero-number test

marioevz:
Adds a single test where a pre-merge (by block number) fork is not zero. 

Catches https://github.com/ethereum/go-ethereum/issues/28134 (on v1.13.0) and passes on latest master.


marioevz:
Rebased and tested on latest. Will merge now.

PR #869, opened 2023-09-19
simulators/ethereum/engine: run tests in parallel

lightclient:
I applied the same parallelization technique as in the `ethereum/consensus` simulator.

Before parallelization, we see running some blob tests takes XXX:
```console
time go run . --client-file clients.yaml --sim ethereum/engine --sim.limit "engine-cancun/blob"
...
4.18s user 6.83s system 3% cpu 5:23.29 total
```
After parallelization, we see running good improvement with blob tests taking only XXX:
```console
time go run . --client-file clients.yaml --sim ethereum/engine --sim.limit "engine-cancun/blob"
...
4.06s user 6.48s system 5% cpu 2:59.13 total
```
In both examples, the simulator docker image is compiled, so that shouldn't play into the difference.

### DIFF
--- a/simulators/ethereum/engine/main.go
+++ b/simulators/ethereum/engine/main.go
@@ -3,170 +3,221 @@ package main
 import (
 	"fmt"
 	"math/big"
+	"os"
+	"strconv"
+	"sync"
 	"time"
 
 	"github.com/ethereum/hive/hivesim"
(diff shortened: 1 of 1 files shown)


marioevz (approved):
Awesome! Thanks so much!

PR #874, opened 2023-09-22
simulators/ethereum/engine: use generic error check for missing block

lightclient:
Right now, many of these RPC tests expect clients to return `-39001` if `safe` or `finalized` block is not found. Geth doesn't support custom RPC errors for the `eth` namespace so I really prefer if we just expect an error and be happy if we get either `39001` or `-32000`.

Screenshot 2023-09-22 at 06 57 23

### DIFF
--- a/simulators/ethereum/engine/suites/engine/rpc.go
+++ b/simulators/ethereum/engine/suites/engine/rpc.go
@@ -50,7 +50,7 @@ func (b BlockStatus) Execute(t *test.Env) {
 			number = Finalized
 		}
 		p := t.TestEngine.TestHeaderByNumber(number)
-		p.ExpectErrorCode(-39001)
+		p.ExpectError()
 	}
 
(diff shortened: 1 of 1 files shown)


fjl:
> Geth doesn't support custom RPC errors for the eth namespace

Can you clarify? Pretty sure we could fix this in go-ethereum if we wanted to.

lightclient:
@fjl sorry to clarify, we can of course support this, but it would be the only non-`-32000` error (afaik) in the `eth` API. So we don't really have specialized error codes like this yet. It's also extremely specialized to only be necessary while the merge transition is finalizing:

> Before the merge transition is finalized, any call querying for `finalized` or `safe` block MUST be responded to with `-39001: Unknown block` error'

I think it would be better to relax this test slightly and not verify exact error code.

marioevz (approved):

PR #883, opened 2023-09-26
Snooper client

jflo:
Example of a dumb client that will proxy through to the host network.

All traffic is output to stdout, but since clients request gzip, responses are illegible. Being able to switch that off would be handy.

the mapper.jq included is a besu specific thing, so I can just copy that from test output and run my IDE using it. Would be kinda cool if the test could output the genesis (mapped or not, idk) to a volume.


jflo:
closing as this was just a PoC and i don't have the bandwidth to finish it out

PR #884, opened 2023-09-26
simulators/ethereum: Update beacon root address 0x000F..Beac02

danceratopz:
Updates to the final EIP-4788 Beacon Root Address, cf https://github.com/ethereum/execution-spec-tests/pull/312.

PR #885, opened 2023-09-26
simulators/ethereum/engine: rm 0x prefix on 4788 code

lightclient:
Looks like in #884 there was a small issue where `0x` prefix was added to the code. This is not accepted by `common.Hex2Bytes`, which fails silently when the prefix exists.


marioevz (approved):
LGTM! Thanks!

PR #886, opened 2023-09-26
simulators/ethereum/engine: Fix several tests

marioevz:
## Changes Included
- Small fix to the setup of the generic versioning tests
- Fixes the tx replacement that affected one Cancun test
- Fixed the Unique ID test for the beacon root attribute
- Added a per-client map for payload IDs (allows collisions between clients for some clients that use a simple counter e.g. Erigon)


marioevz:
I verified locally and the tests are now fixed. Merging now.

PR #887, opened 2023-09-26
clients/teku-bn: Add presets to config

marioevz:
This was necessary for Teku to correctly start after `MAX_PER_EPOCH_ACTIVATION_CHURN_LIMIT` was added to Deneb.

I'm not sure if this is the correct approach or this could be a real issue.

cc @zilm13


marioevz:
Closing as this might be unnecessary.

PR #889, opened 2023-09-27
clients/besu: use ubuntu base image and install java

macfarla:
intermittent startup errors reported using older docker base image.
Aiming to use the same base image we use for building besu elsewhere. 
See https://github.com/hyperledger/besu/pull/5951


marioevz (approved):
LGTM, tested locally and seems to be working, thanks! :)

PR #890, opened 2023-09-27
clients/erigon: Increase allowed blobcount in remote txs

somnathb1:
Erigon now has a built-in spam protector for blob txs, this flag would increase the count of allowed blobs per account from the default 48 to a 1000


marioevz (approved):
LGTM

PR #894, opened 2023-10-02
graphql: add test case for eip-4844 fields

s1na:
The test was filled using the new `graphqltestgen fill` [feature](https://github.com/s1na/graphqltestgen#filling-tests)


NikhilSharmaWe:
Hey @s1na how did you get the blockchain data for eip-4844. What if someone want the same but for Shanghai version?

s1na:
@NikhilSharmaWe please check https://github.com/s1na/graphqltestgen

NikhilSharmaWe:
@s1na What if we cannot run get the node running ethereum.
Is there any resource from where I can get port of a blockchain of Shanghai version?

PR #895, opened 2023-10-02
simulators/ethereum/engine: Fixes from interop meeting

marioevz:
## Changes Included
- Invalid payload fields that can be detected by static check before sync are now optional, therefore `INVALID` and `SYNCING` can be accepted for payloads that can be detected to be invalid even before sync
- Invalid payloads with an invalid transaction now use more plausible values. E.g. `nonce < 2**64-1`, `gasLimit > 0`
- `ForkchoiceUpdatedV2 then ForkchoiceUpdatedV3 Valid Payload Building Requests` now does not send blob transactions before Cancun fork.

@lightclient @g11tech @acolytec3  @MarekM25 @marcindsobczak @yperbasis @jflo


marioevz:
Not sure who is the right person to tag from reth but tagging @Rjected since they've contributed to this repo before :)

marioevz:
I'll merge now since there have not been any comments.

PR #899, opened 2023-10-04
clients/teku-bn: Add rate limit parameters

marioevz:
Adds a rate limit parameter to teku, necessary for it to not rate limit peers based on RPC requests.

@tbenr


marioevz:
Hey @tbenr, I joined parameters `--Xp2p-gossip-scoring-enabled=false --Xpeer-rate-limit=100000 --Xpeer-request-limit=1000` under the umbrella hive term `"HIVE_ETH2_DISABLE_PEER_SCORING"`, which I will enable on blob p2p tests for teku. Let me know if this makes sense to you.

tbenr:
Hey @marioevz, sorry for the delay. It makes sense to me. The only doubt I have is disconnection if the peer misbehaves during rpc calls. I need to double check if in that case you need to specify it as a static-peer. I'll take a look tomorrow.

marioevz:
I will merge this and we can address the RPC disconnection later since there are no tests at the moment that make RPC calls from the p2p.

PR #900, opened 2023-10-04
simulators/ethereum/engine: Add test to re-org to mempool and back in

marioevz:
Adds a re-org test where a tx that was already included in a block is re-org'd back out to the mempool and the client needs to pick it out from the mempool to build the next block.

This hits an issue where besu was building invalid blocks because the tx was not preserving the blobs when it was sent back to the mempool again.

Geth is passing the test.

Nethermind, Erigon and EthereumJS are failing this test because the tx that was re-org'd out is not being included in the next block, so I think it's not making its way back to the mempool for some reason.

```
./hive --client-file ./configs/cancun.yaml --client <client> --sim ethereum/engine --sim.limit "engine-cancun/Transaction Re-Org, Re-Org Back In"
```

@MarekM25 @marcindsobczak @yperbasis @jflo @g11tech @acolytec3

### DIFF
--- a/simulators/ethereum/engine/suites/engine/reorg.go
+++ b/simulators/ethereum/engine/suites/engine/reorg.go
@@ -116,11 +116,19 @@ func (spec SidechainReOrgTest) Execute(t *test.Env) {
 }
 
 // Test performing a re-org that involves removing or modifying a transaction
+type TransactionReOrgScenario string
+
+const (
+	TransactionReOrgScenarioReOrgOut            TransactionReOrgScenario = "Re-Org Out"
+	TransactionReOrgScenarioReOrgBackIn         TransactionReOrgScenario = "Re-Org Back In"
+	TransactionReOrgScenarioReOrgDifferentBlock TransactionReOrgScenario = "Re-Org to Different Block"
+	TransactionReOrgScenarioNewPayloadOnRevert  TransactionReOrgScenario = "New Payload on Revert Back"
+)
+
 type TransactionReOrgTest struct {
 	test.BaseSpec
-	ReorgOut            bool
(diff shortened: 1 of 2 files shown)


jflo:
am just wrapping up some testing on this... besu will simply not re-add the transactions if we don't have the blobs they originally came with, am curious if other clients are pursuing more sophisticated implementations?

acolytec3:
EthereumJS just forgets the transaction at this point if its not resubmitted.  Is there some execution spec that says we should automatically put a txn back in the mempool?  It makes sense on the surface but quickly gets tricky in the case of a multiple block reorg.

g11tech:
> EthereumJS just forgets the transaction at this point if its not resubmitted. Is there some execution spec that says we should automatically put a txn back in the mempool? It makes sense on the surface but quickly gets tricky in the case of a multiple block reorg.

I think we shouldn't prune out the txs till there is finalization or till there is sufficient chain build on the block.but i somehow remember that we have such a test in txpool spec.

marioevz:
I just realized this was not merged.

> > EthereumJS just forgets the transaction at this point if its not resubmitted. Is there some execution spec that says we should automatically put a txn back in the mempool? It makes sense on the surface but quickly gets tricky in the case of a multiple block reorg.
> 
> I think we shouldn't prune out the txs till there is finalization or till there is sufficient chain build on the block.but i somehow remember that we have such a test in txpool spec.

@g11tech  I think it's not a big deal if the execution client forgets about a re-org'd tx, in reality the user will resubmit, and if the tx did make it into the chain, yet the execution client forgot about it, it's the consensus client's job to look it up in the beacon chain.

PR #901, opened 2023-10-05
simulators/ethereum/engine: Update post-merge forks to activate later in pre-merge test

spencer-tb:
Adds an update to the test here: https://github.com/ethereum/hive/pull/867, where a pre-merge fork is non-zero.

Previously, this test would start by loading a genesis config where the pre-merge fork (London) occured at the same block (block 1) as the post merge fork under test.

Now the genesis config starts with a pre-merge fork (London) and merge fork (Paris) both occuring on block 1. Then the post merge-fork under test occurs at block 2.

cc @marioevz @flcl42


marioevz (approved):
Looks really good :+1: 

Just could you run the set of tests on all clients to make sure nothing breaks before merging? Thanks!

spencer-tb:
Can confirm this doesn't add any additional failures on clients. Should be good to merge :D

PR #907, opened 2023-10-13
simulators/ethereum/engine: Fix naming, safe/finalized on reorg tests

marioevz:
## Changes Included
- Naming in "Invalid Missing Ancestor ReOrg" now does not repeat for some of the tests
- Safe and Finalized hash in re-org tests now always points to 0x00..00


g11tech (approved):
lgtm

PR #911, opened 2023-10-19
clients/lodestar-bn: Add optional flag to disable peer scoring

marioevz:
Adds a flag that can be enabled on a per-simulator basis to disable peer scoring (e.g. p2p blob gossiping tests).

@g11tech


fjl:
If the feature is causing issues, we should probably always disable it. I'm not a fan of adding these kind of overrides. The interface between simulators and clients should be minimal and universal. It's bad if only one client supports a setting, or if the setting is a niche thing that is only required in very special cases.

marioevz:
Ah I understand the comment now, I'm adding support for the same flag in Teku in this PR: https://github.com/ethereum/hive/pull/899

And the other clients should have something similar if possible, it's just not implemented.

This would be enabled on any test that requires some connection to a test p2p client.

marioevz:
It should be disabled by default, to allow the blobber to connect to the clients and not be disconnected we have two options:
- Add the blobber to the list of trusted peers (Lighthouse allows this as a parameter file): `HIVE_ETH2_TRUSTED_PEER_IDS`
- Disable peer scoring (Teku and lodestar): `HIVE_ETH2_DISABLE_PEER_SCORING`

We pass both things on the blob simulator as these variables, but what ends up happening in the client's script varies a bit per client, do you mean that it should not vary at all ?

g11tech (commented):
lgtm as far as the option usage goes :+1:

PR #924, opened 2023-10-20
hivechain: track preimages during generation

lightclient:
Currently preimages is turned off during chain generation. This causes the state dump to fail to output the genesis accounts:

```
WARN [10-20|17:37:46.743] Dump incomplete due to missing preimages missing=19
```


fjl:
Thanks! I was looking for the cause of that message but couldn't figure it out.

PR #923, opened 2023-10-20
hivesim: only log test and suite skips at high log levels

spencer-tb:
## Description

When debugging a single test,
```bash
... --sim.limit engine-cancun/"ForkchoiceUpdatedV2 To Request Shanghai Payload, Zero Beacon Root" --docker.output
```
more often than not the output is flooded with all the tests that are skipped, for example:
```
[4b993452] skipping test "Invalid VersionedHashes Version NewPayload (Cancun) (go-ethereum_cancun-git)" because it doesn't match test pattern engine-cancun/ForkchoiceUpdatedV2 To Request Shanghai Payload, Zero Beacon Root
[4b993452] skipping test "Invalid VersionedHashes Version NewPayload - Syncing (Cancun) (go-ethereum_cancun-git)" because it doesn't match test pattern engine-cancun/ForkchoiceUpdatedV2 To Request Shanghai Payload, Zero Beacon Root
[4b993452] skipping test "Invalid Incomplete VersionedHashes NewPayload (Cancun) (go-ethereum_cancun-git)" because it doesn't match test pattern engine-cancun/ForkchoiceUpdatedV2 To Request Shanghai Payload, Zero Beacon Root
[4b993452] skipping test "Invalid Incomplete VersionedHashes NewPayload - Syncing (Cancun) (go-ethereum_cancun-git)" because it doesn't match test pattern engine-cancun/ForkchoiceUpdatedV2 To Request Shanghai Payload, Zero Beacon Root
[4b993452] skipping test "Invalid Extra VersionedHashes NewPayload (Cancun) (go-ethereum_cancun-git)" because it doesn't match test pattern engine-cancun/ForkchoiceUpdatedV2 To Request Shanghai Payload, Zero Beacon Root
[4b993452] skipping test "Invalid Extra VersionedHashes NewPayload - Syncing (Cancun) (go-ethereum_cancun-git)" because it doesn't match test pattern engine-cancun/ForkchoiceUpdatedV2 To Request Shanghai Payload, Zero Beacon Root
[4b993452] skipping test "Build Payload with Invalid ChainID Transaction (BlobTransactions) (Cancun) (go-ethereum_cancun-git)" because it doesn't match test pattern engine-cancun/ForkchoiceUpdatedV2 To Request Shanghai Payload, Zero Beacon Root
[...]

PR #927, opened 2023-10-24
clients: only output full genesis file at high log levels

spencer-tb:
## Description

Similar to https://github.com/ethereum/hive/pull/923, when debugging more than one test, the genesis config for each test can flood the simulation output making it difficult to debug.

## Proposed Change
We only log the entire `genesis.json` config when the `HIVE_LOGLEVEL > 3`.

### DIFF
--- a/clients/besu/besu.sh
+++ b/clients/besu/besu.sh
@@ -63,9 +63,19 @@ esac
 FLAGS="--logging=$LOG --data-storage-format=BONSAI"
 
 # Configure the chain.
-jq -f /mapper.jq /genesis.json > /besugenesis.json
-echo -n "Genesis: "; cat /besugenesis.json
-FLAGS="$FLAGS --genesis-file=/besugenesis.json "
+mv /genesis.json /genesis-input.json
+jq -f /mapper.jq /genesis-input.json > /genesis.json
+FLAGS="$FLAGS --genesis-file=/genesis.json "
+
+# Dump genesis. 
+echo "Supplied genesis state:"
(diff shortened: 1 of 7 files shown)


clients/besu/besu.sh
```diff
@@ -63,9 +63,19 @@ esac
...
+FLAGS="$FLAGS --genesis-file=/genesis.json "
+
+# Dump genesis. 
+echo "Supplied genesis state:"
+if [ "$HIVE_LOGLEVEL" -lt 4 ]; then
+    jq 'del(.alloc[] | select(.balance == "0x123450000000000000000"))' /genesis.json
+else
+    # Log full genesis only at higher log levels.
+    cat /genesis.json
+fi
```
marioevz, line 77:
```suggestion
if [ "$HIVE_LOGLEVEL" -lt 4 ]; then
    echo "Supplied genesis state (trimmed, use --sim.loglevel 5 for full output):"
    jq 'del(.alloc[] | select(.balance == "0x123450000000000000000"))' /genesis.json
else
    # Log full genesis only at higher log levels.
    echo "Supplied genesis state:"
    cat /genesis.json
fi
```

marioevz (changes requested):
Maybe adding a small comment on the output so the users know that the output was trimmed and how to get the full version if they need it.

clients/besu/besu.sh
spencer-tb: Thanks! Added below

marioevz (approved):

PR #928, opened 2023-10-24
simulators/ethereum/engine: Trim log output of requests

marioevz:
Trims the logging of the round-trip requests to the client in order to not grow the log sizes too much.


spencer-tb (approved):
Awesome - LGTM!

PR #929, opened 2023-10-24
hivesim: add --sim.limit.exact hive flag

spencer-tb:
## Description

On occasion we may use a regex pattern that will run more tests than intended. 

The intention behind the `sim.limit` below may be to solely run a single test in isolation:
```bash
... --sim.limit engine-cancun/"InvalidPayloadAttributes: Missing BeaconRoot" --docker.output
```

Due to the increasing number of tests within the engine simulator alongside its programtic naming of tests, the above `sim.limit` will run and additional test using regex matching: 
```
InvalidPayloadAttributes: Missing BeaconRoot
InvalidPayloadAttributes: Missing BeaconRoot (Syncing)
```

## Proposed Change

Add an additional flag: `--sim.limit.exact` that can be used to specify the exact test to be ran (not using regex). This somewhat acts as an alternatice solution to https://github.com/ethereum/hive/pull/914.

Work in progress...

### DIFF
--- a/hive.go
+++ b/hive.go
@@ -26,6 +26,7 @@ func main() {
 		dockerOutput          = flag.Bool("docker.output", false, "Relay all docker output to stderr.")
 		simPattern            = flag.String("sim", "", "Regular `expression` selecting the simulators to run.")
 		simTestPattern        = flag.String("sim.limit", "", "Regular `expression` selecting tests/suites (interpreted by simulators).")
+		simTestExact          = flag.String("sim.limit.exact", "", "Exact `expression` match for tests/suites (interpreted by simulators).")
 		simParallelism        = flag.Int("sim.parallelism", 1, "Max `number` of parallel clients/containers (interpreted by simulators).")
 		simTestLimit          = flag.Int("sim.testlimit", 0, "[DEPRECATED] Max `number` of tests to execute per client (interpreted by simulators).")
 		simTimeLimit          = flag.Duration("sim.timelimit", 0, "Simulation `timeout`. Hive aborts the simulator if it exceeds this time.")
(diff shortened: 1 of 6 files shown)


fjl:
Note you can run a single test by anchoring the regular expression, e.g. `--sim.limit "engine-cancun/InvalidPayloadAttributes: Missing BeaconRoot$"

spencer-tb:
Perfect! Closing in favor of anchoring with regex :D

spencer-tb:
Ah, we have to use the following:
```
--sim.limit 'engine-cancun/Invalid PayloadAttributes: Missing BeaconRoot \(Cancun\) \(nethermind_cancun-git\)$'
```

Going to re-open and try to make the usage slightly easier without the escaped characters

holgerd77:
Can we see that we try to properly integrate something like this in the Hive documentation (might need some overhaul in general to some extend)? That would be really helpful! 🙂 🙏

fjl:
The docs are here: https://github.com/ethereum/hive/blob/master/docs/commandline.md#simulation-options

holgerd77:
> We can definitely add more info about `--sim.limit` if needed.

Yes, that would be nice. I am not tech-savvy enough to side-read out of the current docs how to run a single test e.g. with this `$` addition.

fjl:
Is this still a wanted feature?

spencer-tb:
Will look into adding the relavant information in a future docs overhaul.

fjl:
I think we could also add this in the hive command only. The 'exact' flag could just escape the regexp and also anchor it.

spencer-tb:
Ah okay, personally yes I do think it is useful to have.

I created a new PR here: https://github.com/ethereum/hive/pull/969 as I didn't realize force pushing to a branch meant you could not re-open it.

PR #902, opened 2023-10-06
simulators/ethereum/engine: Fix engine-cancun fcu version tests

spencer-tb:
## Description

There are a few bugs within some of the engine-cancun fcu version tests. This PR aims to fix them and add more clarity to the test descriptions. Some fixes are from @jangko within https://github.com/ethereum/hive/pull/930.

PR #930, opened 2023-10-24
simulators/ethereum/engine: Fix Cancun fcUV2 tests bugs

jangko:
1. "ForkchoiceUpdatedV2 To Request Shanghai Payload, Zero Beacon Root"
    - test desc: "Payload Attributes uses Shanghai timestamp"
    - bug: `ForkHeight: 1`, generated timestamp is Cancun.
    - fix: `ForkHeight: 2`, generated timestamp is Shanghai.

2. "ForkchoiceUpdatedV2 To Request Cancun Payload, Zero Beacon Root"
   - test desc: "Payload Attributes' Beacon Root zero"
   - bug: Payload Attributes' Beacon Root is not set/nil.
   - fix: set Payload Attributes's Beacon Root to zero.


jangko:
Close in favor of #902

PR #934, opened 2023-10-27
simulators/ethereum/engine: Minor fix to re-org tests

marioevz:
Sends the last known block to the CL Mock to the execution client in order for it to be on the expected head.

Previously, the test sent an FcU which overrode what the CL Mock last instructed, which resulted in some clients ending up on the incorrect head (Some clients ignored the directive because an FcU to an ancestor of the current head is optional according to the spec: https://github.com/ethereum/execution-apis/blob/main/src/engine/paris.md#specification-1).

cc @jochem-brouwer


jochem-brouwer:
image

Passes the specific test in question which raised this PR :smile: (will approve)

jochem-brouwer (approved):

PR #939, opened 2023-11-06
clients/ethereumjs: verbose logging

jochem-brouwer:
We have updated the way the client logs RPC messages. This change ensures that all RPC requests/responses are logged completely (without this, and with the new PR, the client would not none at all).

See https://github.com/ethereumjs/ethereumjs-monorepo/pull/3125


fjl:
BTW, if there is a way to disable colored log output, please open a PR for it too!

PR #940, opened 2023-11-07
ethereumjs: add faster local test dockerfile

jochem-brouwer:
This PR updates the local dockerfile such that changes to the local monorepo are built much quicker. This diverges from the normal Dockerfile which pulls our latest `master`, and uses `node` on the built version of the client instead of `ts-node` which uses the unbuilt client (but the built other libraries) to speedup this process.

### DIFF
--- a/clients/ethereumjs/Dockerfile.local
+++ b/clients/ethereumjs/Dockerfile.local
@@ -6,10 +6,81 @@ FROM node:18-alpine
 RUN apk update && apk add --no-cache bash git jq curl python3 gcc make g++ \
     && rm -rf /var/cache/apk/*
 
+RUN npm i -g ts-node
+
 # Default local client path: clients/ethereumjs/<ethereumjs-monorepo>
 ARG local_path=ethereumjs-monorepo
-COPY $local_path ethereumjs-monorepo
+
+
+COPY ${local_path}/package-lock.json ethereumjs-monorepo/package-lock.json
+COPY ${local_path}/package.json ethereumjs-monorepo/package.json
+
+COPY ${local_path}/.editorconfig ethereumjs-monorepo/.editorconfig
+COPY ${local_path}/.githooks ethereumjs-monorepo/.githooks
+COPY ${local_path}/.github ethereumjs-monorepo/.github
+COPY ${local_path}/.gitignore ethereumjs-monorepo/.gitignore
+COPY ${local_path}/.gitmodules ethereumjs-monorepo/.gitmodules
+COPY ${local_path}/.prettierignore ethereumjs-monorepo/.prettierignore
+COPY ${local_path}/.vscode ethereumjs-monorepo/.vscode
+COPY ${local_path}/Dockerfile ethereumjs-monorepo/Dockerfile
+COPY ${local_path}/README.md ethereumjs-monorepo/README.md
+COPY ${local_path}/codecov.yml ethereumjs-monorepo/codecov.yml
+COPY ${local_path}/config ethereumjs-monorepo/config
+COPY ${local_path}/docker ethereumjs-monorepo/docker
+COPY ${local_path}/eslint ethereumjs-monorepo/eslint
+COPY ${local_path}/lint-staged.config.js ethereumjs-monorepo/lint-staged.config.js
+COPY ${local_path}/package-lock.json ethereumjs-monorepo/package-lock.json
+COPY ${local_path}/package.json ethereumjs-monorepo/package.json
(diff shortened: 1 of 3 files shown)


clients/ethereumjs/Dockerfile.local
```diff
@@ -6,23 +6,94 @@ FROM node:18-alpine
...
+COPY ${local_path}/packages/evm/package.json ethereumjs-monorepo/packages/evm/package.json
+COPY ${local_path}/packages/genesis/package.json ethereumjs-monorepo/packages/genesis/package.json
+COPY ${local_path}/packages/rlp/package.json ethereumjs-monorepo/packages/rlp/package.json
+COPY ${local_path}/packages/statemanager/package.json ethereumjs-monorepo/packages/statemanager/package.json
+COPY ${local_path}/packages/trie/package.json ethereumjs-monorepo/packages/trie/package.json
+COPY ${local_path}/packages/tx/package.json ethereumjs-monorepo/packages/tx/package.json
+COPY ${local_path}/packages/util/package.json ethereumjs-monorepo/packages/util/package.json
+COPY ${local_path}/packages/verkle/package.json ethereumjs-monorepo/packages/verkle/package.json
+COPY ${local_path}/packages/vm/package.json ethereumjs-monorepo/packages/vm/package.json
+COPY ${local_path}/packages/wallet/package.json ethereumjs-monorepo/packages/wallet/package.json
```
fjl, line 53: I'm not a fan of splitting out all these COPY lines. I guarantee they will get outdated soon.

clients/ethereumjs/Dockerfile.local
```diff
@@ -6,23 +6,94 @@ FROM node:18-alpine
...
+COPY ${local_path}/packages/devp2p ethereumjs-monorepo/packages/devp2p
+COPY ${local_path}/packages/genesis ethereumjs-monorepo/packages/genesis
+COPY ${local_path}/packages/trie ethereumjs-monorepo/packages/trie
+COPY ${local_path}/packages/statemanager ethereumjs-monorepo/packages/statemanager
+COPY ${local_path}/packages/tx ethereumjs-monorepo/packages/tx
+COPY ${local_path}/packages/evm ethereumjs-monorepo/packages/evm
+COPY ${local_path}/packages/block ethereumjs-monorepo/packages/block
+COPY ${local_path}/packages/ethash ethereumjs-monorepo/packages/ethash
+COPY ${local_path}/packages/blockchain ethereumjs-monorepo/packages/blockchain
+COPY ${local_path}/packages/vm ethereumjs-monorepo/packages/vm
```
fjl, line 78: Same here

clients/ethereumjs/Dockerfile.local
jochem-brouwer: I can put all the files into a single `COPY` command, but some of these are directories. If I add those to the COPY command then the *contents* of that directory will get copied into the target folder, but it will not actually create the target directory. So if I `COPY ${local_path}/config ethereumjs-monorepo/` then it will not create a `config` folder, it will instead dump all the files inside the `ethereumjs-monorepo` folder. I have tried many combinations but I cannot seem to make the COPY command to create the folder.

jochem-brouwer: I am assuming you mean that you want to reduce the amount of COPY commands, right? So put all these COPYd files/directories into a single COPY command? Or do you mean that the file structure here might change? (Deleted/renamed files/directories?)

jochem-brouwer: Relevant: https://stackoverflow.com/questions/37715224/copy-multiple-directories-with-one-command

fjl: Yeah, I meant that the structure of the upstream sources can/will change. So the Dockerfile will require constant maintenance.

jochem-brouwer: I will cleanup these files since not all of these are necessary to run hive / the client. For our packages, it it not likely that we will add any packages in the near future, and even if we do so I would be very surprised if this would be more than 2 per year. However I can likely also prune that command.

jochem-brouwer: I have cleaned up the unnecessary files. However, for the other ones (our packages) I do not want to clean this up, since this means we have to re-install and re-build everything once any of our packages changes (and this is very slow). It is unlikely we add more than 2 packages in a year, and have not planned any new packages so far, so "constant updating" is (likely) not going to happen.

PR #942, opened 2023-11-08
simulators/devp2p: add support for new eth tests

fjl:
This is for making the new test suite in https://github.com/ethereum/go-ethereum/pull/28340 work in hive.

### DIFF
--- a/simulators/devp2p/Dockerfile
+++ b/simulators/devp2p/Dockerfile
@@ -3,7 +3,7 @@
 # Build devp2p tool.
 FROM golang:1-alpine as geth-builder
 RUN apk add --update git gcc musl-dev linux-headers
-RUN git clone --depth 1 --branch update-ethtest https://github.com/lightclient/go-ethereum.git /go-ethereum
+RUN git clone --depth 1 https://github.com/ethereum/go-ethereum.git /go-ethereum
 WORKDIR /go-ethereum
 RUN go build -v ./cmd/devp2p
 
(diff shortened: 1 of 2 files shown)


cmd/hivechain/generate.go
```diff
@@ -200,6 +202,14 @@ func (g *generator) setDifficulty(i int, gen *core.BlockGen) {
 	}
 }
 
+func (g *generator) setParentBeaconRoot(i int, gen *core.BlockGen) {
+	if g.genesis.Config.IsCancun(gen.Number(), gen.Timestamp()) {
+		var h common.Hash
+		rand.Read(h[:])
```
fjl, line 208: The chain won't be deterministic anymore with this. We could change to math/rand and static seed or just use a hash function or something.

cmd/hivechain/genesis.go
```diff
@@ -163,6 +164,10 @@ func (cfg *generatorConfig) createGenesis() *core.Genesis {
 	// dcc := hexutil.MustDecode("0x" + depositCode)
 	// g.Alloc[dca] = core.GenesisAccount{Code: dcc}
 
+	// Deploy beacon roots storage contract (EIP-4788).
+	asm4788 := common.FromHex("0x3373fffffffffffffffffffffffffffffffffffffffe14604d57602036146024575f5ffd5b5f35801560495762001fff810690815414603c575f5ffd5b62001fff01545f5260205ff35b5f5ffd5b62001fff42064281555f359062001fff015500")
```
fjl, line 168: I think you pasted this incorrectly, there is an extra zero byte at the end here.

fjl:
Besu has an issue loading the test chain again. It computes the wrong ForkID somehow.

PR #943, opened 2023-11-09
cmd/hive: Add `--sim.randomseed`

marioevz:
Adds flag `--sim.randomseed <number>` to hive command to pass environment variable `HIVE_RANDOM_SEED` to simulators in order for them to parse and generate a deterministic randomness source.

### DIFF
--- a/internal/libhive/run.go
+++ b/internal/libhive/run.go
@@ -205,6 +205,7 @@ func (r *Runner) run(ctx context.Context, sim string, env SimEnv) (SimResult, er
 			"HIVE_PARALLELISM":  strconv.Itoa(env.SimParallelism),
 			"HIVE_LOGLEVEL":     strconv.Itoa(env.SimLogLevel),
 			"HIVE_TEST_PATTERN": env.SimTestPattern,
+			"HIVE_RANDOM_SEED":  strconv.Itoa(env.SimRandomSeed),
 		},
 	}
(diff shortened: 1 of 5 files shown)


internal/libhive/run.go
```diff
@@ -199,13 +199,17 @@ func (r *Runner) run(ctx context.Context, sim string, env SimEnv) (SimResult, er
...
 
 	// Create the simulator container.
+	envVars := map[string]string{
+		"HIVE_SIMULATOR":    "http://" + server.Addr().String(),
+		"HIVE_PARALLELISM":  strconv.Itoa(env.SimParallelism),
+		"HIVE_LOGLEVEL":     strconv.Itoa(env.SimLogLevel),
+		"HIVE_TEST_PATTERN": env.SimTestPattern,
+	}
+	if env.SimRandomSeed > 0 {
+		envVars["HIVE_RANDOM_SEED"] = strconv.Itoa(env.SimRandomSeed)
```
fjl, line 209: We should just always define this in envVars, even if zero.

marioevz: Addressed in last commit. Simulators should now handle the random seed value being set to zero and behave accordingly if necessary.

fjl: But does it matter that it's zero? Don't we just always want deterministic randomness?

marioevz: Yes I agree, it should always start with the same value. I think we can even improve this a bit by appending the Rand object to the Testing object.

PR #944, opened 2023-11-09
simulators/ethereum/engine: Add randomness seed input

marioevz:
Adds parsing of `HIVE_RANDOM_SEED` introduced in https://github.com/ethereum/hive/pull/943 in order to be able to deterministically replicate previous runs using the same random seed.

`simulator.log` output now prints the randomness source seed used for a given run, which is basically just `time.Now().Unix()` if no predefined random seed was specified.

Also now each test case has its own unique randomness source in order to make the tests deterministic even when running in parallel.


marioevz:
Setting as draft because this PR depends on https://github.com/ethereum/hive/pull/943 to be merged first.

PR #945, opened 2023-11-15
[DRAFT] Add documentation generation

marioevz:
Adds the `--docs` flag to the hive binary, which has the following effects:
- Sets the `HIVE_DOCS_MODE` environment variable for the simulator container, which does the following:
  - The flag can be parsed by the `hivesim` library being used by (currently) all simulators and inhibit the execution of test functions that are not marked as `AlwaysRun`
  - The simulator will still send `StartSuite`, `EndSuite`, `StartTest` and `EndTest` requests to the hive server
- Using the `TestManager`, the hive server collects the test suites and test cases and then proceeds to generate a markdown document for each suite containing every single test case into the simulator folder

The following fields are added to many structures, including the `TestRequest`, `TestSuite` and `TestCase`:
- DisplayName: Allows specifying a different title for a test suite or test case to be displayed in the output markdown
- Category: Allows grouping together many different test cases and add sections in the markdown file

In its current state, simulators still execute each test case in full because the `hivesim` dependency that inhibits execution in docs mode needs to be updated once this change makes it into the master branch.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -9,6 +9,7 @@ require (
 	github.com/evanw/esbuild v0.18.11
 	github.com/fsouza/go-dockerclient v1.9.8
 	github.com/gorilla/mux v1.8.0
+	github.com/lithammer/dedent v1.1.0
 	golang.org/x/exp v0.0.0-20230905200255-921286631fa9
 	golang.org/x/net v0.17.0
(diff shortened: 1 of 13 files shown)


marioevz:
Closed in favor of #948

PR #946, opened 2023-11-16
clients/besu: Remove trusted setup

marioevz:
Removes the now-invalid trusted setup file.

@jflo


jflo (approved):
yep, this is how i run simulators against a local besu in a debugger.

PR #950, opened 2023-11-27
clients/teku-bn: Remove trusted setup file, add `--ignore-weak-subjectivity-period-enabled`

marioevz:
This PR changes two things:
- Removes the insecure trusted setup that was used for testing before the mainnet setup
- Adds `--ignore-weak-subjectivity-period-enabled` parameter in order to be able to run syncing tests, which have been failing as in here: https://hiveinterop.ethdevops.io/suite.html?suiteid=1701068738-f2e0369332162bd71f5efa50420da375.json&suitename=eth2-deneb-sync#


marioevz:
Hey @tbenr, could you let me know if adding `--ignore-weak-subjectivity-period-enabled` is a good enough solution in this case?

This is what the syncing test currently does:
We start the chain, and after a few epochs (2) we start a teku instance and it needs to sync to the head through one of the synced teku nodes, but it was failing with the error here: https://hiveinterop.ethdevops.io/viewer.html?suiteid=1701068738-f2e0369332162bd71f5efa50420da375.json&suitename=eth2-deneb-sync&testid=1&file=%2Fresults%2Fteku-bn_deneb%2Fclient-858adfde58c5214dc92d1e4db372c417864528facde3ac39166459b3aaea175e.log

Adding the parameter results in the test passing.

tbenr:
Hey @marioevz I think in this test it is fair to add it.

PR #964, opened 2024-01-04
simulators/ethereum/consensus: Fix `excessBlobGas` genesis

marioevz:
Fixes passing the correct non-zero `excessBlobGas` to clients' genesis files.


fjl:
Thanks!

PR #969, opened 2024-01-15
hive: Add --sim.exactmatch flag

spencer-tb:
## Description

Following this PR: https://github.com/ethereum/hive/pull/929

> On occasion we may use a regex pattern that will run more tests than intended. 
> 
> The intention behind the `sim.limit` below may be to solely run a single test in isolation:
> ```bash
> ... --sim.limit engine-cancun/"InvalidPayloadAttributes: Missing BeaconRoot" --docker.output
> ```
> 
> Due to the increasing number of tests within the engine simulator alongside its programtic naming of tests, the above `sim.limit` will run and additional test using regex matching: 
> ```
> InvalidPayloadAttributes: Missing BeaconRoot
> InvalidPayloadAttributes: Missing BeaconRoot (Syncing)
> ```

## Updated Proposed Change

Adds an additional boolean flag: `--sim.exactmatch`, that can be used to escape and anchor the regex expression specified within `--sim.limit`.

### DIFF
--- a/hive.go
+++ b/hive.go
@@ -26,6 +26,7 @@ func main() {
 		dockerOutput          = flag.Bool("docker.output", false, "Relay all docker output to stderr.")
 		simPattern            = flag.String("sim", "", "Regular `expression` selecting the simulators to run.")
 		simTestPattern        = flag.String("sim.limit", "", "Regular `expression` selecting tests/suites (interpreted by simulators).")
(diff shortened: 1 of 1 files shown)


spencer-tb:
@danceratopz @marioevz - is this valuable? Maybe useful for non-consume based simulators. Otherwise I can update the docs to show how you can use regex correctly for exact matches within `--sim.limit`.

PR #976, opened 2024-01-31
.circleci: add support for running ci's based off code modified specificed via path

KolbyML:
Here is a PR for allowing for certain CI's to run only when certain CI's are touched. I am proposing this change because when portal-hive code gets merged. We wouldn't want rust ci jobs running for non-rust related code.

@fjl here is my PR which will allow for running rust-ci's only on rust code. Would you be able to ``Use of setup workflows must be enabled in project settings (Project settings > Advanced -> Dynamic config using setup workflows)`` enable this.

I am going to close my current portal-hive -> hive PR since it is out of date. And maybe opened PR's to merge 1 thing at a time if that would be easier?

This pr is based off the instructions here https://github.com/CircleCI-Public/api-preview-docs/blob/path-filtering/docs/path-filtering.md#setup-workflows

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,79 +1,14 @@
 version: 2.1
+
+setup: true
+
 orbs:
-  go: circleci/go@1.5.0
-
-jobs:
-  # This job builds the hive executable and stores it in the workspace.
-  build:
-    docker:
-      - image: cimg/go:1.21
-    steps:
-      # Build it.
-      - checkout
-      - go/load-cache
-      - go/mod-download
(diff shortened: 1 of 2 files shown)


KolbyML:
@fjl here is my PR which will allow for running rust-ci's only on rust code. Would you be able to ``Use of setup workflows must be enabled in project settings (Project settings > Advanced -> Dynamic config using setup workflows)`` enable this.

I am going to close my current portal-hive -> hive PR since it is out of date. And maybe opened PR's to merge 1 thing at a time if that would be easier?

PR #994, opened 2024-02-08
simulators/portal: refactor sims into 1 sim multi suite format

KolbyML:
Felix made a great suggestion on how to structure our simulators/suites :D.

This PR fixes client trin-bridge as a new required param was added to it.

I also updated the rust version as one of our dependencies require it.


KolbyML:
@fjl ping for review

fjl:
I tested it, and this works, but there is an issue with hivesim-rs: it does not seem to interpret the `HIVE_TEST_PATTERN` environment variable. Note that simulator environment variables are documented here: https://github.com/ethereum/hive/blob/master/docs/simulators.md#simulator-environment-variables

So the idea with `HIVE_TEST_PATTERN` is that it's a text pattern passed by hive. The hive command-line tool has a flag `--sim.limit` to set this. The regexp is supposed to be interpreted as a `/` separated string containing regular expressions, such that the first segment matches the suite name and the remaining ones match the test name.

For example, you could run only the 'interop' suite like this: 

```
./hive --sim portal/history --sim.limit history-interop --client trin
```

You could even run a specific test:

```
./hive --sim portal/history --sim.limit 'interop/RecursiveFindContent.*block number 100' --client trin
```

KolbyML:
Yeah hivesim-rs doesn't have testmatch implemented yet was gonna do that in a follow up pr But I will try and get that done today.

Also we don't properly implement role so i will have to fox that when I get some time.

fjl:
The role thing is kind of optional. We added it as a convenience feature for the go hivesim. But I don't really like what we came up with there, and we don't use it a lot. So at some point this will get refactored anyway.

PR #995, opened 2024-02-09
simulators/ethereum/pyspec: Refactor, add sync payload verification

marioevz:
## Changes Included
- Cleanup to `main.go` and `runner.go`: Refactor some types and add methods directly in the structures where appropriate
- Reads the sync header introduced by https://github.com/ethereum/execution-spec-tests/pull/431

cc @spencer-tb let me know what you think, injecting an empty block to our eest tests might be too much but I'm unsure

### DIFF
--- a/simulators/ethereum/pyspec/runner.go
+++ b/simulators/ethereum/pyspec/runner.go
@@ -195,7 +195,7 @@ func (tc *TestCase) run(t *hivesim.T) {
 
 	end := time.Now()
 
-	if tc.FailedErr == nil {
+	if false { // TODO: Activate only on --sim.loglevel > 3
 		t.Logf(`test timing:
 			setupClientEnv %v
  			startClient %v
--- a/simulators/ethereum/pyspec/runner.go
+++ b/simulators/ethereum/pyspec/runner.go
@@ -16,6 +16,10 @@ import (
 	"github.com/ethereum/hive/simulators/ethereum/engine/globals"
 )
 
+var (
+	SyncTimeout = 10 * time.Second
+)
+
 // loadFixtureTests extracts tests from fixture.json files in a given directory,
 // creates a testcase for each test, and passes the testcase struct to fn.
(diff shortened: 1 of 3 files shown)


spencer-tb:
A very welcome refactor from my side!

I think we could tweak a few more things, but its likely extra overhead when we port to EEST.

I commented some more on: https://github.com/ethereum/execution-spec-tests/pull/431, and tried this out with some blob tx tests - it worked well.

I wonder if we should also check the syncing between different clients?

simulators/ethereum/pyspec/runner.go
```diff
@@ -99,123 +92,110 @@ func (tc *testcase) run(t *hivesim.T) {
...
+			select {
+			case <-timeoutCtx.Done():
+				tc.Fatalf("timeout waiting for sync of secondary client")
+			default:
 			}
+			time.Sleep(time.Second)
 		}
 	}
+
 	end := time.Now()
```
spencer-tb, line 197: I don't think it would hurt if we used `ValidatePost()` here for the second client as well

spencer-tb (commented):
We also have the test timing at the end:
```
test timing:
[3032d996]                      setupClientEnv 32.882µs
[3032d996]                      startClient 692.127172ms
[3032d996]                      sendAllPayloads 4.700414ms
[3032d996]                      setNewHeadOfChain 1.731108ms
[3032d996]                      checkStorageValues 1.743853945s
[3032d996]                      totalTestTime 2.442445521s
```

I think we could remove it in the PR - unless you find it useful (I certainly don't :D)

simulators/ethereum/pyspec/runner.go
marioevz: The problem was that, technically it's not the same post, this is due to the extra syncing block.
So I was a little bit afraid to activate the post check here without giving it enough thought.

We could revisit it once we start marking some more tests as sync-verified.

PR #974, opened 2024-01-29
simulators/ethereum/engine: Fixes for execution-apis#498

marioevz:
## Description
Fixes issues on tests from changes described on [execution-apis/pull/498](https://github.com/ethereum/execution-apis/pull/498)

Fixes #961

## Changes included
- Change expected error from `INVALID_PARAMS_ERROR` to `INVALID_PAYLOAD_ATTRIBUTES` on following tests:
  - ForkchoiceUpdatedV3 To Request Shanghai Payload, Null Beacon Root
  - ForkchoiceUpdatedV2 To Request Shanghai Payload, Non-Null Beacon Root
  - ForkchoiceUpdatedV2 To Request Cancun Payload, Non-Null Beacon Root
- Expectation of test "Invalid PayloadAttributes, Missing BeaconRoot, Syncing=True" is modified according to point 7 of https://github.com/ethereum/execution-apis/blob/main/src/engine/paris.md#specification-1, which now strictly specifies that no error should be returned on an invalid payload attributes when the applied forkchoice state is `SYNCING` or `INVALID`
- Check for correct error on "Inconsistent .* in ForkchoiceState" tests, "-38002: Invalid forkchoice state" according to point 6 of https://github.com/ethereum/execution-apis/blob/main/src/engine/paris.md#specification-1

PR #996, opened 2024-02-10
simulators/ethereum/engine: Expect error code for invalid payload attributes tests

spencer-tb:
## Description

Following this PR: https://github.com/ethereum/hive/pull/974

Some clients are passing: `"Invalid PayloadAttributes, Missing BeaconRoot, Syncing=False"` returning the invalid parameters error code -32602, this should be -38003. Although the purpose of the test is to check that the fcu head is updated, the following change adds further verification by additionally expecting the correct error code response.


marioevz (approved):
Thanks!

PR #1003, opened 2024-02-28
simulators/eth2: Fix Docker golang version

marioevz:
Fixes the golang docker image version to `golang:1.20-alpine` instead of `golang:1-alpine`, since the former now redirects to `golang:1.21-alpine` which breaks compilation.


marioevz:
Verified working locally, merging now.

PR #1004, opened 2024-02-29
clients/nethermind: enable snap serving

spencer-tb:
## Description

Adds the `$HIVE_NODETYPE` env var to the nethermind start up script. Sets the appropriate flags if the node type is set to snap.

Passes the previously failed sync tests: https://hivetests.ethdevops.io/#suite=snap&client=nethermind

### DIFF
--- a/clients/nethermind/mkconfig.jq
+++ b/clients/nethermind/mkconfig.jq
@@ -109,6 +109,9 @@ def base_config:
       "BlocksDir": "/blocks",
       "KeysDir": "/keys"
     },
+    "Sync": {
+      "SnapServingEnabled": true,
+    },
   }
 ;
 


fjl:
Why not enable it by default?

fjl:
I mean, why should we only enable this when `HIVE_NODETYPE == "snap"`, can't it be enabled regardless of that variable?

spencer-tb:
> I mean, why should we only enable this when `HIVE_NODETYPE == "snap"`, can't it be enabled regardless of that variable?

Yeah, we should set `FLAGS="$FLAGS  --Sync.SnapSync true"` by default!

I believe snap serving performance is still being improved, and is slow-ish on the nethermind master branch, hence adding the snap serving flag for the specified node type. Do we have any tests for snap serving as opposed to syncing?

spencer-tb:
Set both flags to default using the mapper following @asdacap's PR

asdacap:
You only have one flag now, and you can turn it on all thetime.

fjl:
exciting times!

PR #1020, opened 2024-03-04
simulators/ethereum/engine: Optimize Blob generation, KZG pre-compute

marioevz:
Optimizes blob transaction generation in test simulations by optimizing the deterministic blob generation, and also embedding pre-computed KZG commitments and proofs.

cc @spencer-tb


marioevz:
Example test - Parallel Blob Transactions:
`BlobDataGenerator` execution time before:
- 0.7458 s on average
- 50 executions
- 37.29 s in total

`BlobDataGenerator` execution time after:
- 0.00199 s on average
- 50 executions
- 0.0997 s in total

marioevz:
Ran locally with all clients. Merging now to include changes in [hivecancun](https://hivecancun.ethdevops.io/).

PR #1021, opened 2024-03-04
circleci: Use default ubuntu machine

marioevz:
Circleci smoke-test begun failing on march 4th as part of their brownout schedule of deprecated images.

This PR changes the machine to the default one as recommended in their post: https://discuss.circleci.com/t/linux-image-deprecations-and-eol-for-2024/50177/9


marioevz:
@fjl although not recommended in the post from circle ci let me know if you'd like to pin to a specific version.

PR #1024, opened 2024-03-05
Enable snap serving on nethermind

asdacap:
- Add` flag to turn on snap serving on nethermind.
- To test `./hive --client nethermind --sim devp2p --sim.limit "snap/*" --sim.parallelism 32`


asdacap:
Was not aware of #1004

PR #1027, opened 2024-03-05
clients/nimbus-bn,vc: Use ethpandaops prebuilt docker images

marioevz:
Currently nimbus images take too long to generate and they seem to randomly trigger a full rebuild even when the image was just recently built.

This PR changes the default Dockerfile to use the prebuilt ethpandaops images to speed up test execution while leaving the original file as the optional `Dockerfile.git` to be able to build from source and run hive tests.


fjl:
I don't like it because the images are not official. But the long build time is also annoying.

PR #1029, opened 2024-03-05
clients/fluffy: use --networks cli option in fluffy client start script

kdeme:
The cli option was added to individually enable networks in Portal for fluffy client.


kdeme:
cc @KolbyML

PR #1052, opened 2024-03-14
simulators/ethereum/engine/cancun: Maintain devp2p client peering

spencer-tb:
# Description

- Fixes the engine-cancun devp2p test: `Request Blob Pooled Transactions` for geth.
- After the following PR: https://github.com/ethereum/go-ethereum/pull/29026, to obtain a `NewPooledTransactionHashes` devp2p msg for blob txs, we must now peer first before sending blob txs.
- Adds an additional test where we send multiple blob txs instead of a single blob tx.

### DIFF
--- a/simulators/ethereum/engine/suites/cancun/steps.go
+++ b/simulators/ethereum/engine/suites/cancun/steps.go
@@ -27,6 +27,7 @@ import (
 type CancunTestContext struct {
 	*test.Env
 	*TestBlobTxPool
+	DevP2PConnections map[uint64]*devp2p.Conn
 }
 
 // Interface to represent a single step in a test vector
@@ -757,6 +758,8 @@ func (step SendModifiedLatestPayload) Description() string {
 type DevP2PClientPeering struct {
 	// Client index to peer to
 	ClientIndex uint64
+	// Maintain connection for later steps
+	MaintainConnection bool
 }
 
 func (step DevP2PClientPeering) Execute(t *CancunTestContext) error {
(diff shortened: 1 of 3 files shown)


simulators/ethereum/engine/suites/cancun/steps.go
```diff
@@ -802,13 +814,17 @@ func (step DevP2PClientPeering) Execute(t *CancunTestContext) error {
 }
 
 func (step DevP2PClientPeering) Description() string {
-	return fmt.Sprintf("DevP2PClientPeering: client %d", step.ClientIndex)
+	return fmt.Sprintf("DevP2PClientPeering: client %d, connection maintained: %b ", step.ClientIndex, step.MaintainConnection)
```
spencer-tb, line 817: Ahh will update this for `DevP2PRequestPooledTransactionHash` too

simulators/ethereum/engine/suites/cancun/steps.go
```diff
@@ -27,6 +27,7 @@ import (
 type CancunTestContext struct {
```
marioevz, line 27: Could we take this opportunity to rename this to something more generic, like `TestContext`.

And also create a `func NewTestContext() *TestContext {` that performs all init operations required such as the `t.DevP2PConnections = make(map[uint64]*devp2p.Conn)` that we do in line 782.

marioevz: We could also create a `func (t *TestContext) Close() error {` that we could `defer t.Close()` after we call `NewTestContext()` when creating the context to do some proper cleanup such as the closing of the connections that are still open in `t.DevP2PConnections`.

marioevz (commented):
Some comments, thanks!

simulators/ethereum/engine/suites/cancun/steps.go
spencer-tb: Very happy to extend to this!

My only rebutle is that we'd be adding the extra initialization steps for all the CancunBaseSpec tests, where its only required for 2 of them.

Maybe we have a generic `TestContext` and additionally a specific `DevP2PTestContext`, the latter having the init steps you mentioned. Where we only init these steps for the DevP2P tests?

simulators/ethereum/engine/suites/cancun/steps.go
spencer-tb: Added here: https://github.com/ethereum/hive/pull/1052/commits/5178de4f0d95e988a015e862dd55886e1e7790b5

simulators/ethereum/engine/suites/cancun/steps.go
spencer-tb: I added you suggestions in: [25ffdf9](https://github.com/ethereum/hive/pull/1052/commits/25ffdf9e225b550f693787a07a680e25ff0c256f)

Let me know what you think :)

PR #1061, opened 2024-03-20
all: use golang 1.22 in Dockerfiles

Rjected:
Our hive pyspec builds recently broke because the images were using 1.20, and `go.mod` files used go toolchains, which were released in 1.21. This just changes any images using 1.20, to using a 1.22 image.


holiman:
Thought: wouldn't it be better to use `latest` rather than pin versions?

PR #1074, opened 2024-04-04
cmd/hiveview: Update max entries on hiveview

spencer-tb:
Currently the hiveview public CI only displays up to 200 entries.

In some cases this means we only display logs for the past 3-4 days, even though we have logs for many more.

I updated this manually within https://hivetests.ethdevops.io/ and note it now stores more than 200 entries.

This is a quick hack, we could alternatively set this value as a cmd line option when serving or deploying `./hiveview`.

### DIFF
--- a/cmd/hiveview/listing.go
+++ b/cmd/hiveview/listing.go
@@ -14,7 +14,7 @@ import (
 	"github.com/ethereum/hive/internal/libhive"
 )
 
-const listLimit = 200 // number of runs reported
+const listLimit = 10000 // number of runs reported
 
 // generateListing processes hive simulation output files and generates a listing file.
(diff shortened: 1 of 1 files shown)


fjl:
The reason why this limit exists is because we want to limit the summary file size for quick load. If we index all runs, it takes a lot longer to load initially. 

There are a couple possible fixes. One would be creating a separate file for the full index, and then loading that asynchronously. Or we could change the UI to parse the file and update the display while it is loading, to reduce the initial display latency.

spencer-tb:
This makes sense thanks. I'll look into the first option you mention next week:
> One would be creating a separate file for the full index, and then loading that asynchronously.

skylenet:
Closing this PR. There's now a `-limit` flag on the `hiveview` command which allows doing this. By default it's 200.

PR #1103, opened 2024-05-08
clients/trin-bridge: update outdated bridge args

njgheorghita:
we no longer support the `--node-count` flag inside trin-bridge


njgheorghita:
@KolbyML

PR #1104, opened 2024-05-08
Fix ethereumjs docker build steps

acolytec3:
Fixes the docker build step for `ethereumjs` so hive can run again.

I've run this locally with both `dockerfile` and `dockerfile.local` and hive appears to run again.


acolytec3:
@g11tech, could you verify this when you have a chance?

g11tech (approved):
lgtm

acolytec3:
Thanks!  Now we see if Hive starts working again

PR #1110, opened 2024-05-31
configs: replace cancun config for prague

spencer-tb:
1) Removes the `configs/cancun.yaml` file as its now redundant.
2) Adds the `configs/prague.yaml` file.


spencer-tb:
@marioevz

spencer-tb:
Set to draft. We shall wait until clients finalize branches for devnet-1.

PR #1112, opened 2024-06-03
fix: remove blob fields from ethereum/sync newPayloadV2

Rjected:
Reth does not support these fields as `null` when using newPayloadV2, and AFAIK it does not need to. We currently do not progress in this test because we report the request as malformed. This removes the `blobGasUsed` and `excessBlobGas` fields from the `ethereum/sync` test.


fjl:
Ahh, it looks like this is a bug in the hivechain tool. I'll get that fixed!

PR #1113, opened 2024-06-06
simulators/portal/beacon: Remove incomplete gossip/offer tests

kdeme:
In order to test offer/gossip in the Portal beacon network, the node needs to first also do light client sync to be able to validate the offered content. This requires these test to be adjusted to provide the proper light client data first in order to be able to sync. Remove these incomplete tests for now to avoid false failures.

### DIFF
--- a/simulators/portal/beacon/src/suites/interop.rs
+++ b/simulators/portal/beacon/src/suites/interop.rs
@@ -15,7 +15,6 @@ use itertools::Itertools;
 use serde_json::json;
 use serde_yaml::Value;
 use std::collections::HashMap;
-use tokio::time::Duration;
 
 // This is taken from Trin. It should be fairly standard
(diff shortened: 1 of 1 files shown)


kdeme:
cc @KolbyML

KolbyML:
Wouldn't the PR be to do the light client sync? I am not sure if we are in a rush to remove these, as they aren't ran on the main portal hive instance anyways

kdeme:
> Wouldn't the PR be to do the light client sync? I am not sure if we are in a rush to remove these, as they aren't ran on the main portal hive instance anyways

Well, with these invalid tests removed, all other tests pass for Fluffy and thus could be moved to main portal hive.

A PR that actually fixes the tests would be better of course, but that shouldn't block the actual test suite to show proper results.

PR #1117, opened 2024-06-06
simulators/rpc: make rpc suite post-merge

Rjected:
This makes the `ethereum/rpc` suite post-merge by setting the genesis and the environment variables. This is sort of a reth-specific fix, since our dev mode only works if the genesis is post merge. However, it doesn't seem like this should break other clients either.


marioevz:
Hey @fjl do you know if it will be possible to merge with the false-positive found in the GitGuardian checks?

fjl:
Yeah, sure!

fjl:
I think this PR will break the tests for some other clients though, since they won't be producing blocks after the merge happened. Updating it properly for the merge is a lot of work, I tried...

I think we should actually deprecate this suite in favor of ethereum/engine, or maybe integrate some of the tests here into it. We also have rpc-compat which explores a lot of API edge cases. The ethereum/rpc suite was supposed to check end-to-end use cases of the RPC API, i.e. check functionality that normal apps actually use.

Rjected:
> I think this PR will break the tests for some other clients though, since they won't be producing blocks after the merge happened. Updating it properly for the merge is a lot of work, I tried...
> 
> I think we should actually deprecate this suite in favor of ethereum/engine, or maybe integrate some of the tests here into it. We also have rpc-compat which explores a lot of API edge cases. The ethereum/rpc suite was supposed to check end-to-end use cases of the RPC API, i.e. check functionality that normal apps actually use.

Makes sense, I'd be very supportive of deprecating this suite / incorporating the logic into other suites. We do run the other test suites as well, so I'm going to close this PR

PR #1118, opened 2024-06-11
clients/fluffy: Add --disable-state-root-validation to Fluffy startup options

web3-developer:
Add --disable-state-root-validation to Fluffy startup options to support Portal state hive tests.

I'm working on updating Fluffy in order to pass the portal state hive tests and this new parameter is now required when running the state tests against Fluffy.


kdeme (approved):
cc @KolbyML

KolbyML (approved):
:shipit:

PR #1122, opened 2024-06-18
add legacy-consensus simulators

winsvega:
A bit of copy paste, but its ok because legacy tests won't be changing. 
So this adds simulators 

`legacy-consensus`  -  to run all the tests we had at the times generated from Frontier to Constantinople revisions
`legacy-consensus-cancun` - to run all the tests from Constantinople up to Cancun

and from >=Cancun and beyond the tests might change and will be migrated to Pyspecs. so the format might change. 
those will be fun by `consensus` simulator regulary. those tests are all vectors we have in Cancun only revision. so each run will save time. 

then legacy- ones could be run once a week or so. this will save hive stats the running time.


fjl:
It's not great to simply copy the consensus simulator including its implementation. We could just integrate it into the existing consensus simulator as a separate test suite.

winsvega:
The legacy simulators should run not on a daily basis.

Ideally, the logic should be abstracted out.
Because only the folder change

winsvega:
@fjl ok, how about this approach? 
https://github.com/ethereum/hive/pull/1130

PR #1125, opened 2024-06-20
internal/libdocker: specify API version

fjl:


KolbyML (approved):
image

In this image I show running the pull command before this PR and after on a newer version of docker.

Given the response I am assuming it worked and checking desktop it looks like it worked. I can't swear on my life, but it looks good

KolbyML:
I inspected my docker containers using podman and can confirm pull did work. Very cool 🚀

KolbyML:
> Can it run your simulator though? A downside with this change is that any Docker interaction might now fail because we hard-code the API version, but may accidentally send a client option that requires a later version. So we need to test all the interactions including running containers. It did work in my testing as well, just want to make sure it doesn't break for you.

Just ran our history simulator and it is running fine 👍

PR #1130, opened 2024-06-27
simulators/consensus: split test suites

winsvega:
adds --sim.argument option to be set as env variable inside simulator dockers
this to run consensus tests on different revisions:
```
# run develop bc tests
./hive --sim ethereum/consensus --client go-ethereum  --sim.limit "/add11" 

# run legacy tests from constantinople fix to cancun
./hive --sim ethereum/consensus --client go-ethereum  --sim.limit "/add11"  --sim.argument "legacy-cancun"

# run legacy tests from frontier to constantinople
./hive --sim ethereum/consensus --client go-ethereum  --sim.limit "/add11"  --sim.argument "legacy"
```

we only need to regulaly run develop tests. and sometimes legacy tests. to save execution time.


winsvega:
@fjl how about this way?

fjl:
I think what we should do is, expose the test groups as "suites", which is an existing concept in hive. Then you could it like

```
# run legacy tests from constantinople fix to cancun
./hive --sim ethereum/consensus --client go-ethereum  --sim.limit "legacy-cancun/add11"
```

PR #1131, opened 2024-06-27
clients/ultralight: set network radius in script

ScottyPoi:
Updates the startup script for ultralight at `clients/ultralight/ultralight.sh`

Adds command line flags to set the network radius for ultralight nodes.  Recent updates to Ultralight code cause some hive tests to fail if these flags are not set.

### DIFF
--- a/clients/ultralight/ultralight.sh
+++ b/clients/ultralight/ultralight.sh
@@ -4,7 +4,7 @@
 set -e
 
 IP_ADDR=$(hostname -i | awk '{print $1}')
-FLAGS=""
+FLAGS="--radiusHistory=256 --radiusState=256 --radiusBeacon=256"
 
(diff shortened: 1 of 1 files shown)


KolbyML:
I am confused why do we need magic flags, to pass tests which should pass with default settings?

ScottyPoi (commented):
> I am confused why do we need magic flags, to pass tests which should pass with default settings? 

what are default settings?  is there an expected default radius?

KolbyML:
> > I am confused why do we need magic flags, to pass tests which should pass with default settings?
> 
> what are default settings? is there an expected default radius?

Wait how did ultralights radius used to work and how does it now?

Trin/Fluffy/Shisui start at 100% and start lowering once their database is filled to the storage limit

ScottyPoi:
Okay.  I thought about it, and I'll adjust things in the ultralight code instead.

PR #1133, opened 2024-07-03
simulators/ethereum/engine: lower withdrawals test timeout to 240s

joshieDo:
Lowers the withdrawal test timeout to a more sane value


fjl:
@marioevz

marioevz (approved):

PR #1146, opened 2024-08-07
simulators/portal: update ethportal-api version

KolbyML:
Updates ethportal api so state tests use the latest version of ethportal-api which supports the latest state spec

@web3-developer


web3-developer:
@KolbyML Reviewed. These changes look good and all the state tests are passing for me when I run locally, both when running with Fluffy and Trin. The Fluffy changes to support addressHash instead of address in the state content keys have already been merged as well.

PR #1148, opened 2024-08-07
simulators/ethereum/engine: Fix state scheme issue in Geth node

### LINKED ISSUE #1135: `TEST ISSUE` in some missing ancestor reorg tests
When running:
```
Invalid Missing Ancestor Syncing ReOrg, GasLimit, EmptyTxs=False, CanonicalReOrg=True, Invalid P8
```

I get:
```
FAIL (Invalid Missing Ancestor Syncing ReOrg, GasLimit, EmptyTxs=False, CanonicalReOrg=True, Invalid P8): TEST ISSUE - Failed to set invalid block: triedb layer [0x5bb856b70f63d46b95dff7d2b07700194311a87c8a7494ad0de0e72de23d624b] is disk layer
```

We caught these when testing recent updates to our CI which should improve catching individual test regressions. I'm not really sure how to fix this, but it says `TEST ISSUE` and fails on reth / geth / erigon, so I assume this is a hive issue.

marioevz:
Fixes #1135 

Issue had to do with the new path state scheme, and taking a look at the differences between our `SetBlock` routine and Geth's `writeBlockWithState`, it seems like the `triedb` commitments are skipped in the new scheme:
https://github.com/ethereum/go-ethereum/blob/aa55f5ea200dfd07618fdf658d9d2741c3b376a8/core/blockchain.go#L1481-L1485

Solution suggested by this PR is to (1) hard-code the configuration to use the new Path scheme from now on, and (2) if we are using this Path scheme, skip the trie database commitments.

All the "Invalid Missing Ancestor" tests from the engine simulator are now working.


marioevz:
@Rjected please let me know if this fixes the issue for you.

Rjected:
> @Rjected please let me know if this fixes the issue for you.

yep, I no longer get `TEST ISSUE`, thanks!

PR #1150, opened 2024-08-15
hivesim-rs: allow test data to be generic

morph-dev:
## Problem

Currently, the `hivesim-rs` allows only specific type as a `TestData`. This is limiting in many ways:
- Many of the current simulations don't use any test data
- `TestData` only contains `String` objects, which require simulations to serialize/deserialize objects unnecessary
- Some simulations need exactly one content key-value pair, but they are using `Vec` because that's how it's defined
- When simulations need new specific type, they would first need to update `hivesim-rs` in one PR, in order to use that commit as dependency in their own PR.

## Solution

Make test data generic. This removes all above problems without any downsides (in my opinion).
Example of how portal simulators would be updated can be seen here: https://github.com/ethereum/hive/pull/1151/commits/3cd8c95ac4ad30f624579552af2b348c2fed60c8 (see #1151)


KolbyML (approved):
:shipit:

PR #1152, opened 2024-08-16
simulators/portal: store block header before gossip and offer state requests

morph-dev:
The state network should verify that content is anchored to the canonical blockchain. For that, in needs access to the block header, which we have to seed at the start of the test.

Now that some tests require more than one portal network running, I refactored how that is configured.


morph-dev:
Note regard GitGuardian failing check: This isn't an issue as these "secret" are just for test purposes only (and it's not something introduced in this pr).

KolbyML (approved):
:shipit:

PR #1165, opened 2024-09-18
clients/trin-bridge: remove unnecessary trin bridge cli arg

njgheorghita:
Trin bridge no longer needs the cli flag specifying the client type.


njgheorghita:
ping @KolbyML

PR #1168, opened 2024-09-25
Add --rpc-api flag to Fluffy startup script

web3-developer:
Fluffy's default RPC APIs are now eth and portal but the Discv5 API is used on the portal hive tests so this is needed to enable it.


kdeme (approved):

kdeme:
cc @KolbyML

PR #1170, opened 2024-10-02
simulators/portal: add test for if GetContent checks locally

KolbyML:
A PR was made on our specs which was caused by `portal_*GetContent` not being fully tested on Hive causing different results between clients.

Fluffy didn't check the local database
Trin did

`portal_*GetContent` should always check locally fix as it is the cheapest

fixes https://github.com/ethereum/portal-network-specs/pull/344


kdeme (commented):
I'm fine with adding the test, but the specification should be updated first, see also:

https://github.com/ethereum/portal-network-specs/pull/344#issuecomment-2390798692

PR #1175, opened 2024-10-17
simulators/portal: update recursive_find_content to get_content

KolbyML:
@kdeme @bhartnett ping for review

I tested locally  and everything appears to be working for trin/fluffy with the updated tests


bhartnett:
> @kdeme @bhartnett ping for review
> 
> I tested locally and everything appears to be working for trin/fluffy with the updated tests

These changes look good to me. I can also confirm that we have already updated Fluffy to use the new `portal_*GetContent` name.

PR #1179, opened 2024-11-15
hive: add --sim.buildarg flag

marioevz:
Adds `--sim.buildarg` as a parameter to add build arguments to the simulator Dockerfile.

### DIFF
--- a/hive.go
+++ b/hive.go
@@ -13,8 +13,26 @@ import (
 
 	"github.com/ethereum/hive/internal/libdocker"
 	"github.com/ethereum/hive/internal/libhive"
+	docker "github.com/fsouza/go-dockerclient"
 	"gopkg.in/inconshreveable/log15.v2"
 )
+
+type buildArgs []docker.BuildArg
+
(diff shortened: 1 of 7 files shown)


marioevz:
@spencer-tb @danceratopz

KolbyML:
> This change will allow us to specify the "test fixture" release version used in the ethereum/execution-spec-test simulators on the command-line, just as we already specify the client version under test. Just wanted to share: perhaps it's useful for the Portal simulators?

I would find this ability useful, since I want/plan to be able to sub in testnet test data for our `portal` simulators.

spencer-tb (approved):
LGTM! Just tried this out and it works well.

I created a PR on top of this to update the consume dockerfiles: #1180

fjl:
I pushed a small update before merging. The update is about using `map[string]string` instead of `[]docker.BuildArg` as the internal representation of build args. I did this because the packages `internal/libhive`, `internal/fakes` should not depend on `go-dockerclient` directly. They should only use this package via an interface.

PR #1180, opened 2024-11-27
simulators/eth/eest: use `--sim.buildarg` for consume

spencer-tb:
## Description

Updates the consume simulator dockerfiles to use input build args. This will allow us to pass different release fixtures dynamically to both `consume rlp` and `consume engine`.

The default remains as `latest-stable-release`, as these are the fixtures that validate client mainnet releases.

## Requirements

The PR requires that #1179 is merged first, as it utilizes the `--sim.buildarg` flag that the latter adds.


marioevz (approved):
LGTM!

Issue #1185, opened 2024-12-14
`engine-withdrawals` suite fails with nil pointer dereference on reth


Rjected:
We are running into issues with the `engine-withdrawals` suite:
```
CLMocker: Adding engine client d65f8fd0
>> (94bc3602) {"jsonrpc":"2.0","id":1,"method":"eth_getBlockByNumber","params":["latest",false]}
<< (94bc3602) {"jsonrpc":"2.0","id":1,"result":{"hash":"0xf62b6ae287c6cdc2f1bda80d0ce83cff49aa8665d0477605f9c157355885f511","parentHash":"0x0000000000000000000000000000000000000000000000000000000000000000","sha3Uncles":"0x1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347","miner":"0x0000000000000000000000000000000000000000","stateRoot":"0xb13210209bd72e7902c597ee31321eddddbfc739eeffbd902092ac50a0326fe2","transactionsRoot":"0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421","receiptsRoot":"0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421","logsBloom":"0x00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000","difficulty":"0x0","number":"0x0","gasLimit":"0x2fefd8","gasUsed":"0x0","timestamp":"0x1234","extraData":"0x","mixHash":"0x0000000000000000000000000000000000000000000000000000000000000000","nonce":"0x0000000000000000","baseFeePerGas":"0x3b9aca00","withdrawalsRoot":"0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421","size":"0x221","uncles":[],"transactions":[],"withdrawals":[]}}
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x530e78]

goroutine 7324 [running]:
[...]

PR #1186, opened 2024-12-16
simulators/ethereum/engine: remove terminal total difficulty

fjl:
Since the merge has long passed, and clients are starting to remove support for the merge transition, we can update the engine API tests to start at genesis.


fgimenez:
this should close https://github.com/ethereum/hive/issues/1185 right?

fjl:
Yes!

fjl:
There are a couple broken tests though, I don't understand all the tests enough to know how to fix them.

One type of test that still uses a chain with a PoW prefix is the sync test. I would vote for removing the sync tests in ethereum/engine simulator, and leave them for the ethereum/sync simulator. It uses a cancun-based chain at the moment.
We could extend ethereum/sync to try more modes.

marioevz (approved):
Thanks so much, this really makes it much cleaner now that TTD is unnecessary.

PR #1196, opened 2025-01-02
simulators/portal: update gossip to put content tests

KolbyML:
updates tests to match spec https://github.com/ethereum/portal-network-specs/pull/346

@bhartnett for a quick look

### DIFF
--- a/simulators/portal/src/suites/state/interop.rs
+++ b/simulators/portal/src/suites/state/interop.rs
@@ -9,7 +9,7 @@ use ethportal_api::jsonrpsee::http_client::HttpClient;
 use ethportal_api::types::execution::header_with_proof::{
     BlockHeaderProof, HeaderWithProof, SszNone,
 };
-use ethportal_api::types::portal::{FindContentInfo, GetContentInfo};
+use ethportal_api::types::portal::{FindContentInfo, GetContentInfo, PutContentInfo};
 use ethportal_api::types::portal_wire::MAX_PORTAL_CONTENT_PAYLOAD_SIZE;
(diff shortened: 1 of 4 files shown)


simulators/portal/src/suites/state/interop.rs
```diff
@@ -461,10 +461,10 @@ dyn_async! {
         for TestData { header, key: content_key, offer_value: content_offer_value, .. } in test_data.clone() {
             store_header(header, &client_b.rpc).await;
 
-            match client_a.rpc.gossip(content_key.clone(), content_offer_value.encode()).await {
-                Ok(nodes_gossiped_to) => {
-                   if nodes_gossiped_to != 1 {
-                        panic!("We expected to gossip to 1 node instead we gossiped to: {nodes_gossiped_to}");
+            match client_a.rpc.put_content(content_key.clone(), content_offer_value.encode()).await {
+                Ok(PutContentInfo { peer_count, .. }) => {
```
bhartnett, line 465: Maybe the test should check that the storedLocally flag is true is the response.

KolbyML: This assumes the radius is 100 percent, which I don't think we can assume

bhartnett (approved):

simulators/portal/src/suites/state/interop.rs
bhartnett: Yes true. I guess it depends how the nodes are setup in the tests.

KolbyML: Maybe in the future we can standardize some startup params around this kind of thing

PR #1198, opened 2025-01-03
all: Add hive instance information endpoint

marioevz:
Adds a `/hive` endpoint to the server API that returns an object containing information about the hive instance that is currently running.

#### Hive info object contents
- `command`: Command line used to start the hive server.
- `clientFile`: Contents of the file specified in `--client-file` if any.
- `commit`: Git commit used build the hive binary.
- `date`: Date of build of the hive binary.

### DIFF
--- a/hive.go
+++ b/hive.go
@@ -168,20 +168,24 @@ func main() {
 			clientList = libhive.FilterClients(clientList, filter)
 		}
 	}
+	hiveInfo := libhive.HiveInfo{
+		Command:    os.Args,
+		ClientFile: clientList,
+	}
 
 	// Build clients and simulators.
 	if err := runner.Build(ctx, clientList, simList, simBuildArgs); err != nil {
 		fatal(err)
 	}
 	if *simDevMode {
-		runner.RunDevMode(ctx, env, *simDevModeAPIEndpoint)
+		runner.RunDevMode(ctx, env, *simDevModeAPIEndpoint, hiveInfo)
 		return
 	}
 
 	// Run simulators.
 	var failCount int
 	for _, sim := range simList {
-		result, err := runner.Run(ctx, sim, env)
+		result, err := runner.Run(ctx, sim, env, hiveInfo)
 		if err != nil {
 			fatal(err)
 		}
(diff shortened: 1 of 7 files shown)


fjl:
I don't fully understand why this is a part of the simulation API. From discussion with @parithosh, I learned that this info is wanted for the purpose of displaying it in test results.

The simulation API is the endpoint for the simulator. The simulator doesn't really need to know the hive version. If hive wants to write information about itself into the output file, it can just do it.

marioevz:
> I learned that this info is wanted for the purpose of displaying it in test results.

Yes that's the only reason.

> The simulation API is the endpoint for the simulator. The simulator doesn't really need to know the hive version. If hive wants to write information about itself into the output file, it can just do it.

But isn't the simulator created by hive? We don't have any other way to communicate with hive, other than doing some hacky stuff like looking at which process is using the port and then somehow get the information about the output file.

Perhaps we could look it like a way of looking for the information about the hive version that was used to create the simulator.

fjl:
I guess I still don't understand what you want to achieve. I think what you want to do is to display the hive version in the suite results page of the viewer, so that viewing a result makes it clear which version of the hive repo was used to run it.

marioevz:
@fjl this is an example of what I mean: https://hive.pectra-devnet-5.ethpandaops.io/suite.html?suiteid=1737336360-cdc9f3b55f2e828c17ab8b6c364934e8.json&suitename=eest%2Fconsume-engine#

There is no easy way to know what `nethermind_pectra-devnet-5` means.

The `simulator.log` does not tell you because it doesn't know, it will just use whatever hive gives it, but it has no way of asking about the clients' metadata.

If you look at the test case description of any of the failing tests, which is generated by the simulator, and ignoring the incorrect html format for now, it gives you a command that should be run to reproduce the test case, but the parameter it uses is `--client-file configs/prague.yaml`, but it (the simulator) has no way of telling you exactly what the contents of `configs/prague.yaml` are.

fjl:
Ahh, I understand now! Makes sense.

PR #1199, opened 2025-01-06
simulators/ethereum/engine: TTD related fixes

marioevz:
Contains the following fixes:
- Fixed an incorrect handling of the `Process` returned value in the geth-node module.
- The genesis configure method now does actually set the TTD because the geth-node module fails to start if the value is nil.
- Remove `HIVE_CLIQUE_PERIOD`, `HIVE_CLIQUE_PRIVATEKEY`, and `HIVE_MINER` are removed.
- Remove Total Difficulty check in the devp2p library because it resulted in nil pointer exception.
- Fix TTD in the auth test suite.


fjl:
deployed on https://hivetests2.ethdevops.io

PR #1207, opened 2025-01-14
Add blob schedule to hive genesis files

parithosh:


cmd/hivechain/genesis.go
```diff
@@ -100,6 +102,16 @@ func (cfg *generatorConfig) createChainConfig() *params.ChainConfig {
 			chaincfg.CancunTime = &timestamp
 		case "prague":
 			chaincfg.PragueTime = &timestamp
+			if cfg.blobSchedule != nil {
```
fjl, line 105: Can you remove this for now? We will add it back when go-ethereum has the change to add BlobSchedule on master.

parithosh: I removed the change!

marioevz (approved):
LGTM

PR #1210, opened 2025-01-16
client/shisui: change docker image repo

fearlessfe:
client shisui has moved to [zen-eth](https://github.com/zen-eth/shisui) organization, so the docker image should be update


KolbyML (approved):
:shipit: looks good

PR #1225, opened 2025-01-22
clients/ultralight: move .git to root directory

KolbyML:
In ultralight's new ping extension branch they require the .git folder be accessible in the base directly in this case it is the root direclty.

@acolytec3 for review


acolytec3 (approved):
LGTM

PR #1227, opened 2025-01-23
cmd/hiveview: new test summary and general UI improvements

skylenet:
Some new features:

General:
- Allow passing `--limit` to the `hiveview --listings` command. So that we can decide how many we want to index. Still kept 200 as default.
- Added client versions to the `hiveview --listings` output, so that we can show them on the index page.

Index:
- Summary on top, that has a box for each simulator/client pair that was run. Sortable by name,coverage or time.
- Clicking on the box brings the user to the table view, where all the past executed tests can be seen (filtered by sim/client pairs)
- Added a diff column, that shows a counter of + or - passed
 tests, compared to the previous test run (from the same simulator + client pairs) 
- Generally adding a bit more colors and smaller design adjustments
- Show the specific client version in the table
- Added some keyboard shortcuts

Suite page:
- Allow rendering html within the test description. Used a lot by the ethereum/eest simulators, where they show how to reproduce a specific test.
- Show total passed/failed tests stats on the top right and add colors to it (green -> passed, red -> failed)

Viewer:
- Add log syntax highlighting using prism

localhost_8001_ (2)


marioevz (approved):
Looks absolutely amazing, thanks for this!

cc @fjl on final check for this since he is the owner of all of these files :)

PR #1228, opened 2025-01-27
clients/go-ethereum: fix default TTD

s1na:
I believe this is the correct default value because EEST tests will not provide a value for TOTALTERMINALDIFFICULTY when testing HFs prior to the merge and in that case we shouldn't assume to be post-PoW.


fjl:
Let's try!

PR #1229, opened 2025-01-27
clients/go-ethereum: add cancun blob schedule

s1na:
As per https://github.com/ethereum/go-ethereum/pull/31002 every configured HF needs to have a blob schedule entry. I don't see any reason to make the cancun schedule configurable but happy to change if people disagree.


lightclient (approved):
LGTM

PR #1230, opened 2025-01-28
clients/reth: combine remaining blocks before import 

joshieDo:
```
eest/consume-rlp : tests/prague/eip2935_historical_block_hashes_from_state/test_block_hashes.py::test_block_hashes_history[fork_Prague-blockchain_test-full_history_plus_one_check_blockhash_first]-reth
```

The above test would call the import command block by block. For a total of 8192 blocks. This would result in a timeout, since staged sync (used on `reth import`) is not suited for single blocks.

Combines all individual blocks into one file, before importing them into reth

### DIFF
--- a/clients/reth/reth.sh
+++ b/clients/reth/reth.sh
@@ -101,11 +101,10 @@ fi
 # Load the remainder of the test chain
 echo "Loading remaining individual blocks..."
 if [ -d /blocks ]; then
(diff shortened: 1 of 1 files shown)


mattsse (approved):

rkrasiuk (approved):

marioevz (approved):
LGTM, thanks!

PR #1233, opened 2025-01-29
clients/nimbus-el: add .git approach Dockerfile

skylenet:
similar problem as in https://github.com/ethereum/hive/pull/1232


marioevz (approved):
LGTM, thanks

PR #1237, opened 2025-02-01
clients/reth: combine all blocks except the last before import 

joshieDo:
Previous PR https://github.com/ethereum/hive/pull/1230 fixed some, but broke other tests setup. This should cover both cases.

> First import as many blocks as possible, and only then import the last one.
> This is important because usually tests expecting a failure will assert that the last valid inserted block is at `last - 1`. If we attempted to import all of them the pipeline would unwind the whole range.

run: https://github.com/paradigmxyz/reth/actions/runs/13084266651

### DIFF
--- a/clients/reth/reth.sh
+++ b/clients/reth/reth.sh
@@ -103,7 +103,7 @@ echo "Loading remaining individual blocks..."
 mapfile -t BLOCKS < <(ls /blocks/*.rlp 2>/dev/null | sort -n)
 
 if [[ ! -d "/blocks" || ${#BLOCKS[@]} -eq 0 ]]; then
-    echo "Error: No blocks found."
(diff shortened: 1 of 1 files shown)


marioevz (approved):
LGTM, thanks for following this up.

fjl:
I'm not so sure about this whole thing. The idea with individual block files was being able to provide two or more forked chains to the client for import. The EEST should just provide a concatenated chain for all clients.

jochem-brouwer:
Isn't this a bug in Reth? If one tries to import multiple blocks (from a combined RLP file it seems), but there's one "wrong" block, then it rejects all blocks (including the valid ones?).

This approach will also not work if there are two invalid blocks, then the block import where first all blocks except the last one imported will fail.

PR #1243, opened 2025-02-10
simulators/portal: fix race condition in `beacon-sync` suite

acolytec3:
**The issue**:
The `portal/beacon-sync` test suite currently pulls test data from a live beacon node in order to test syncing the Portal beacon light clients. 

The bridge client that retrieves the finalized and optimistic updates continues to pull new updates from the beacon client as long as the test is running.

This can be an issue in the `optimistic update` test where a new optimistic update is produced every slot.  

Only one optimistic update is pushed into the network and the simulator then retrieves a new update from the bridge to compare to when validating the test results.  

**The solution**

Add a parameter to the `start` method in the `BridgeService` that tells the bridge whether to continue to retrive new updates from the Beacon Node provider or not.  In the the `optimistic update` test, setting this parameter to `false` will ensure that we don't accidentally pull a new optimistic update when validating the test result since this is current test only tests for a single update.

### DIFF
--- a/simulators/portal/src/suites/beacon/bridge/service.rs
+++ b/simulators/portal/src/suites/beacon/bridge/service.rs
@@ -126,7 +126,6 @@ impl BridgeService {
                     break;
                 }
             }
-            println!("Bridge service stopped");
         });
     }
 }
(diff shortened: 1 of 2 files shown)


acolytec3:
@njgheorghita see above

KolbyML:
It looks good, just need CI fixed

acolytec3:
Lint should be cleaned up.  Thanks!

KolbyML (approved):
:shipit: looks good

PR #1248, opened 2025-02-25
simulators/ethereum: add eest/execute simulator

spencer-tb:
## Description

Adds execute from [EEST](https://github.com/ethereum/execution-spec-tests) as a hive simulators, utilizing its hive mode extension.

Specifically runs transactions tests.

### DIFF
--- a/simulators/ethereum/eest/execute/Dockerfile
+++ b/simulators/ethereum/eest/execute/Dockerfile
@@ -4,10 +4,12 @@ FROM ghcr.io/astral-sh/uv:python3.10-bookworm-slim
 ## Default fork/marker/git-ref
 ARG fork=Prague
 ENV FORK=${fork}
-ARG marker_expr=
+ARG marker_expr=transaction_test
 ENV MARKER_EXPR=${marker_expr}
 ARG branch=main
 ENV GIT_REF=${branch} 
+ARG keyword=
+ENV KEYWORD=${keyword}
 
 ## Clone and install EEST
 RUN apt-get update && apt-get install -y git
@@ -23,4 +25,5 @@ WORKDIR /execution-spec-tests
 RUN uv sync
 
 ## Define `execute hive` entry point using the local fixtures
-ENTRYPOINT uv run execute hive --fork "$FORK" -m "$MARKER_EXPR"
+ENTRYPOINT uv run execute hive --fork "$FORK" -m "$MARKER_EXPR" -k "$KEYWORD" -v
+
--- a/simulators/ethereum/eest/execute/Dockerfile
+++ b/simulators/ethereum/eest/execute/Dockerfile
@@ -1,9 +1,11 @@
 # Builds and runs the EEST (execution-spec-tests) execute hive simulator
 FROM ghcr.io/astral-sh/uv:python3.10-bookworm-slim
 
-## Default fixtures/git-ref
-ARG fixtures=stable@latest
-ENV FIXTURES=${fixtures}
+## Default fork/marker/git-ref
+ARG fork=Prague
+ENV FORK=${fork}
+ARG marker=
+ENV MARKER=${marker}
 ARG branch=main
 ENV GIT_REF=${branch} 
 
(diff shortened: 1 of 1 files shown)


draghtnight (approved):

simulators/ethereum/eest/execute/Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+RUN uv sync
+
+# Cache the fixtures. This is done to avoid re-downloading the fixtures every time
+# the container starts.
+# If newer version of the fixtures is needed, the image needs to be rebuilt.
+# Use `--docker.nocache` flag to force rebuild.
+RUN uv run consume cache --input "$FIXTURES"
+
+## Define `execute hive` entry point using the local fixtures
+ENTRYPOINT uv run execute hive --fork=Prague -m transaction_test
```
marioevz, line 30: `-m transaction_test` could be an argument.

I don't remember if blank spaces were correctly handled by docker though (between `-m` and `transaction_test`).

simulators/ethereum/eest/execute/Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+    git checkout FETCH_HEAD;
+
+WORKDIR /execution-spec-tests
+RUN uv sync
+
+# Cache the fixtures. This is done to avoid re-downloading the fixtures every time
+# the container starts.
+# If newer version of the fixtures is needed, the image needs to be rebuilt.
+# Use `--docker.nocache` flag to force rebuild.
+RUN uv run consume cache --input "$FIXTURES"
```
marioevz, line 27: We don't need to cache anything here since we don't use the fixtures, but this raises an interesting question regarding how to pin `execute` to a release.

simulators/ethereum/eest/execute/Dockerfile
```diff
@@ -0,0 +1,30 @@
+# Builds and runs the EEST (execution-spec-tests) execute hive simulator
+FROM ghcr.io/astral-sh/uv:python3.10-bookworm-slim
+
+## Default fixtures/git-ref
+ARG fixtures=stable@latest
+ENV FIXTURES=${fixtures}
```
marioevz, line 6: Fixtures are not used by `execute`.

marioevz (commented):
Thanks for this! Some comments.

simulators/ethereum/eest/execute/Dockerfile
spencer-tb: Added `--fork` and `-m` as an argument.
Fork defaults to Prague for now. `-m` works by default is the arg is set empty.

simulators/ethereum/eest/execute/Dockerfile
```diff
@@ -0,0 +1,26 @@
+# Builds and runs the EEST (execution-spec-tests) execute hive simulator
+FROM ghcr.io/astral-sh/uv:python3.10-bookworm-slim
+
+## Default fork/marker/git-ref
+ARG fork=Prague
+ENV FORK=${fork}
+ARG marker=
+ENV MARKER=${marker}
```
danceratopz, line 8: We're kinda leaking pytest-speak into the hive world (what's a marker?), which I'd prefer to avoid, but I don't think it's possible, so...

In general this could be a marker expression, this is a big ugly, but would better convey this possibility:
```suggestion
ARG marker_expr=
ENV MARKER_EXPR=${marker_expr}
```

Would be awesome to provide help from the command-line on these possible args, don't know how yet tho.

Btw, from the pytest help:
```
  -m MARKEXPR           Only run tests matching given mark expression. For
                        example: -m 'mark1 and not mark2'.
```

danceratopz (approved):
LGTM, just one small suggestion, but perhaps it's too messy.

marioevz (approved):
LGTM, thanks for this!

simulators/ethereum/eest/execute/Dockerfile
```diff
@@ -0,0 +1,26 @@
+# Builds and runs the EEST (execution-spec-tests) execute hive simulator
+FROM ghcr.io/astral-sh/uv:python3.10-bookworm-slim
+
+## Default fork/marker/git-ref
+ARG fork=Prague
+ENV FORK=${fork}
+ARG marker_expr=
+ENV MARKER_EXPR=${marker_expr}
```
marioevz, line 8: If the default is empty, but we do pass the `-m` parameter unconditionally, does it break on default?

spencer-tb: Nice! Let me change the default to be transaction test

spencer-tb:
CIFO #1291 and future execute improvements: https://github.com/ethereum/execution-spec-tests/issues/1664

PR #1259, opened 2025-03-26
simulators/ethereum/eest: update consume dockerfiles

spencer-tb:
## 🗒️ Description
Updates the consume rlp and engine dockerfiles.

### Before

- Used the `python:3.10-slim` as the base docker image installing additional packages including uv. 

### After 
- Now uses the `ghcr.io/astral-sh/uv:python3.10-bookworm-slim` as the base docker image. No longer requires wget, tar, and uv installs.
- Renames the `INPUT` env var to `FIXTURES`.
- Combines some `RUN` commands to make the Dockerfiles more compact overall.

### DIFF
--- a/simulators/ethereum/eest/consume-engine/Dockerfile
+++ b/simulators/ethereum/eest/consume-engine/Dockerfile
@@ -1,28 +1,30 @@
 # Builds and runs the EEST (execution-spec-tests) consume engine simulator
-FROM python:3.10-slim
+FROM ghcr.io/astral-sh/uv:python3.10-bookworm-slim
 
-## Default fixtures and branch
+## Default fixtures/git-ref
 ARG fixtures=stable@latest
-ENV INPUT=${fixtures}
+ENV FIXTURES=${fixtures}
 ARG branch=main
-ENV BRANCH=${branch}
-
-## Install dependencies
-RUN apt-get update && \
-    apt-get install -y git wget tar && \
-    apt-get clean && \
-    rm -rf /var/lib/apt/lists/*
+ENV GIT_REF=${branch} 
 
 ## Clone and install EEST
-RUN git clone https://github.com/ethereum/execution-spec-tests.git --branch "$BRANCH" --single-branch --depth 1
-WORKDIR execution-spec-tests
-RUN pip install uv && uv sync
+RUN apt-get update && apt-get install -y git
+
+# Allow the user to specify a branch or commit to checkout
+RUN git init execution-spec-tests && \
+    cd execution-spec-tests && \
+    git remote add origin https://github.com/ethereum/execution-spec-tests.git && \
+    git fetch --depth 1 origin $GIT_REF && \
+    git checkout FETCH_HEAD;
+
+WORKDIR /execution-spec-tests
+RUN uv sync
 
 # Cache the fixtures. This is done to avoid re-downloading the fixtures every time
 # the container starts.
 # If newer version of the fixtures is needed, the image needs to be rebuilt.
 # Use `--docker.nocache` flag to force rebuild.
-RUN uv run consume cache --input "$INPUT"
+RUN uv run consume cache --input "$FIXTURES"
 
 ## Define `consume engine` entry point using the local fixtures
-ENTRYPOINT uv run consume engine -v --input "$INPUT"
+ENTRYPOINT uv run consume engine -v --input "$FIXTURES"
--- a/simulators/ethereum/eest/consume-rlp/Dockerfile
+++ b/simulators/ethereum/eest/consume-rlp/Dockerfile
@@ -1,28 +1,30 @@
 # Builds and runs the EEST (execution-spec-tests) consume rlp simulator
-FROM python:3.10-slim
+FROM ghcr.io/astral-sh/uv:python3.10-bookworm-slim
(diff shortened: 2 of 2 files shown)


simulators/ethereum/eest/consume-engine/Dockerfile
```diff
@@ -1,28 +1,26 @@
...
-RUN apt-get update && \
-    apt-get install -y git wget tar && \
-    apt-get clean && \
-    rm -rf /var/lib/apt/lists/*
-
 ## Clone and install EEST
-RUN git clone https://github.com/ethereum/execution-spec-tests.git --branch "$BRANCH" --single-branch --depth 1
-WORKDIR execution-spec-tests
-RUN pip install uv && uv sync
+RUN apt-get update && apt-get install -y git && \
```
danceratopz, line 11: I personally prefer to not `&&` these commands as it makes it hard to understand which command was responsible in case of fail.

simulators/ethereum/eest/consume-engine/Dockerfile
```diff
@@ -1,28 +1,26 @@
...
+    git clone https://github.com/ethereum/execution-spec-tests.git --branch "$BRANCH" --single-branch --depth 1 && \
```
danceratopz, line 12: Can you check if there is a non-breaking way to allow any ref here? I think `--branch` only allows branch names and tags but not commit hashes? Could be I'm wrong, plz check. Would be really handy for debugging and testing, see: 
- https://github.com/ethereum/execution-spec-tests/issues/1354

danceratopz (commented):
Thanks! Didn't test yet, but added one comment and one request! :pray:

simulators/ethereum/eest/consume-engine/Dockerfile
spencer-tb: Should be closer to what it was before!

simulators/ethereum/eest/consume-engine/Dockerfile
```diff
@@ -1,28 +1,34 @@
...
-RUN pip install uv && uv sync
+RUN apt-get update && apt-get install -y git
+
+# Use commit if specified, otherwise use branch env variable
+RUN if [ -z "$COMMIT" ]; then \
+        git clone https://github.com/ethereum/execution-spec-tests.git --branch "$BRANCH" --single-branch --depth 1; \
+    else \
+        git clone https://github.com/ethereum/execution-spec-tests.git && \
+        cd execution-spec-tests && \
+        git checkout "$COMMIT"; \
```
danceratopz, line 21: Currently, cloning execution-spec-tests is quite slow, cf https://github.com/ethereum/execution-spec-tests/issues/1369
Even after #1369 is solved, this trick should speed-up grabbing a specific commit:
```suggestion
        git init execution-spec-tests && \
        cd execution-spec-tests && \
        git remote add origin https://github.com/ethereum/execution-spec-tests.git && \
        git fetch --depth 1 origin $COMMIT && \
        git checkout FETCH_HEAD; \
```

With this change, the following takes about 30 seconds, without about 450.0s seconds - we really need to solve #1369!
```
docker build --no-cache --build-arg commit=b4e09b1e2750d36d205ccac345a499e868fc3286 .
```

simulators/ethereum/eest/consume-rlp/Dockerfile
```diff
@@ -1,28 +1,34 @@
...
-RUN pip install uv && uv sync
+RUN apt-get update && apt-get install -y git
+
+# Use commit if specified, otherwise use branch
+RUN if [ -z "$COMMIT" ]; then \
+        git clone https://github.com/ethereum/execution-spec-tests.git --branch "$BRANCH" --single-branch --depth 1; \
+    else \
+        git clone https://github.com/ethereum/execution-spec-tests.git && \
+        cd execution-spec-tests && \
+        git checkout "$COMMIT"; \
```
danceratopz, line 21:
```suggestion        
        git init execution-spec-tests && \
        cd execution-spec-tests && \
        git remote add origin https://github.com/ethereum/execution-spec-tests.git && \
        git fetch --depth 1 origin $COMMIT && \
        git checkout FETCH_HEAD; \
```

danceratopz (commented):
Thanks for a lot for adding this, just suggested an optimization below!

simulators/ethereum/eest/consume-engine/Dockerfile
danceratopz, line 21: If it is possible to use this approach with $BRANCH, then we don't need an if or a separate command-line variable, e.g., we could something like this:
```suggestion
RUN git init execution-spec-tests && \
        cd execution-spec-tests && \
        git remote add origin https://github.com/ethereum/execution-spec-tests.git && \
        git fetch --depth 1 origin $BRANCH && \
        git checkout FETCH_HEAD; \
```

spencer-tb:
Thanks @danceratopz! Added your suggestions.

We can now clone by both branch/commit using `--sim.buildarg branch=main/e4793e6e...`

The build arg `branch` is kept to not break anything for other clients downstream.

danceratopz (approved):
Thanks so much for adding this, really fast builds now: 17 s locally for me!

PR #1263, opened 2025-03-27
clients/nimbus-el: add base docker image

advaita-saha:
Now uses a docker image from the registry. The git build approach is used in Dockerfile.git


marioevz (approved):
LGTM, verified to be working locally for me. Thanks!

PR #1269, opened 2025-04-10
simulators/portal: bump ethportal-api for offer json-rpc changes

KolbyML:
The json-rpc https://github.com/ethereum/portal-network-specs/pull/370 was modified

### DIFF
--- a/simulators/portal/Cargo.toml
+++ b/simulators/portal/Cargo.toml
@@ -10,7 +10,7 @@ alloy-rlp = "0.3.11"
 alloy-primitives = "0.8.24"
 anyhow = "1.0"
 chrono = "0.4"
-ethportal-api = { git = "https://github.com/ethereum/trin", rev = "4d605acb22874778f2cebd3b06fea69d1325199d" }
+ethportal-api = { git = "https://github.com/ethereum/trin", rev = "f584d6e745b6900e1d7e63fd4814cafeb43c91a9" }
 futures = "0.3.25"
 hivesim = { git = "https://github.com/ethereum/hive", rev = "4408fc1de3fee3ac23acd25a812c117756af2f39" }
 itertools = "0.14"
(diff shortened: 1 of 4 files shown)


simulators/portal/Cargo.toml
```diff
@@ -10,7 +10,7 @@ alloy-rlp = "0.3.11"
 alloy-primitives = "0.8.24"
 anyhow = "1.0"
 chrono = "0.4"
-ethportal-api = { git = "https://github.com/ethereum/trin", rev = "4d605acb22874778f2cebd3b06fea69d1325199d" }
+ethportal-api = { git = "https://github.com/ethereum/trin", rev = "f584d6e745b6900e1d7e63fd4814cafeb43c91a9" }
```
morph-dev, line 13: What do you think about not using git as source and use published version instead?

It would probably not work right now (as we didn't publish new version). But if Mike does it today (he should), it would work.

Up to you

simulators/portal/src/suites/history/interop.rs
```diff
@@ -223,7 +224,10 @@ dyn_async! {
             Err(err) => panic!("Error getting node info: {err:?}"),
         };
 
-        let _ = client_a.rpc.offer(target_enr, vec![(target_key.clone(), target_value.encode())]).await;
+        let accept_info = client_a.rpc.offer(target_enr, vec![(target_key.clone(), target_value.encode())]).await.expect("Failed to send offer");
```
morph-dev, line 227: nit: this line also looks a bit too long...

morph-dev (approved):
Few suggestions. All of them are up to you

:ship:

simulators/portal/Cargo.toml
KolbyML: I will leave it as is, just because we will probably ending up switching it back to git commit in the future.

PR #1272, opened 2025-04-14
clients/reth: improve version parsing in dockerfiles

danceratopz:
Hiveview reports reth's version without a commit hash: `reth Version: 1.3.8`. 

This PR adds the build's commit hash: `1.3.8+52c3e3cc`.

Try it out with the full output of `reth --version`:
```
echo "reth Version: 1.3.8
  Commit SHA: df6d5dd1dd23da180ecd2e15585b3bebc3b5c656
  Build Timestamp: 2025-04-14T08:43:20.271808115Z
  Build Features: jemalloc
  Build Profile: release" | head -2 | awk 'NR==1 {gsub("reth Version: ", ""); version=$0} NR==2 {gsub("Commit SHA: ", ""); sha=substr($0, 1, 8)} END {print version"+"sha}'
```

It's more brittle, but I don't think unexpected output would prevent the image getting built. 

These checks should be outsourced to a helper library that we can independently unit test.

### DIFF
--- a/clients/reth/Dockerfile
+++ b/clients/reth/Dockerfile
@@ -20,7 +20,7 @@ ADD enode.sh /hive-bin/enode.sh
 RUN chmod +x /hive-bin/enode.sh
 
 # Create version.txt
-RUN /usr/local/bin/reth --version | sed -e 's/reth \(.*\)/\1/' > /version.txt
+RUN /usr/local/bin/reth --version | head -2 | awk 'NR==1 {gsub("reth Version: ", ""); version=$0} NR==2 {gsub("Commit SHA: ", ""); sha=substr($0, 1, 8)} END {print version"+"sha}' > /version.txt
 
 # Export the usual networking ports to allow outside access to the node.
 EXPOSE 8545 8546 30303 30303/udp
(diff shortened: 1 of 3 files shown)


fjl:
> These checks should be outsourced to a helper library that we can independently unit test.

I kind of disagree on the helper library. This kind of stuff is why we have hive in the first place, to paper over differences in the clients. Where else would we use this library then?

danceratopz:
> > These checks should be outsourced to a helper library that we can independently unit test.
> 
> I kind of disagree on the helper library. This kind of stuff is why we have hive in the first place, to paper over differences in the clients. Where else would we use this library then?

Sorry for the lack of explanation. I didn't want to take the logic out of hive, just out of the dockerfiles. I would suggest adding a `clients/scripts/` directory with this logic. Then we could:
```
ADD scripts/ /scripts/
```
This would allow functions added to this library to be "unit" tested (and reduce code duplication across 3 dockerfiles).

spencer-tb:
Slight nudge on this! Or review from @marioevz - would be awesome to get this in for https://hive.ethpandaops.io/ :)

PR #1273, opened 2025-04-15
Update ethereumjs Dockerfile

scorbajio:
This change update the ethereumjs dockerfile used in hive to pull the base image from the ethereumjs image from the official registry.

### DIFF
--- a/clients/ethereumjs/ethereumjs.sh
+++ b/clients/ethereumjs/ethereumjs.sh
@@ -46,7 +46,7 @@
 # Immediately abort the script on any error encountered
 set -e
 
-CLIENT_DIRECTORY=/ethereumjs-monorepo/packages/client
+CLIENT_DIRECTORY=/usr/app/packages/client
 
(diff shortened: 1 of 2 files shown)


clients/ethereumjs/ethereumjs.sh
```diff
@@ -46,7 +46,7 @@
 # Immediately abort the script on any error encountered
 set -e
 
-CLIENT_DIRECTORY=/ethereumjs-monorepo/packages/client
+CLIENT_DIRECTORY=/usr/app/packages/client
```
marioevz, line 49: If the path in this file was modified, it feels to me like the other two Docker files also need to be updated (Dockerfile.git, Dockerfile.local)

scorbajio: I've made an update to the ethereumjs dockerfile to use the same client directory path name as before. Once [#3987](https://github.com/ethereumjs/ethereumjs-monorepo/pull/3987) is merged, this should be addressed.

marioevz (approved):
LGTM, thanks!

PR #1278, opened 2025-04-24
feat: enable beacon subnetwork for trin client and in portal simulators

morph-dev:
With recent  change in trin (https://github.com/ethereum/trin/pull/1775), we need to enable Beacon subnetwork in all tests and add it as a default flag for trin client.

I also updated ethportal-api library version.


KolbyML (approved):
I kind of wish our code took the networks the user wanted enabled and just did the rest of the work itself, better UX.

PR #1288, opened 2025-05-15
simulators/portal: fix off by 1 error in flair calculation

KolbyML:
I added test data for testing if we handle fork boundaries properly and from this I found we have an off by one error when calculating the flair for the test title


Dsorken (approved):
Nice catch

PR #1291, opened 2025-05-23
simulators/ethereum/eest: New simulator `execute-blobs`

marioevz:
Adds new simulator `execute-blobs` which uses the EEST `execute` command to run blob tests:
- Send blob transactions with cell proofs to the client.
- Use `engine_getBlobsV2` to get proofs back from the client
- Verify proofs

At the moment contains only three tests and relies on branch [`devnets/osaka/0`](https://github.com/ethereum/execution-spec-tests/tree/devnets/osaka/0), but more tests will be added soon.

### DIFF
--- a/simulators/ethereum/eest/execute-blobs/Dockerfile
+++ b/simulators/ethereum/eest/execute-blobs/Dockerfile
@@ -1,0 +1,24 @@
+# Builds and runs the EEST (execution-spec-tests) consume engine simulator
+FROM ghcr.io/astral-sh/uv:python3.10-bookworm-slim
+
+## Default git-ref/fork
+ARG branch=main
+ENV GIT_REF=${branch} 
+ARG fork=Osaka
+ENV FORK=${fork}
+
+## Clone and install EEST
+RUN apt-get update && apt-get install -y git
+
+# Allow the user to specify a branch or commit to checkout
+RUN git init execution-spec-tests && \
+    cd execution-spec-tests && \
+    git remote add origin https://github.com/ethereum/execution-spec-tests.git && \
+    git fetch --depth 1 origin $GIT_REF && \
+    git checkout FETCH_HEAD;
+
+WORKDIR /execution-spec-tests
+RUN uv sync
+
+## Define `consume engine` entry point using the local fixtures
+ENTRYPOINT uv run execute hive --fork=$FORK -v -m get_blobs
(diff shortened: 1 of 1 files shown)


marioevz:
Great question, my reasoning was:
1) These tests might be unstable so it could be better to run them in isolation
2) Clients might want to see these results on their own to assert fusaka readiness more easily

But yes, I think eventually it should just be run with the rest of the tests, but is it the suggestion to skip this PR and rather add it as a config modification PR to the hive view page directly?

simulators/ethereum/eest/execute-blobs/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+    cd execution-spec-tests && \
+    git remote add origin https://github.com/ethereum/execution-spec-tests.git && \
+    git fetch --depth 1 origin $GIT_REF && \
+    git checkout FETCH_HEAD;
+
+WORKDIR /execution-spec-tests
+RUN uv sync
+
+## Define `consume engine` entry point using the local fixtures
+ENTRYPOINT uv run execute hive --fork=$FORK -v -m get_blobs
```
danceratopz, line 24:
```suggestion
ENTRYPOINT uv run execute hive --fork=$FORK -v -m get_blobs
```
About the `-m get_blobs`...

Here's just one other small bit of relevant info... and that's that we only parametrize a simulator with the format if it's supported by the particular simulator (`_supported_fixture_formats` is defined in the simulator's local `conftest.py`):

```python
def pytest_generate_tests(metafunc):
    """
    Generate test cases for every test fixture in all the JSON fixture files
    within the specified fixtures directory, or read from stdin if the directory is 'stdin'.
    """
    if "cache" in sys.argv:
        return

    test_cases = metafunc.config.test_cases
    param_list = []
    for test_case in test_cases:
        if test_case.format.format_name not in metafunc.config._supported_fixture_formats:
            continue
```
https://github.com/ethereum/execution-spec-tests/blob/020f728eefedc75a68e6fba6dbc65322e1d9f490/src/pytest_plugins/consume/consume.py#L411-L423

This avoids having to add these command-line args, which is beneficial if we want to allow the user to specify other markers (as in #1248).

spencer-tb (approved):
Approving as discussed on call! TLDR; merge now and add to hive CI. Improve over the coming weeks following @danceratopz suggestions.

Future improvements summarized below from my understanding:
1) Remove the marker usage `-m` from the dockerfile.
2) Replace this directly within EEST where we specify the test types we want to execute hive mode to run.
3) Run execute hive mode using pytests session level so we can easily filter tests within the hive ui like we do for consume!

marioevz:
Thanks for the issue @spencer-tb !

I'll merge this and ping the EthPandaOps team to add the simulator :)

PR #1298, opened 2025-05-30
feat: allow enabling system wide docker auth

skylenet:
The current approach only allowed for credential helpers when using docker auth. Credential helpers are not the only way to deal with auth. There's also plain auth + credential stores. 

This PR allows using the underlying libs `docker.NewAuthConfigurationsFromDockerCfg()` which will support any of the usual auth configs that exist in the system wide config files. More info at https://pkg.go.dev/github.com/fsouza/go-dockerclient@v1.11.2#NewAuthConfigurationsFromDockerCfg .

I've added the  `--docker.auth` flag and marked the `--docker.cred-helper` flag as deprecated . By enabling this flag you'll now  he able to use whatever the system wide config is. 

This PR refactors the Docker authentication mechanism to use the system-wide Docker configuration instead of the previous credential helper approach.

The new approach automatically handles multiple Docker config file locations in the standard order:
```
$DOCKER_CONFIG/plaintext-passwords.json
$DOCKER_CONFIG/config.json
$HOME/.docker/plaintext-passwords.json
$HOME/.docker/config.json
$HOME/.dockercfg
```


marioevz (approved):
LGTM.

PR #1302, opened 2025-06-09
feat: add optional GOPROXY build arg and use it as an env var

skylenet:
Sometimes we have some CI runs that get blocked by the default proxy. This changed makes it possible to define our own proxy when fetching go packages.


marioevz (approved):
LGTM

PR #1303, opened 2025-06-10
Update devp2p chain to pos 

mpaulucci:
This PR has two changes:

- Instead of using the Geth chain, let's decouple them and generate our own inside Hive repo.
- Updates the chain to pos. This will allow only-pos clients (see [ethrex](https://github.com/lambdaclass/ethrex)) to be able to not have to mantain a hive fork. Also, starting on eth/69, we don't even serve pre merge blocks. The chain generation script has been added to `simulators/devp2p/mkchain.sh`

The the Geth one for reference: https://github.com/ethereum/go-ethereum/blob/master/cmd/devp2p/internal/ethtest/mkchain.sh


fjl:
I'm updating it upstream in go-ethereum instead. We want to use the same chain for snap tests, so it makes sense to update it for both.

mpaulucci:
Thanks!

PR #1315, opened 2025-07-01
clients/nethermind: Change runtime image

flcl42:
A precompile update requires an additional native dependency available in -noble flavor of dotnet runtime image


spencer-tb (approved):
LGTM!

PR #1321, opened 2025-07-20
clients/nimbus-el: activate admin rpc

jangko:
https://github.com/status-im/nimbus-eth1/pull/3431 add a new namespace for rpc. And it is not activated by default. This PR fix it.
The admin rpc required when calling `admin_nodeInfo`. Previous version always enable it without tied to any namespace configuration.


marioevz (approved):
LGTM, thanks!

PR #1327, opened 2025-08-06
refactor(nethermind): Refactor enode parsing to not use logs

fselmo:
Using logs is too fragile. It's not ideal to get the enode from the logs in the first place - and, as in my case, if the log level is set lower it's never able to parse the `enode` from the logs as it never shows up. All other clients use the `admin_nodeInfo` request for this purpose and we should do so for Nethermind as well.

This also turns on `admin` module as this was necessary for this change.


marioevz (approved):
LGTM

PR #1329, opened 2025-08-13
clients: add bpo1 to client mappers

spencer-tb:
## Description

Integrates the BPO1 ruleset addition added in EEST: https://github.com/ethereum/execution-spec-tests/pull/2035/

Required for testing BPO forks with consume rlp/engine. BPO2/3/4/5 can be added in the future.

### DIFF
--- a/clients/go-ethereum/mapper.jq
+++ b/clients/go-ethereum/mapper.jq
@@ -79,9 +79,21 @@ def to_bool:
         "target": (if env.HIVE_BPO1_BLOB_TARGET then env.HIVE_BPO1_BLOB_TARGET|to_int else 9 end),
         "max": (if env.HIVE_BPO1_BLOB_MAX then env.HIVE_BPO1_BLOB_MAX|to_int else 14 end),
         "baseFeeUpdateFraction": (if env.HIVE_BPO1_BLOB_BASE_FEE_UPDATE_FRACTION then env.HIVE_BPO1_BLOB_BASE_FEE_UPDATE_FRACTION|to_int else 8832827 end)
+      },
+      "bpo2": {
+        "target": (if env.HIVE_BPO2_BLOB_TARGET then env.HIVE_BPO2_BLOB_TARGET|to_int else 9 end),
+        "max": (if env.HIVE_BPO2_BLOB_MAX then env.HIVE_BPO2_BLOB_MAX|to_int else 14 end),
+        "baseFeeUpdateFraction": (if env.HIVE_BPO2_BLOB_BASE_FEE_UPDATE_FRACTION then env.HIVE_BPO2_BLOB_BASE_FEE_UPDATE_FRACTION|to_int else 8832827 end)
+      },
+      "bpo3": {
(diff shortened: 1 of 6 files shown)


clients/go-ethereum/mapper.jq
```diff
@@ -76,6 +76,7 @@ def to_bool:
         "baseFeeUpdateFraction": (if env.HIVE_OSAKA_BLOB_BASE_FEE_UPDATE_FRACTION then env.HIVE_OSAKA_BLOB_BASE_FEE_UPDATE_FRACTION|to_int else 5007716 end)
```
marioevz, line 76: We should add `"bpo1"` here, very similar to how we have `"osaka"` and `"prague"`, etc, for this and the rest of the clients.

marioevz (commented):
Awesome :D just one comment I think we should resolve.

clients/go-ethereum/mapper.jq
marioevz: Interesting, I'm not sure if we should set up a default, could you check what happens if we don't put a default?
Ideally, we should only have the `"bpo1"` entry if we pass any of the `HIVE_BPO1_BLOB_...` constants IMO.

If it adds too much complexity I guess it would not hurt to add a default.

clients/nimbus-el/mapper.jq
```diff
@@ -94,8 +94,14 @@ def to_bool:
         "target": (if env.HIVE_OSAKA_BLOB_TARGET then env.HIVE_OSAKA_BLOB_TARGET|to_int else 6 end),
         "max": (if env.HIVE_OSAKA_BLOB_MAX then env.HIVE_OSAKA_BLOB_MAX|to_int else 9 end),
         "baseFeeUpdateFraction": (if env.HIVE_OSAKA_BLOB_BASE_FEE_UPDATE_FRACTION then env.HIVE_OSAKA_BLOB_BASE_FEE_UPDATE_FRACTION|to_int else 5007716 end)
+      },
+      "bpo1": 
+        "target": (if env.HIVE_BPO1_BLOB_TARGET then env.HIVE_BPO1_BLOB_TARGET|to_int else 9 end),
+        "max": (if env.HIVE_BPO1_BLOB_MAX then env.HIVE_BPO1_BLOB_MAX|to_hex else 14 end),
+        "baseFeeUpdateFraction": (if env.HIVE_BPO1_BLOB_BASE_FEE_UPDATE_FRACTION then env.HIVE_BPO1_BLOB_BASE_FEE_UPDATE_FRACTION|to_hex else 8832827 end)
```
marioevz, line 101:
```suggestion
      "bpo1": {
        "target": (if env.HIVE_BPO1_BLOB_TARGET then env.HIVE_BPO1_BLOB_TARGET|to_int else 9 end),
        "max": (if env.HIVE_BPO1_BLOB_MAX then env.HIVE_BPO1_BLOB_MAX|to_int else 14 end),
        "baseFeeUpdateFraction": (if env.HIVE_BPO1_BLOB_BASE_FEE_UPDATE_FRACTION then env.HIVE_BPO1_BLOB_BASE_FEE_UPDATE_FRACTION|to_int else 8832827 end)
```

marioevz (approved):

PR #1345, opened 2025-09-24
simulators/ethereum/rpc-compat: special handling for speconly tests

kclowes:
This PR looks for the `speconly` flag that comes in in the comments from execution-api tests and only checks the structure of the JSON, not the values. If we decide to pull it in, I'll mark at least the estimateGas tests and maybe others(?) as speconly.

Based on the convo [here](https://discord.com/channels/595666850260713488/804019759934078987/1420084584586022972). 

cc @fjl and @lightclient


fjl:
Cool idea! I'll check tomorrow.

PR #1411, opened 2026-03-27
internal/libdocker: fix empty container IP on Docker Desktop for macOS

qu0b:
## Summary

- On newer macOS Docker Desktop versions, `NetworkSettings.IPAddress` from Docker's inspect API returns empty for containers on the default bridge network. This causes hiveproxy to report `addr=:8081` (no IP), resulting in `HIVE_SIMULATOR=http://:8081` which the simulator cannot reach (`connection refused`).
- Falls back to iterating `NetworkSettings.Networks` endpoint map when the top-level IP is empty — the canonical Docker API source for per-network IPs.
- Fails fast with a clear error if no IP is found from either source, instead of silently propagating an empty IP.

## Test plan

- [ ] Run hive on macOS Docker Desktop and verify the simulator connects successfully
- [ ] Run hive on Linux Docker to confirm no behavioral change (fallback is gated on empty IP)

### DIFF
--- a/internal/libdocker/container.go
+++ b/internal/libdocker/container.go
@@ -174,6 +174,29 @@ func (b *ContainerBackend) StartContainer(ctx context.Context, containerID strin
 	}
 	info.IP = container.NetworkSettings.IPAddress
 	info.MAC = container.NetworkSettings.MacAddress
+	// On some Docker setups (e.g. Docker Desktop for macOS), the top-level
+	// NetworkSettings.IPAddress field may be empty even though the container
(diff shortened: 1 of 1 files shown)


qu0b:
@fjl @spencer-tb I'm currently not driving a mac device do you have someone that has already tested this maybe?

fjl:
I cannot reproduce the original issue. It just works for me on Docker Desktop 4.37.2

fjl:
I could reproduce it on Docker Desktop 4.72.0

PR #1415, opened 2026-04-08
internal/libdocker: handle build of hiveproxy docker image if it already exists

siddarthkay:
Else we see race conditions like this when running parallel simulations
```

[2026-04-08T06:54:05.058Z] Apr  8 06:54:04.610 ERR image build failed
image=hive/hiveproxy err="failed to create an image
docker.io/hive/hiveproxy:latest with target
sha256:d7c16d655368a6901ca972aa6ef30e2cfd77458cdf54b9d861e89a354980011d
after deleting the existing one: AlreadyExists: image
\"docker.io/hive/hiveproxy:latest\": already exists"

[2026-04-08T06:54:05.058Z] failed to create an image
docker.io/hive/hiveproxy:latest with target
sha256:d7c16d655368a6901ca972aa6ef30e2cfd77458cdf54b9d861e89a354980011d
after deleting the existing one: AlreadyExists: image
"docker.io/hive/hiveproxy:latest": already exists
```

### DIFF
--- a/internal/libdocker/proxy.go
+++ b/internal/libdocker/proxy.go
@@ -16,7 +16,11 @@ const hiveproxyTag = "hive/hiveproxy"
 
 // Build builds the hiveproxy image.
 func (cb *ContainerBackend) Build(ctx context.Context, b libhive.Builder) error {
-	return b.BuildImage(ctx, hiveproxyTag, hiveproxy.Source)
+	err := b.BuildImage(ctx, hiveproxyTag, hiveproxy.Source)
+	if err != nil && !imageAlreadyExists(err) {
+		return err
+	}
+	return nil
 }
 
 // ServeAPI starts the API server.
--- a/internal/libdocker/builder.go
+++ b/internal/libdocker/builder.go
@@ -86,6 +86,13 @@ func (b *Builder) BuildImage(ctx context.Context, name string, fsys fs.FS) error
 	return nil
 }
 
+// imageAlreadyExists reports whether a build failed only because a concurrent
+// hive process already produced the identical image, a benign cross-process
+// race when multiple simulations share one Docker daemon.
+func imageAlreadyExists(err error) bool {
+	return strings.Contains(err.Error(), "AlreadyExists")
+}
+
 func (b *Builder) buildConfig(ctx context.Context, name string) docker.BuildImageOptions {
 	nocache := false
 	if b.config.NoCachePattern != nil {
--- a/internal/libdocker/builder.go
+++ b/internal/libdocker/builder.go
@@ -80,7 +80,11 @@ func (b *Builder) BuildImage(ctx context.Context, name string, fsys fs.FS) error
 
 	b.logger.Info("building image", "image", name, "nocache", opts.NoCache, "pull", b.config.PullEnabled)
 	if err := b.client.BuildImage(opts); err != nil {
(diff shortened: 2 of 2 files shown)


siddarthkay:
Hi @KolbyML @danceratopz : can I please request a review from you for the changes in this PR?
They're pretty minimalistic and self contained.

internal/libdocker/proxy.go
```diff
@@ -16,7 +17,14 @@ const hiveproxyTag = "hive/hiveproxy"
...
 func (cb *ContainerBackend) Build(ctx context.Context, b libhive.Builder) error {
-	return b.BuildImage(ctx, hiveproxyTag, hiveproxy.Source)
+	if err := b.BuildImage(ctx, hiveproxyTag, hiveproxy.Source); err != nil {
+		// If it already Exists its okay, no need to fail
+		if strings.Contains(err.Error(), "AlreadyExists") {
+			return nil
+		}
+		return err
+	}
+	return nil
```
danceratopz: How about adding a helper in `builder.go` (same package), which gives the magic string a documented home? I'd put it in in `builder.go`, near `BuildImage`. 

```go
// in builder.go, near BuildImage (strings is already imported there)

// imageAlreadyExists reports whether a build failed only because a concurrent
// hive process already produced the identical image, a benign cross-process
// race when multiple simulations share one Docker daemon.
func imageAlreadyExists(err error) bool {
	return strings.Contains(err.Error(), "AlreadyExists")
}
```

Then we can simplify the `Build` function body to:

```suggestion
	err := b.BuildImage(ctx, hiveproxyTag, hiveproxy.Source)
	if err != nil && !imageAlreadyExists(err) {
		return err
	}
	return nil
```

internal/libdocker/proxy.go
```diff
@@ -6,6 +6,7 @@ import (
 	"log/slog"
 	"net"
 	"net/http"
+	"strings"
```
danceratopz: If the helper moves to `builder.go` as suggested above, this `strings` import is no longer needed.
```suggestion
```

danceratopz (commented):
Thanks for adding this! My suggestion below intends to make the code more inline with that of this repo.

 `Builder.BuildImage` logs `b.logger.Error("image build failed", "image", name, "err", err)` before returning the error, so even after this PR swallows it, every parallel run still prints that exact `ERR image build failed ... AlreadyExists` followed by a successful run. Perhaps we should downgrade the log there to avoid a confusing log entry here in L83:
https://github.com/ethereum/hive/blob/e021983cd0d4d46823b29951f807acd82f6477b2/internal/libdocker/builder.go#L75-L83

Summoning @fjl to double-check he approves the change, respectively the suggestions below 🙏

siddarthkay:
Hi @danceratopz : Thanks for the review, I've incorporated the feedback as per your comments.
Do let me know if you need any more changes to this PR.

Cheers

fjl:
I managed to reproduce this, and the fix does work. But it's only a surface-level fix. There are also these kinds of errors:

```
Jul 17 16:28:00.287 ERR image build failed image=hive/clients/go-ethereum:latest err="creating image docker.io/hive/clients/go-ethereum:latest failed because it already exists, but accessing it also failed: No such image: hive/clients/go-ethereum:latest"
```

`already exists, but accessing it also failed`

I think this will be harder to fix.

PR #1422, opened 2026-04-11
clients, ‎simulators: Add Lean RPC compatibility tests

Dsorken:
Add standard lean RPC compatibility tests. Tests that require post genesis are ran against the leanspec client. Configured to work for devnet3 and devnet4 through a cli option


KolbyML (approved):
:shipit: looks good

PR #1426, opened 2026-04-17
clients, simulators: Leanspec helper stability and cleanup

Dsorken:
Added a fix to occasional non-deterministic test fails caused by the leanspec client instability. Additionally Cleaned up client dockerfiles and devnet enum.


KolbyML (approved):
:shipit: looks good

PR #1429, opened 2026-04-17
clients/gean: add Gean lean consensus client

menawar:
Add `gean` as a new lean consensus client, with support for the `devnet3` and `devnet4` base images from `ghcr.io/geanlabs/gean`


KolbyML (approved):
:shipit: looks good

PR #1432, opened 2026-04-18
clients/lantern: Add lantern devnet-4 support

uink45:
- Packages both `piertwo/lantern:v0.0.3` and `:v0.0.4` in the lantern client image and selects the binary at startup based on `HIVE_LEAN_DEVNET_LABEL`.
- Adds a `prepare_lean_client_assets.py` tweak so lantern v0.0.4's YAML parser accepts the `annotated_validators.yaml` the simulator generates.


KolbyML (approved):
:shipit: looks good

PR #1433, opened 2026-04-18
simulators: post genesis test timeout

Dsorken:
Currently there isn't a dedicated timeout for when a given client fails to start during a post-genesis test, leading to an indefinite stall. There is now a 3 minute timeout and an error in the test log to indicate the client that timed out so that multi-client runs can continue.


KolbyML (approved):
:shipit: looks good

PR #1434, opened 2026-04-19
clients/zeam, simulators/lean: remove zeam devnet3 runs, since it causes the simulator to wait 3 minutes

ch4r10t33r:
## Why

The zeam image bundles two binaries -- `zeam-devnet3` (built from source at a pinned revision with the `epoll-v2` patch) and `zeam-devnet4` (copied from `blockblaz/zeam:devnet4` on Docker Hub). `zeam.sh` picks one at runtime from `HIVE_LEAN_DEVNET_LABEL`.

The `devnet3-builder` stage runs the full toolchain bootstrap (apt + rustup + `git clone blockblaz/zeam` at the pinned SHA + `zig build`) and takes **~20 minutes** on every image build. The `nametag` in `simulators/lean/clients/devnet{3,4}.yaml` is only a runtime switch, so hive has no existing way to tell the Dockerfile "this run only touches devnet4, skip the source build". Today we pay the ~20 min cost on every CI run even when exercising only devnet4.

## What

Split `devnet3-builder` into two variants selected by a new build arg:

| stage                      | selected when                         | behaviour                                                                                     |
|----------------------------|---------------------------------------|-----------------------------------------------------------------------------------------------|
| `devnet3-builder-false`    | `zeam_skip_devnet3=false` *(default)* | the real source build, unchanged                                                              |
| `devnet3-builder-true`     | `zeam_skip_devnet3=true`              | stub stage: writes a placeholder binary so the final-stage `COPY` resolves                   |

The selector `FROM devnet3-builder-${zeam_skip_devnet3} AS devnet3-builder` then feeds the existing unchanged `COPY --from=devnet3-builder /src/zig-out/bin/zeam /usr/local/bin/zeam-devnet3`. `zeam_skip_devnet3` is declared as a global ARG (before the first `FROM`) so it's available to the selector.

The stub binary isn't silent: if anyone ever launches a skip-built image with `HIVE_LEAN_DEVNET_LABEL=devnet3`, `zeam.sh` will `exec` the stub, which prints a clear explanation and exits with code 64 (`EX_USAGE`). The same message already surfaces during image build in the `(/usr/local/bin/zeam-devnet3 --version || echo …)` version-capture step, which is a nice early smoke test.

`simulators/lean/clients/devnet4.yaml` opts in via `build_args` so the common case (`./hive --client-file simulators/lean/clients/devnet4.yaml`) gets the fast path out of the box. `devnet3.yaml` is untouched (still gets the full source build).

## Compatibility

- **Default behaviour is unchanged** -- `zeam_skip_devnet3` defaults to `false`, so any existing invocation that doesn't set the arg still produces both binaries exactly like master. Nothing pre-this-PR breaks.
- **devnet3 profile**: `simulators/lean/clients/devnet3.yaml` does not set the arg, so it keeps the full source build and real binary.
- **Unknown build args** are only logged as warnings by `validateClients`, so older hive binaries reading the updated `devnet4.yaml` keep working (they'll just warn and pass the arg through).

## Measured impact

Local build of `clients/zeam/Dockerfile`:

| build arg                       | time     |
|---------------------------------|----------|
| *(default, master today)*       | ~22 min  |
| `zeam_skip_devnet3=true`        | **~25 s** |

The fast path is dominated by pulling `blockblaz/zeam:devnet4` and `ubuntu:24.04` layers; the stub stage itself takes <1s.

## Test plan

- [x] `docker build --build-arg zeam_skip_devnet3=true ...` succeeds end-to-end (image tagged `zeam-hive-skip`), version-capture step prints the stub's failure message as expected.
- [x] `docker buildx build --check` passes on both code paths (no warnings from the new ARG/selector).
- [ ] `./hive --sim lean --client zeam --client-file simulators/lean/clients/devnet4.yaml` run (covered by downstream zeam CI on merge).
- [ ] `./hive --sim lean --client zeam --client-file simulators/lean/clients/devnet3.yaml` still exercises the real devnet3 binary (unchanged default path).


KolbyML (approved):
We are removing devnet3 when devnet5 development starts, so we might as well just remove it

PR #1437, opened 2026-04-20
clients/qlean, ‎simulators/lean: Connect Qlean and get tests working

Kayden-ML:
Connects Qlean and gets tests working


KolbyML (approved):
:shipit: looks good

PR #1441, opened 2026-04-22
clients/nlean: switch to --custom-network-config-dir

GrapeBaBa:
## Summary

nlean v0.4.0-devnet4 dropped the per-file \`--annotated-validators\` / \`--hash-sig-key-dir\` / \`--validator-config\` flags in favour of a single \`--custom-network-config-dir\` entry point (the convention ethlambda / gean already use). Point \`clients/nlean/nlean.sh\` at \`\$ASSET_ROOT\` directly so the prepared assets (annotated_validators.yaml + hash-sig-keys/) are discovered by canonical filenames.

## Test plan
- [ ] Lean simulator boots nlean and reaches post-genesis finalization

### DIFF
--- a/clients/nlean/nlean.sh
+++ b/clients/nlean/nlean.sh
@@ -55,7 +55,7 @@ else
 fi
 
 FLAGS=(
-    --annotated-validators "$ASSET_ROOT/annotated_validators.yaml"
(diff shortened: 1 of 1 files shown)


KolbyML (approved):
:shipit: looks good

PR #1448, opened 2026-04-25
clients, simulators: Checkpoint  sync test + stability

Dsorken:
Added a general checkpoint sync test that runs the tested client against a mesh of leanspec nodes to ensure that the tested client can sync and catch up to head. Additionally refactored helper methods into a specific file to cleanup the sync suite. Added stability improvements for the sync test.


Dsorken:
cc: @KolbyML

KolbyML (approved):
:shipit: looks good

PR #1450, opened 2026-04-27
clients, simulators: client startup and test fix

Dsorken:
Fixed client startup for qlean, nlean and ream. Fixed conditions on final rpc_compat test.


Dsorken:
cc @KolbyML

KolbyML (approved):
:shipit: looks good

PR #1455, opened 2026-04-28
simulators: client interop matrix test

Dsorken:
Added the client interop matrix test to test each client against every other client in majority and minority pairings.


Dsorken:
cc @KolbyML

KolbyML (approved):
:shipit: looks good

PR #1456, opened 2026-04-28
fix(discv5): resolve eth1 ENR via admin_nodeInfo

emmajam:
The eth1 discv5 suite passed the enode URL to `devp2p discv5 test`,
but enode URLs only describe the discv4/RLPx endpoint. Clients that
bind discv5 to a different UDP port (reth defaults to :9200) were
unreachable, causing all eth1 discv5 tests to time out.

Fetch the signed ENR via admin_nodeInfo, mirroring the pattern already
used for the beacon and portal roles. The ENR's udp/udp6 keys are the
canonical source of the discv5 endpoint. All EL client manifests in
hive already enable admin RPC and ship enode.sh scripts that use
admin_nodeInfo, so this is safe across clients.

### DIFF
--- a/simulators/devp2p/main.go
+++ b/simulators/devp2p/main.go
@@ -55,7 +55,7 @@ func main() {
 				},
 				AlwaysRun: true,
 				Run: func(t *hivesim.T, c *hivesim.Client) {
-					runDiscv5Test(t, c, (*hivesim.Client).EnodeURL)
+					runDiscv5Test(t, c, getEth1ENR)
 				},
 			},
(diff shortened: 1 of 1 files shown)


emmajam:
also fixing in reth here: https://github.com/paradigmxyz/reth/pull/23613

cskiraly:
I see two issues with this:
- it is still not working for me, even if using this PR
- we need to port to be parsable from the enode URL
We can keep this here to see if discv5 comms works, but this is not the long term solution, so we should not merge.

PR #1461, opened 2026-05-04
clients/ethrex: import block files individually

edg-l:
- ethrex's `import` aborts on the first invalid block, so when the consume-rlp simulator drops a directory containing both rejected and valid blocks (e.g. EEST fork-transition tests with `exception=...`), later valid blocks never get applied and the head stays at genesis.
- Loop over the files instead of passing the directory. `set +ex` above this block already keeps a per-file failure from killing the script.
- Matches the pattern used by `clients/erigon`, `clients/ethereumjs`, and `clients/nimbus-el`.

### DIFF
--- a/clients/ethrex/ethrex.sh
+++ b/clients/ethrex/ethrex.sh
@@ -53,9 +53,17 @@ else
 fi
 
 # Load the remainder of the test chain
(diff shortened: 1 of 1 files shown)


danceratopz (approved):
Thanks!

PR #1467, opened 2026-05-07
cmd/hiveview: order clients based on ranking across all sections

shariqnaiyer:
image

We would like a client to follow one column vertically across the lean latest page.


shariqnaiyer:
@KolbyML

KolbyML (approved):
:shipit: looks good

PR #1471, opened 2026-05-07
simulators/lean: fix advanced_head validator collision

### LINKED ISSUE #1470: lean simulator: reqresp/status/advanced_head has a hash-ordering race when ClientUnderTestRole::Validator collides with helper validator 0
## Summary

`reqresp/status/advanced_head` ([`simulators/lean/src/scenarios/reqresp.rs:631`](https://github.com/ethereum/hive/blob/master/simulators/lean/src/scenarios/reqresp.rs#L631)) snapshots the helper's head **once** at T+10s and then polls the client for an exact-root match for 180s.

When the test data uses `ClientUnderTestRole::Validator` together with `source_helper_validator_indices: None` (which is exactly what the `advanced_head` data block configures, [`reqresp.rs:124-149`](https://github.com/ethereum/hive/blob/master/simulators/lean/src/scenarios/reqresp.rs#L124)), both the helper and the client-under-test run validator 0. They each propose at every slot where `slot % num_validators == 0`. As a result, every such slot has **two competing sibling blocks with the same parent**.

The leanSpec fork-choice tiebreaker (`ForkChoiceStore.compute_head`, `head = max(children, key=lambda x: (weights[x], x))`) picks the **lexicographically larger root** when weights are equal. So the canonical head at the contested slot deterministically depends on which root has more leading bits — i.e. on a **random-looking byte comparison between two unrelated state roots**.

This is what makes the test's outcome a coin flip. The match between `client_fork_choice.head` and the snapshot only happens when both nodes have the *same* fork-choice answer at the snapshot moment, which happens reliably only if:

* the client's competing block has a **smaller** root than the helper's — both nodes pick the helper's, helper snapshots its own root, client reorgs to it, match; or
* the client's competing block has a **larger** root **and** the helper has already received and reorged onto it before T+10s.

The second case loses to gossip latency in practice, so any client whose contested-slot block happens to lex-sort larger than the helper's fails the test even when it is fully spec-compliant.

## Reproduction (zeam_devnet4, devnet-4 suite, run from 2026-05-07)

Suite: https://hive.leanroadmap.org/suite.html?suiteid=1778164266-ea7641a3b8c224ad3bd80e36b6b8cc77.json&suitename=reqresp&client=zeam_devnet4

zeam log around the contested slot 18 (proposer V0):

```
13:39:33.001 [s=18 i=0] [chain] processed block with root=0xf2a2d395... slot=18  // zeam's own V0 block, parent=0xbaf71c (slot 17 helper)
13:39:33.798 [s=18 i=1] [chain] constructing attestation data at slot=18 with chain head={"root":"0xf2a2d395...","slot":18}
13:39:34.823 [s=18 i=2] [chain] processed block with root=0xa05a96a3... slot=18  // helper's V0 block, parent=0xbaf71c
                                                                                  // sibling of f2a2; lex tiebreak f2 > a0 keeps head on f2a2
13:39:36.4xx                received gossip attestations slot=18 V0,V1,V2 from helper
13:39:37.000 [s=18 i=4] [forkchoice] slot_interval=4   // acceptNewAttestations promotes V1/V2 to a05; weights f2a2=1, a05=2
13:39:37.166 [s=19 i=0]    received gossip block slot=19 parent_root=0xa05a96... // helper extends a05
13:39:37.201 [s=19 i=0] [forkchoice] fork choice reorg detected: depth=1 old_head_slot=18 new_head_slot=19
```

zeam's head trajectory through the polling window: f2a2(18) → 891b(19) → 9a36(20) → eb35(21, own V0) → 93ef(24, helper) → ... It is never equal to `0xa05a96…` at a 1-second poll boundary.

## Why this is a test-design issue, not a client bug

* zeam's tiebreaker matches `ForkChoiceStore.compute_head` in leanSpec exactly (lex-larger root wins on equal weight). This is the same rule the helper itself runs.
* If the helper were given enough time to reorg onto zeam's `f2a2` (the larger root), `source_fork_choice.head` would be `f2a2` and the test would pass — both nodes are on the same head. The snapshot is just taken too eagerly.

The cross-client failure pattern in the same suite supports this: 5 of 8 clients fail `advanced_head` (`gean_devnet4`, `lantern_devnet4`, `nlean_devnet4`, `qlean_devnet4`, `zeam_devnet4`), 3 pass (`ethlambda_devnet4`, `grandine_lean_devnet4`, `ream_devnet4`). For example, ethlambda's slot-18 root was `0x6863…` < helper's `0x8917…`, so ethlambda reorgs to the helper's block and the snapshot is stable. Roughly the 50/50 split you'd expect from a hash race.

ch4r10t33r:
The `reqresp/status/advanced_head` test had a hash-ordering race that flipped pass/fail per client.

Helper runs validators 0,1,2; client runs as `ClientUnderTestRole::Validator` which defaults to V0. Both nodes propose competing blocks at every V0 slot. The leanSpec tie-break (larger root wins on equal weight) then decides the canonical chain by hash, so whether the helper's T+10s snapshot ever matches the client's head is a coin flip — about half the clients fail this test on any given run.

Fix: have the helper own only V1+V2 in this test so V0 belongs exclusively to the client. No proposer collision, no race.

Closes #1470.

### DIFF
--- a/simulators/lean/src/scenarios/reqresp.rs
+++ b/simulators/lean/src/scenarios/reqresp.rs
@@ -1,5 +1,6 @@
 use crate::utils::helper::{
     start_post_genesis_sync_context, HelperGossipForkDigestProfile, PostGenesisSyncTestData,
+    LEAN_SPEC_SOURCE_VALIDATORS_EXCLUDING_V0,
 };
 use crate::utils::libp2p_mock::{
     client_multiaddr, compute_client_peer_id, decode_request, encode_request, encode_request_raw,
@@ -137,7 +138,10 @@ dyn_async! {
                         use_checkpoint_sync: true,
                         connect_client_to_lean_spec_mesh: true,
                         client_role: ClientUnderTestRole::Validator,
-                        source_helper_validator_indices: None,
+                        // Helper owns V1+V2 so client (default V0) has exclusive proposer slots; see #1470.
+                        source_helper_validator_indices: Some(
+                            LEAN_SPEC_SOURCE_VALIDATORS_EXCLUDING_V0.to_string(),
+                        ),
                         helper_peer_count: 2,
                         helper_fork_digest_profile: if selected_lean_devnet() == LeanDevnet::Devnet4 {
                             HelperGossipForkDigestProfile::SelectedDevnet
@@ -633,12 +637,16 @@ dyn_async! {
 
         sleep(Duration::from_secs(10)).await;
 
-        let source_fork_choice = context.load_live_helper_fork_choice().await;
-
+        // Compare against the helper's *live* head re-fetched each iteration
+        // rather than a single T+10s snapshot. The helper keeps producing
+        // blocks (it's still aggregating V1+V2) so a frozen snapshot becomes
+        // unreachable as soon as the helper moves on; equality with a moving
+        // target is the right invariant for "client has caught up".
         let mut caught_up = false;
         let deadline = std::time::Instant::now() + Duration::from_secs(REQRESP_SYNC_TIMEOUT_SECS);
 
         while std::time::Instant::now() < deadline {
+            let source_fork_choice = context.load_live_helper_fork_choice().await;
             let client_fork_choice = load_fork_choice_response(&context.client_under_test).await;
 
             if client_fork_choice.head == source_fork_choice.head {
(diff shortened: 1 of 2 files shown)


KolbyML (approved):
:shipit: looks good

PR #1472, opened 2026-05-07
cmd/hiveview: display client interop self runs

Dsorken:
Added support for self run displays in the client-interop grid


Dsorken:
cc @KolbyML

KolbyML (approved):
:shipit: looks good

PR #1473, opened 2026-05-07
simulators/lean: enable client self interop

Dsorken:
Enabled client self interop runs in the client-interop suite.


Dsorken:
cc @KolbyML

KolbyML (approved):
:shipit: looks good

PR #1481, opened 2026-05-11
fix grandine_lean peer id derivation mismatch

bomanaps:
Match MockNode's compute_client_peer_id to the Python helper's deterministic key derivation by using lean_client_kind instead of split('_').next(), so compound names  like grandine_lean produce the right peer id. Inject HIVE_CLIENT_PRIVATE_KEY in client_under_test_environment so the lean client actually boots with that key, which  
fixes Outbound failure: DialFailure in the BlocksByRoot reqresp tests.


bomanaps:
cc @KolbyML

KolbyML (approved):
:shipit: looks good

PR #1487, opened 2026-05-12
clients, simulators: head delay sync tests

Dsorken:
Added sync tests for when local head falls behind network finalized and when local head falls behind but is not passed by network finalized. Included some stability improvements for the leanspec client.


Dsorken:
cc @KolbyML

KolbyML (approved):
:shipit: looks good

PR #1490, opened 2026-05-13
cmd/hiveview: gridview history selectors

Dsorken:
Added grid specific historical run selection as well as global historical run selection. The selected runs are also reflected in the client scoring grid.


Dsorken:
cc @KolbyML

KolbyML (approved):

PR #1495, opened 2026-05-15
simulators/ethereum/engine: Enforce expected blobs order by setting maxFeePerBlobGas

fab-10:
Test `Blob Transaction Ordering, Multiple Accounts (Cancun)` sends a bunch of blob txs, all with the same fees and expects a specific tx ordering by the EL clients, but since txpool is implementation dependent the assumption that the tie-break is equal in all client is weak, for example Besu does not pass the test because the internal order of the txs is different and it prefers to add the txs with the single blob first.

To make the test more predictable in putting txs with more blobs first, we can simply raise a bit their `maxFeePerBlobGas` value, as it was already done for `Blob Transaction Ordering, Multiple Clients` test, probably for the same reason.


marioevz (approved):
Makes sense, thanks!

PR #1499, opened 2026-05-19
cmd/hiveview: Subnet interop matrix support

Dsorken:
Added support to show the subnet interop matrix if it is present. Additionally added a small edge case fix to the lean latest graph


Dsorken:
cc: @KolbyML

KolbyML (approved):
:shipit: looks good

PR #1501, opened 2026-05-20
fix(discv5): add discovery port configuration for reth client

emmajam:
Hive is testing Reth’s discv5 on UDP port 30303.

Reth’s discv5 defaults to UDP port 9200.

Hive’s Reth wrapper sees `HIVE_DISCV5=1` but does not pass a Reth flag to move discv5 onto 30303, so the test sends PING/TALKREQ/FINDNODE to the wrong port and times out.


cskiraly (approved):
LGTM

PR #1502, opened 2026-05-21
simulators/lean: fetch spec fixtures from leanSpec releases

KolbyML:


Kayden-ML (approved):
Looks good nice work👍

PR #1503, opened 2026-05-21
fix(reth): add flag for startup to mark idle

emmajam:
this flag tells reth to mark startup sync as idle when no backfill is running, so the transaction manager stops treating the node as “initially syncing.” should fix the generic/eth devp2p tests


cskiraly:
@emmajam, isn't this interfering with some other tests? If not, LMK and I merge.

PR #1505, opened 2026-05-22
cmd/hiveview: aggregator specific matrix support

Dsorken:
Added support for the now split interop matrix for displaying the minority and majority aggregator results.


Dsorken:
cc: @KolbyML

KolbyML (approved):
:shipit: looks good

PR #1506, opened 2026-05-22
simulators: Aggregator interop distinct matrices

Dsorken:
Added distinct matrices for when the aggregator in the client-interop regular matrix test is either the minority or the majority client of the client pairing.


Dsorken:
cc: @KolbyML

KolbyML (approved):
:shipit: looks good

PR #1507, opened 2026-05-25
clients, cmd/hivechain: Source depositContractAddress from HIVE_DEPOSIT_CONTRACT_ADDRESS

smartprogrammer93:
The EL client genesis/chainspec mappers hardcoded the mainnet deposit contract address (`0x000…7705Fa`), so every client reported it back in `eth_config` regardless of the chain under test. That breaks the `rpc-compat` `eth_config/get-config` fixture, which expects the zero address.

### DIFF
--- a/clients/besu/mapper.jq
+++ b/clients/besu/mapper.jq
@@ -99,7 +99,7 @@ def to_int:
     "bpo3Time": env.HIVE_BPO3_TIMESTAMP|to_int,
     "bpo4Time": env.HIVE_BPO4_TIMESTAMP|to_int,
     "bpo5Time": env.HIVE_BPO5_TIMESTAMP|to_int,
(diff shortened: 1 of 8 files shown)


fjl:
Ahh wait. There is a problem with this. We generally do not use the `config` part of the input genesis JSON. So this should be specified as an environment variable `HIVE_DEPOSIT_CONTRACT_ADDRESS`.

smartprogrammer93:
@fjl fixed. also in hivechain as well.

kclowes:
@fjl can we get this one in? All clients are failing rpc-compat on this.

marioevz (approved):
LGTM, verified change to be working and proceeding to merge to unblock failing tests. Thanks.

PR #1508, opened 2026-05-25
simulators/lean: document make build prerequisite and add example commands

mananuf:
## Summary

- Adds a **Prerequisites** section to `simulators/lean/README.md` telling contributors to run `make build` from the repo root before invoking `./hive --sim lean …`. The lean simulator is part of the Rust workspace and gets compiled inside a Docker image when Hive runs it; `make build` (which wraps `cargo build --workspace --all-targets`) surfaces compile errors in seconds instead of waiting for them to fail inside a slow Docker build. I hit exactly that loop on my first run and only learned afterward that the pre-flight build step wasn't documented.
- Adds two worked **Examples** (sync against a single client, and a `client-interop` run between two clients with scoped `--docker.nocache` and tee'd logs) so new contributors can copy-paste a realistic invocation instead of inferring flag combinations from `docs/commandline.md`.
- Adds a short **Flag reference** table covering only the flags used in those two examples (`--sim`, `--client-file`, `--client`, `--sim.limit`, `--docker.output`, `--docker.nocache`, `--results-root`), with a pointer to `docs/commandline.md` for the full list.

No code changes — README only.

## Test plan

- [ ] `make build` succeeds from a clean checkout (validates the prerequisite the README now claims).
- [ ] The sync example command starts the lean simulator end-to-end and writes results under `./workspace/logs`.
- [ ] README renders correctly on github.com — fenced code blocks and the flag-reference table.
- [ ] `git diff master...HEAD` shows only `simulators/lean/README.md`.

### DIFF
--- a/simulators/lean/README.md
+++ b/simulators/lean/README.md
@@ -20,6 +20,23 @@ LeanSpec helper image generates and caches a small fixed set of validator keys
 at image-build time so the tests do not need to regenerate them on every
 container start.
 
+## Prerequisites
+
+The lean simulator is part of the Rust workspace, so its source is compiled
+inside a Docker image when you invoke `./hive --sim lean …`. Before running
+any simulation, build the workspace locally from the repo root:
+
+```
+make build
+```
+
+This wraps `cargo build --workspace --all-targets` and surfaces Rust
+compilation errors in seconds, instead of waiting for them to fail inside a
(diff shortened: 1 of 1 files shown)


mananuf:
cc: @KolbyML

KolbyML (approved):
:shipit: looks good

PR #1511, opened 2026-05-28
hivesim-rs: hive run plan

Dsorken:
Adds a generated run plan log for any suite being ran, which includes the suites intended to be ran and the tests inside the suites that are intended to be ran. The information is generated with run specific info and is compatible with sim.limit flag limitations.


Dsorken:
cc: @KolbyML

KolbyML (approved):
:shipit: looks good

PR #1513, opened 2026-05-28
cmd/hiveview: lean latest active progress bar

Dsorken:
Adds compatibility for the new hive plan log to display a progress bar of active runs, displaying percentage complete for each suite in order of suite execution. It is only displayed when a run is in progress.


KolbyML (approved):
:shipit: looks good

PR #1518, opened 2026-05-29
simulators/lean: add keys

Dsorken:
Adds missing keys causing the client-interop two-subnet tests to fail


Dsorken:
cc: @KolbyML

KolbyML (approved):
:shipit: looks good

PR #1520, opened 2026-06-02
cmd/hiveview: two subnet aggregator split matrix support

Dsorken:
Adds support for the two subnet aggregator minority and majority matrices in the client interop suite


Dsorken:
cc: @KolbyML

KolbyML (approved):
:shipit: looks good

PR #1521, opened 2026-06-02
simulators/lean: two subnet aggregator split

Dsorken:
Adds a specified matrix for minority and majority aggregator client selection in the two subnet client interop tests


Dsorken:
cc: @KolbyML

KolbyML (approved):
:shipit: looks good

PR #1523, opened 2026-06-05
clients/besu: add parentheses to mapper compatible w jq 1.7

macfarla:
resolves this error when docker cache has an older version of jq, and is also compatible with the newer 1.7.1 version

```
jq: error: syntax error, unexpected //, expecting '}' (Unix shell quoting issues?) at <top-level>, line 102:
    "depositContractAddress": env.HIVE_DEPOSIT_CONTRACT_ADDRESS // "0x0000000000000000000000000000000000000000",                                                                
jq: 1 compile error
```


spencer-tb (approved):
LGTM!

PR #1524, opened 2026-06-05
fix(ethrex): parenthesize // in mapper.jq depositContractAddress

edg-l:
#1507 added this to clients/ethrex/mapper.jq:

```jq
"depositContractAddress": env.HIVE_DEPOSIT_CONTRACT_ADDRESS // "0x0000000000000000000000000000000000000000"
```

jq doesn't allow a bare `//` as an object value (object values are parsed with a restricted grammar; binary operators have to be parenthesized), so the mapper fails to compile:

```
jq: error: syntax error, unexpected //, expecting '}' at line 110
```

That means no genesis.json gets written and the ethrex client can't start, so every test fails with `could not start client ... terminated unexpectedly`.

Quick repro:

```
echo '{}' | jq '{ "a": env.X // "y" }'    # error
echo '{}' | jq '{ "a": (env.X // "y") }'  # ok
```

Fix is just wrapping it in parens, like the other fields in the file already do. Confirmed the mapper compiles and runs after the change.


spencer-tb (approved):
LGTM!

PR #1526, opened 2026-06-05
fix deposit contract

Soubhik-10:


Soubhik-10:
sorry I didnt see the other open pr

PR #1528, opened 2026-06-06
simulators/lean: bad peer sync test

Dsorken:
Added a bad peer sync test to ensure that clients ignore an outlier peer and do not use it as a source to attempt to catch up to head


Dsorken:
cc: @KolbyML

KolbyML (approved):
Looks good

PR #1529, opened 2026-06-06
cmd/hiveview: small screen header size fix

Dsorken:
Made the progress bar switch from wrapping to side scrolling when the screen is small such as on mobile


Dsorken:
cc: @KolbyML

KolbyML (approved):
Looks good

PR #1530, opened 2026-06-08
clients: restore mainnet deposit contract address default in mappers

edg-l:
#1507 made `depositContractAddress` fall back to the **zero address** when `HIVE_DEPOSIT_CONTRACT_ADDRESS` is unset. Only `cmd/hivechain` sets that env; the eels `consume` sims build genesis from EEST fixtures and never set it, so clients watch `0x0` for deposit logs, derive no deposit requests, and reject every deposit / multi-type-with-deposit block as INVALID.

Fix: restore the mainnet address (`0x00000000219ab540356cBB839Cbe05303d7705Fa`) as the fallback in the 7 mappers from #1507. The env override is preserved for hivechain custom networks; EEST fixtures always use the mainnet deposit contract.

Verified with `consume-engine` (`tests-bal@v7.2.0`, `.*eip6110_deposits.*fork_Amsterdam.*`, identical client image, only the mapper default changed):
- this PR: 74/74 pass
- master: 46 pass, 28 fail (the 28 valid-deposit cases)

### DIFF
--- a/clients/besu/mapper.jq
+++ b/clients/besu/mapper.jq
@@ -99,7 +99,7 @@ def to_int:
     "bpo3Time": env.HIVE_BPO3_TIMESTAMP|to_int,
     "bpo4Time": env.HIVE_BPO4_TIMESTAMP|to_int,
     "bpo5Time": env.HIVE_BPO5_TIMESTAMP|to_int,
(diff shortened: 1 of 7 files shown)


qu0b (approved):

spencer-tb (approved):
LGTM!

PR #755, opened 2023-04-13
simulators/ethereum/pyspec: Comment Dockerfile to run local fixtures

marioevz:
Adds a simple comment for instructions on how to run locally generated fixtures.

@spencer-tb

PR #1531, opened 2026-06-08
rpc-compat: validate speconly responses against the OpenRPC schema

MysticRyuujin:
## Problem

`speconly` tests in `rpc-compat` are meant to check a response "for schema validity" (per the fixture comment), but the implementation compared the response to the recorded `<<` example with a **bidirectional structural diff** (`checkJSONStructure`) — it rejects both *missing* and *unexpected* keys.

That cannot accommodate a method whose response shape is client- or config-specific. The motivating case is `eth_capabilities` (execution-apis #755): each resource's optional `oldestBlock`/`deleteStrategy` fields appear or not depending on gcmode, state scheme, and retention windows. No single recorded example is portable — **even between two geth nodes launched with different flags** (archive/hash omits `deleteStrategy`; full/path includes it). A fixture generated from one config fails structural matching against another, e.g.:

```
.result.state.deleteStrategy: unexpected key in response
.result.tx.deleteStrategy: unexpected key in response
```

## Fix

Validate `speconly` responses against the method's **OpenRPC result schema** — the same `santhosh-tekuri/jsonschema` validation [execution-apis' speccheck](https://github.com/ethereum/execution-apis/tree/main/tools/cmd/speccheck) performs — instead of structural-matching the recorded example. Any spec-valid response passes regardless of which optional fields a given client/config includes; genuinely invalid responses (unknown keys, wrong types, missing required fields) are still rejected because the schemas use `additionalProperties: false` and `required`.

- `schema.go` — `parseSpec` (loads the result schema per method from `openrpc.json`) and `validateResult`.
- `main.go` — `speconly` tests now validate against the schema; methods absent from the spec fall back to the previous structural matching. Ordinary (non-`speconly`) tests are unchanged and still compare exactly.
- `Dockerfile` — ships `openrpc.json` into the simulator image alongside the fixtures (with a commented local-override `ADD`, mirroring the existing `tests` convention).
- `schema_capabilities_test.go` + `testdata/openrpc-capabilities.json` — regression test: archive-mode, full-mode, and a disabled-resource response all validate; an unknown-key response is rejected.

## Verification

Ran `rpc-compat` against go-ethereum built from source: `eth_capabilities/get-capabilities` passes, and all existing `speconly` methods (`txpool_status`, `txpool_content`, `txpool_contentFrom`, `eth_estimateGas`, `eth_feeHistory`, `eth_createAccessList`) still pass — `tests=16 failed=0`.

## Companion change

Pairs with **ethereum/execution-apis#819**, which adds a generator for the `eth_capabilities` fixture (`SpecOnly`) and removes the impossible hand-written fixtures. This hive change is the prerequisite that makes that fixture pass `rpc-compat`.


simulators/ethereum/rpc-compat/Dockerfile
```diff
@@ -30,5 +34,6 @@ ADD . /source
 WORKDIR /source
 COPY --from=builder /source/rpc-compat .
 COPY --from=builder /execution-apis/tests ./tests
+COPY --from=builder /execution-apis/openrpc.json ./openrpc.json
```
s1na, line 52: Should probably test this but afaict the openrpc.json file needs to be built. It's not on git

simulators/ethereum/rpc-compat/main.go
```diff
@@ -25,6 +25,12 @@ var (
 		"genesis.json": "./tests/genesis.json",
 		"chain.rlp":    "./tests/chain.rlp",
 	}
+
+	// specMethods holds the OpenRPC result schema for each method, used to
+	// validate "speconly" tests against the spec. It is nil if the spec could
```
s1na: > If the spec could not be loaded, in which case speconly tests fall back to structural

Hm or we just quit and raise an error if the schema couldn't be loaded. Then we can also remove the structural matching. The schema validation should be strictly superior to it.

simulators/ethereum/rpc-compat/schema.go
```diff
@@ -0,0 +1,77 @@
...
+	// Set $schema explicitly to force jsonschema to use draft 2019-09, the
+	// draft the execution-apis schemas are written against.
+	draft := openrpc.Schema("https://json-schema.org/draft/2019-09/schema")
+	schema.Schema = &draft
+
+	b, err := json.Marshal(schema)
+	if err != nil {
+		return fmt.Errorf("unable to marshal schema to json: %w", err)
+	}
+	s, err := jsonschema.CompileString("result", string(b))
```
s1na: Ideally we do the compilation once and re-use that for every test.

simulators/ethereum/rpc-compat/Dockerfile
MysticRyuujin: Good catch — confirmed: `openrpc.json` is gitignored in execution-apis (generated by `make build` via specgen), so it was absent from the clone and this COPY would have failed the build. The Dockerfile now builds it from the cloned spec source with specgen before the COPY (specgen pulls zero go-ethereum packages, so it's cheap). Verified end-to-end: a full production-path run (cloning execution-apis at `refs/pull/819/head`, building the spec in-image, no local override) is green — `tests=16 failed=0`.

simulators/ethereum/rpc-compat/main.go
MysticRyuujin: Done — we now `panic` if the spec can't be loaded, and `checkJSONStructure` (the structural matching) is removed along with its test. speconly responses are always validated against the schema; a speconly method missing from the spec now fails loudly. Confirmed all current speconly methods (incl. `testing_buildBlockV1`) have result schemas in `openrpc.json`.

simulators/ethereum/rpc-compat/schema.go
MysticRyuujin: Done — each method's result schema is now compiled once in `parseSpec` (stored as `*jsonschema.Schema`) and reused for every test; `validateResult` just runs `schema.Validate`.

simulators/ethereum/rpc-compat/schema.go
```diff
@@ -0,0 +1,84 @@
...
+
+	openrpc "github.com/open-rpc/meta-schema"
+	"github.com/santhosh-tekuri/jsonschema/v5"
+)
+
+// methodSchemas maps a JSON-RPC method name to its compiled result schema,
+// parsed from an OpenRPC document. It is used to validate the responses of
+// "speconly" tests against the spec rather than byte-comparing them to a
+// recorded example, which is necessary for methods whose response is client- or
+// config-specific (e.g. eth_capabilities retention windows).
```
s1na:
```suggestion
// parsed from an OpenRPC document.
```

simulators/ethereum/rpc-compat/schema.go
```diff
@@ -0,0 +1,84 @@
...
+		schemas[name] = schema
+	}
+	return schemas, nil
+}
+
+// compileSchema compiles an OpenRPC result schema into a reusable validator.
+// This mirrors the validation done by execution-apis' speccheck tool, so
+// "speconly" tests in hive enforce the same contract: any response that is
+// valid per the OpenRPC schema passes, regardless of which optional fields a
+// particular client or configuration includes.
```
s1na:
```suggestion
```

s1na (approved):
LGTM

PR #1533, opened 2026-06-08
simulators/graphql: pass HIVE_CANCUN_TIMESTAMP to client and fix test expectations

lupin012:
The graphql simulator was not passing HIVE_CANCUN_TIMESTAMP to the
  client. Without it, the client mapper leaves cancunTime unset in the
  genesis config, causing the client to reject post-Cancun blocks during
  chain import. The last block of the test chain (testBlockchain.blocks)
  is a Cancun block, so without this fix clients reject it and only load
  34 blocks instead of 35.

  Update test expectations to reflect the additional block now correctly
  loaded (latest block 0x22 instead of 0x21).


simulators/ethereum/graphql/testcases/12_eth_getBalance_toobig_bn.json
```diff
@@ -1,5 +1,5 @@
 {
-  "request": "{block (number: 33) {account(address: \"0x6295ee1b4f6dd65047762f924ecd367c17eabf8f\") { balance } }}",
+  "request": "{block (number: 35) {account(address: \"0x6295ee1b4f6dd65047762f924ecd367c17eabf8f\") { balance } }}",
```
danceratopz, line 2: 33 was stale prio to this PR.

danceratopz (approved):
Thanks!

PR #1536, opened 2026-06-09
internal/libdocker: disable IP masquerade on hive test networks

### LINKED ISSUE #1535: libdocker: network1 inter-network MASQUERADE breaks PingMultiIP discv5 test on Mac Docker Desktop
## Summary

The `PingMultiIP` discv5 test (and the equivalent discv4 test) always fail on Mac Docker Desktop. The test passes on a Linux host.

## Root cause

The `createTestNetwork` function in `simulators/devp2p/main.go` creates a secondary Docker network (`network1`) and connects both the simulation container and the client container to it, so the test tool can send traffic from two different source IPs.

On Mac Docker Desktop, all containers run inside a Linux VM. The VM's nftables has a `POSTROUTING` rule that Docker adds for each bridge network:

```
oifname != "br-<network1>" ip saddr <network1-subnet> MASQUERADE
```

When the test tool sends a UDP packet **from** `network1`'s IP **to** the client's bridge IP, the packet crosses Docker network boundaries. The outgoing interface is `docker0` (not `br-<network1>`), so the `oifname != "br-<network1>"` condition is true, MASQUERADE fires, and the source IP is rewritten to `172.17.0.1` (the bridge gateway).

The client echoes the observed source IP in the discv5 PONG's `ToIP` field. The test asserts `pong.ToIP == listen2.IP` and fails:

```
wrong destination IP 172.17.0.1 in PONG, want 172.18.0.2
```

On Linux, Docker containers typically handle inter-network routing without triggering MASQUERADE for container-to-container traffic (L2 bridging within the same bridge device bypasses the NAT chain), so the source IP is preserved.

## Fix

Docker's bridge driver supports a per-network option `com.docker.network.bridge.enable_ip_masquerade=false` that prevents Docker from inserting the MASQUERADE rule for that network. Setting this option when `network1` is created means the source IP is no longer rewritten, and the test passes on both Linux and Mac.

The fix is a one-line change in `internal/libdocker/container.go`'s `CreateNetwork`:

```go
network, err := b.client.CreateNetwork(docker.CreateNetworkOptions{
    Name:       name,
    Attachable: true,
    Options: map[string]interface{}{
        "com.docker.network.bridge.enable_ip_masquerade": "false",
    },
})
```

## Verification

- Tested on Mac Docker Desktop (Apple Silicon)
- All 8 discv5 tests now pass (previously `PingMultiIP` failed consistently)
- The option has no effect on Linux where the issue doesn't occur

usmansaleem:
Fixes #1535

## Problem

The `PingMultiIP` discv5 (and discv4) tests fail consistently on Mac Docker Desktop but pass on Linux.

On Mac, all containers run inside a Linux VM. Docker adds this nftables POSTROUTING rule for each bridge network it creates:

```
oifname != "br-<network>" ip saddr <subnet> MASQUERADE
```

The devp2p simulator connects to two Docker networks to obtain two source IPs (`--listen1 bridgeIP --listen2 net1IP`). When the test tool sends a UDP packet **from** `net1IP` **to** the client's bridge IP, the packet crosses network boundaries inside the VM. The outgoing interface is `docker0` (not `br-<network1>`), so the masquerade rule matches and rewrites the source to the bridge gateway (`172.17.0.1`).

The discv5/discv4 client echoes the observed source IP in the PONG's `ToIP` field. The test asserts `pong.ToIP == listen2.IP` and fails:

```
wrong destination IP 172.17.0.1 in PONG, want 172.18.0.2
```

On Linux, inter-container traffic between two bridge networks in the same host is handled at L2 within the bridge device, bypassing the NAT chain, so the source IP is preserved.

## Fix

Docker's bridge driver accepts a per-network option `com.docker.network.bridge.enable_ip_masquerade=false` that suppresses the MASQUERADE rule for that specific network. Passing this option in `CreateNetwork` means the source IP is no longer rewritten when traffic crosses to another Docker network.

## Verification

Tested on Mac Docker Desktop (Apple Silicon). All 8 discv5 tests now pass:

```
simulation devp2p finished  suites=1 tests=8 failed=0
```

Previously `PingMultiIP` (test 4) always failed. On Linux the option is a no-op; existing behaviour is unchanged.


fjl:
I cannot reproduce this issue with Docker Desktop version 4.82.0

usmansaleem:
Yes, newer docker desktop seems to have fixed the issue, closing the PR.

PR #1538, opened 2026-06-09
clients/reth: add `rpc.compute-state-root-for-eth-simulate` flag for reth

Soubhik-10:
This flag will allow reth to compute state-roots in the rpc-compat eth_Simulate tests


mattsse (approved):

MysticRyuujin (approved):

spencer-tb (approved):
🫡 🖖

PR #1549, opened 2026-06-19
clients/reth: remove `jit` feature to fix docker build

Soubhik-10:


mattsse (approved):
lgtm, that should fix it

spencer-tb (approved):
Thanks!

PR #1550, opened 2026-06-19
clients/nethermind,besu: add EIP-8282 builder request contracts

spencer-tb:
Adds EIP-8282 (builder execution requests) client config for `glamsterdam-devnet-6` spec.


danceratopz (approved):
This looks good to me! Contract addresses correct. If the Besu names turn out to be incorrect, there should be a loud failure and we can update again later.

PR #1552, opened 2026-06-19
clients, simulators: late joiner tests

Dsorken:
Added two late joiner tests to the sync suite, one uses the leanspec helper to create the mesh that the tested client will join, the other uses the tested client to create the mesh.


Dsorken:
cc: @KolbyML

KolbyML (approved):
:shipit: looks good

PR #1558, opened 2026-06-23
fix(lean): release helper p2p port before restart

MegaRedHand:
## Error description

The reqresp/checkpoint-sync simulator restarts the lean-spec local helper when it fails to reach the required finalized slot. Each restart path spawned the replacement helper (binding the same p2p_port) before the previous RunningLocalLeanSpecHelper was dropped, so the old process still held the QUIC/UDP port. The replacement then failed to bind with `OSError: [Errno 98] Address already in use`, every restart attempt re-hit the conflict, and the helper never came up — leaving every positive-path reqresp scenario to time out at 480s (only the negative, disconnect-only tests passed).

## Fix

Kill and reap the old helper child before starting its replacement so the kernel releases the port first. Applied to all four restart-in-place sites (adversarial restart, retryable-exit restart, checkpoint-sync source refresh, and the finalized-slot wait retry loop).

### DIFF
--- a/simulators/lean/src/utils/helper.rs
+++ b/simulators/lean/src/utils/helper.rs
@@ -193,6 +193,8 @@ impl RunningBadCheckpointPeer {
         }
 
         eprintln!("Restarting adversarial {LOCAL_HELPER_KIND} after retryable exit: {error}");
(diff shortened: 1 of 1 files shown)


KolbyML (approved):
:shipit: looks good

PR #1559, opened 2026-06-24
hivesim-rs: stability improvement

Dsorken:
Fixes an issue where tests in the rpc-compat suite either fail and don't count or experience panics not accounted for. Added the standard retry mechanism found in the other suites for transient crashes from tested clients.


Dsorken:
cc: @KolbyML

PR #1560, opened 2026-06-24
simulators: sync suite timeout wrappers

Dsorken:
Right now there are a few edge cases that can cause a sync test to run way longer than appropriate, changing the test function to the timeout variant adds an outer timeout wrapper that prevents this issue.


KolbyML (approved):
Looks good

PR #1562, opened 2026-06-26
simulators: remove retries

Dsorken:
removes client retries and logs client failure reason. Tests now fail if a client crashes on start or during the test.


Dsorken:
cc: @KolbyML

KolbyML (approved):
:shipit: looks good

PR #1564, opened 2026-06-28
hivesim-rs: test shared client stability fix

Dsorken:
Currently in the shared client testing if a suite experiences a panic when using the start_client_with_files method it will cause the whole suite to exit and end prematurely instead of just failing gracefully and logging the failure. The fix is just to better handle panics experienced preventing the suite using the method from also crashing, as well as some more informative logging.


Dsorken:
cc: @KolbyML

KolbyML (approved):
Looks good

PR #1565, opened 2026-07-03
simulators/lean: reqresp stability fix

Dsorken:
Currently if a client fails to start in the reqresp suite the entire suite will exit instead of just moving on to the next test, causing the suite to end early not run any tests for any client after the failure. Handling the failure properly now allows for the suite to continue.


Dsorken:
cc: @KolbyML

KolbyML (approved):
:shipit: looks good

PR #1570, opened 2026-07-08
fix(hivesim-rs): surface real start-client error instead of "missing field `id`"

MegaRedHand:
## Problem

When a client container fails to start, Rust simulators report a misleading error that hides the real cause:

```
Failed to convert start node response to json: reqwest::Error { kind: Decode,
source: Error("missing field `id`", line: 1, column: 57) }
```

This shows up in the `lean` simulator on devnet5 (e.g. `sync`, `rpc-compat`): when a slow LeanSpec helper leaves a client's checkpoint-sync source unreachable, the client exits during startup and the failure is attributed to the cryptic "missing field `id`" rather than what actually happened.

## Root cause

`Simulation::start_client_with_files` decodes **every** response body as a `StartNodeResponse` (`{id, ip}`) regardless of HTTP status:

```rust
.json::<StartNodeResponse>().await.expect("Failed to convert start node response to json")
```

But when startup fails, the `startClient` handler returns a non-2xx status with `serveError`'s body `{"error": "client did not start: ..."}` (`internal/libhive/api.go`). That body has no `id` field, so serde fails with "missing field `id`" and `.expect()` panics — throwing away the message the backend already sent.

## Fix

Parse the response against its status: on success decode the `StartNodeResponse`; otherwise extract the backend's `{"error": ...}` message (falling back to the raw body) and surface it with the status code. The panic-on-failure contract callers rely on is unchanged, so there are no signature or call-site changes.

After:

```
Failed to start client: hive rejected start-client request (HTTP 500): client did not start: <reason>
```

This is a general `hivesim-rs` improvement and benefits every Rust simulator, not just `lean`.

## Testing

- Unit tests for the extracted `parse_start_node_response`: success, backend-error surfacing (with a regression guard that "missing field" no longer leaks), and raw-body fallback.
- End-to-end A/B: drove the real `start_client_with_files` over a TCP socket against a byte-accurate mock of the backend's 500 response. Old code panicked with `missing field \`id\``; fixed code surfaced `client did not start: ...`.
- `cargo test`, `cargo fmt --check`, and `cargo clippy` are clean (no new warnings).

### DIFF
--- a/hivesim-rs/src/simulation.rs
+++ b/hivesim-rs/src/simulation.rs
@@ -1,4 +1,6 @@
-use crate::types::{ClientDefinition, StartNodeResponse, SuiteID, TestID, TestRequest, TestResult};
+use crate::types::{
+    ApiError, ClientDefinition, StartNodeResponse, SuiteID, TestID, TestRequest, TestResult,
+};
 use crate::TestMatcher;
 use std::collections::HashMap;
 use std::env;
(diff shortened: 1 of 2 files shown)


KolbyML (approved):
:shipit: looks good

PR #1573, opened 2026-07-12
refactor(lean): deduplicate validation test setup

groverInnovate:
## Summary

Extracted the duplicated mock node setup from the validation tests into a shared `setup_mock_for_validation()` helper.

The helper performs the common validation test setup:

- create and configure a `MockNode`
- generate and advertise the mock ENR
- subscribe to the block gossip topic
- prepare the client runtime environment
- start the client
- complete the initial Status handshake

This removes the repeated setup code from all three validation tests without changing their behavior.

Also, replaced the remaining `sleep(Duration::from_secs(5))` calls with `mock.process_events_for(Duration::from_secs(5)).await` so all validation tests use the same event-processing mechanism.

## Verification

- `cargo check`
- Validation suite passes on devnet4 & devnet5

### DIFF
--- a/simulators/lean/src/scenarios/validation.rs
+++ b/simulators/lean/src/scenarios/validation.rs
@@ -9,15 +9,80 @@ use crate::utils::util::{
 };
 use alloy_primitives::B256;
 use hivesim::{dyn_async, Client, Test};
+use libp2p::gossipsub::IdentTopic;
 use ssz::Encode;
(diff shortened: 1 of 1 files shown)


groverInnovate:
@KolbyML @Dsorken 
Can you please share your reviews?

KolbyML (approved):
:shipit: looks good

PR #1576, opened 2026-07-15
hiveview: exclude multi-test client contexts from listing counts

danceratopz:
`multiTestContext` test cases (#1402) own the lifecycle of clients shared across multiple tests; they are infrastructure, not test results. Counting them inflates the listing: `eels/consume-engine` and `eels/consume-enginex` run the identical 50,564 fixtures, but enginex reports `ntests: 50568` (one always-passing context case per simulator worker process).

Skip them when counting `ntests`/`passes`/`fails`. They still contribute client names, the `timeout` flag and the suite start time.


spencer-tb (approved):
LGTM!

PR #1579, opened 2026-07-20
fix(lean): require finalized-block endpoint before starting checkpoint-sync client

MegaRedHand:
The checkpoint-sync readiness gate (`wait_for_checkpoint_sync_state_ready`) exempted devnet5 from waiting on `/lean/v0/blocks/finalized`, via `checkpoint_sync_requires_finalized_block_endpoint`.

A checkpoint-syncing client fetches the finalized state and block together to form its anchor pair. With the block endpoint left unchecked, the client under test could boot, race an unready `/lean/v0/blocks/finalized`, hit its own checkpoint-sync timeout, and exit. Hive then reported a spurious "client did not start: terminated unexpectedly" startup failure that was wrongly attributed to the client (observed on ethlambda under devnet5).

The leanSpec helper serves `/lean/v0/blocks/finalized` on both devnets, so require both the state and block endpoints on all devnets and drop the devnet-specific exemption.

### DIFF
--- a/simulators/lean/src/utils/helper.rs
+++ b/simulators/lean/src/utils/helper.rs
@@ -1665,10 +1665,6 @@ fn minimum_source_checkpoint_slot(test_data: &PostGenesisSyncTestData) -> u64 {
     }
 }
 
-fn checkpoint_sync_requires_finalized_block_endpoint(devnet: LeanDevnet) -> bool {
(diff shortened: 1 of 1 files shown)


KolbyML (approved):
:shipit: looks good

PR #1583, opened 2026-07-22
fix(rpc-compat): support large fixture messages

lightclient:
In #1567 it was noted that `rpc-compact` only supports 1MB line size and that means large receipt tests fail. This increases the line limit. I think we would ideally want to keep the large receipt tests for the official test suite. 

Is there reason to not do this? @MysticRyuujin wdyt?


MysticRyuujin:
Yeah, makes sense. The only risk is that the fixture generates an output larger than github's max file size but I don't think that to be the case.

PR #1584, opened 2026-07-22
all: stabilize execution sync interoperability

flcl42:
## Summary

- Bind SNAP and full-sync sources to their own immutable launch parameters, including the source test specification itself, and wait for the Engine endpoint before starting sink checks.
- Configure direct static/trusted peers for Besu, Nimbus EL, and Reth, including Reth's Hive network ID and discovery isolation.
- Enforce LF endings for shell and jq launch helpers used in Linux containers.

## Validation

- Corrected seven-client Ethereum sync matrix: 76/76 passed (20 SNAP + 56 full-sync).
- The run logs confirm SNAP sources launch as snap and full-sync sources launch as full.
- Shell syntax checks, the affected simulator Go build, and diff checks passed.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+*.sh text eol=lf
+*.jq text eol=lf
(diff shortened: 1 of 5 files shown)


fjl:
Thanks!

PR #1586, opened 2026-07-23
fix(sim/lean): scope reqresp "client launch" failure to the failing client

MegaRedHand:
## Problem

The `reqresp` suite starts every selected client under a single shared `reqresp: client launch` test id. Because Hive attributes a test's `clientInfo` to whichever containers were started under that test id, a boot failure by one client (e.g. lantern or qlean) is attributed to **every** selected client, inflating the fail count / client score for healthy clients on the dashboard.

## Fix

Run each client's launch as its own smoke test via the existing `run_data_test_with_timeout` helper, which mints a fresh per-client Hive test id named `reqresp: client launch (<client>)`. A boot failure now scopes only to the client that actually failed.

The parent `reqresp: client launch` (registered as the suite's `PlannedTestSpec` in `main.rs`) stays as an always-passing scheduler with empty `clientInfo` — the same shape the `gossip`/`validation`/`rpc-compat` suites already produce. Child req/resp scenarios are still scheduled unconditionally.

Single file touched: `simulators/lean/src/scenarios/reqresp.rs`.

## Validation

Ran locally on macOS (`./hive --sim lean --sim.limit reqresp`):

- **ethlambda only**: 23/23 pass; the new `reqresp: client launch (ethlambda_devnet5)` row appears.
- **ethlambda + lantern**: 45 tests; lantern's single scenario failure (`blocks_by_root/max_request_limit`) is attributed to **lantern only**, ethlambda unaffected; per-client launch rows present for both clients, and the parent `reqresp: client launch` carries empty `clientInfo`.

Rendered both runs in a locally-built copy of the `lean-hiveview` frontend and headless-checked them (puppeteer + Chrome): **0 console errors, 0 uncaught page errors, 0 failed requests**; the per-client launch rows and client scoring render correctly.

## Caveat

These changes were **not** tried against the UI actually served at https://hive.leanroadmap.org/ — only against a locally-built copy of the `lean-hiveview` frontend. The live deployment may differ (commit, config, accumulated data), so the scoring/rendering there should be sanity-checked before merge.

### DIFF
--- a/simulators/lean/src/scenarios/reqresp.rs
+++ b/simulators/lean/src/scenarios/reqresp.rs
@@ -1,5 +1,5 @@
 use crate::utils::helper::{
-    start_post_genesis_sync_context, try_start_client_under_test, HelperGossipForkDigestProfile,
+    start_client_under_test, start_post_genesis_sync_context, HelperGossipForkDigestProfile,
     PostGenesisSyncTestData, LEAN_SPEC_SOURCE_VALIDATORS_EXCLUDING_V0,
 };
 use crate::utils::libp2p_mock::{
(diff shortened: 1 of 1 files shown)


KolbyML (approved):
:shipit: looks good

PR #1587, opened 2026-07-24
cmd/hivechain: add calltree contract and tx mods for callTracer testing

MysticRyuujin:
### What

Adds a `calltree` contract (geas) plus fixed-address predeploys of the existing `callme`/`callenv`/`callrevert` contracts, and two block modifiers:

- `tx-calltree` — one invocation produces every callTracer frame type: nested CALL (success with value / inner `Error(string)` revert), STATICCALL (success / write-protection failure), DELEGATECALL emitting a depth-2 log, CALLCODE, a precompile call, CREATE of a child that logs and self-destructs, and a top-level LOG1.
- `tx-callrevert` — a whole-transaction revert with a decodable revert reason (`"user error"`).

### Why

The execution-apis test chain currently contains no CALL-family opcodes, inner reverts, CREATE-opcode frames, or SELFDESTRUCTs, so `debug_trace*` fixtures cannot exercise nested call trees. This unblocks the callTracer standardization spec (execution-apis PR to follow).

Predeploy addresses (`0x9dcd…27d0–d3`) are hardcoded in `calltree.eas`, mirrored in `genesis.go`, and guarded by a test asserting the constants appear in the compiled bytecode. The callees are predeployed (rather than reusing the deploy-* mods' instances) because deployment addresses depend on modifier scheduling and cannot be baked into committed bytecode.


lightclient (approved):
LGTM

PR #1588, opened 2026-07-24
simulators/ethereum/rpc-compat: tracer-aware schema selection for speconly tests

MysticRyuujin:
speconly tests validate the client's response against the method's OpenRPC result schema. For the `debug_trace*` methods that schema is an `anyOf` whose named-tracer escape-hatch branch is deliberately unconstrained — so whole-schema validation accepts any value and speconly enforcement is vacuous for tracing methods.

This selects the `anyOf` branch matching the tracer the test requested (mirroring execution-apis' speccheck): no tracer → the opcode branch, `callTracer` → the Call tracer branch, anything else → the whole schema as before. A known tracer with no matching branch is an error rather than a fallback, so a spec title reword cannot silently disable validation. Branches are extracted from the raw document because the generated OpenRPC union types do not survive a second marshal round-trip.

Backward-compatible: against a spec without callTracer branches, opcode speconly fixtures still validate (the "Opcode tracer" branches have existed since the opcode-tracer standardization), and no fixture in that era requests callTracer. Pairs with the callTracer spec PR (execution-apis, to follow).

### DIFF
--- a/simulators/ethereum/rpc-compat/main.go
+++ b/simulators/ethereum/rpc-compat/main.go
@@ -94,6 +94,7 @@ func runTest(t *hivesim.T, c *hivesim.Client, test *rpcTest) error {
 		err       error
 		respBytes []byte
 		method    string
+		request   string
 	)
 
 	for _, msg := range test.messages {
@@ -101,6 +102,7 @@ func runTest(t *hivesim.T, c *hivesim.Client, test *rpcTest) error {
 			// Send request.
 			t.Log(">> ", msg.data)
 			method = gjson.Get(msg.data, "method").String()
+			request = msg.data
 			respBytes, err = postHttp(client, url, strings.NewReader(msg.data))
 			if err != nil {
 				return err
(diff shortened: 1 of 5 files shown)


lightclient (commented):
I'm not sure about this PR. It feels wrong to bolt on conditional schema fetching in rpc-compat. Shouldn't the traces be deterministic? That would remove the need to be smarter about which branch of the `anyOf` to check?

MysticRyuujin:
The problem is that there exists many named tracers, and geth is not the only client that implements named tracers. So the options then become don't validate any named tracer, or try to enforce and validate all named tracers at the same time. It's was a monumental task getting all the clients to agree to opcode tracer. This PR allows us to pick and choose which named tracers are standardized and which are open to schema only validation.

Speconly is for cases where the spec deliberately permits cross-client variance, so we only check the schema. But right now that check doesn't work for trace methods: the unconstrained named-tracer branch means clients can return garbage and pass. This PR is what makes speconly actually enforce anything there.

There is an argument to be made that all traces should be fully deterministic, but today they're not, and we need some way to convey that. As an example `logs[].index` has three different meanings across geth, erigon/reth, and nethermind, and the spec currently leaves it undefined (because who's right and who's wrong is subjective here)

speccheck on the execution-apis side already selects the branch the same way; this mirrors it.

MysticRyuujin:
cc @fjl

fjl:
I think this is something where we can also use the new validation script stuff (https://github.com/ethereum/execution-apis/pull/893)

PR #1589, opened 2026-07-27
Add Glamsterdam support to hivechain

MysticRyuujin:
Adds the hive prerequisites to test Glamsterdam in execution-apis. This is an integration branch for ethereum/execution-apis#851. It can be rebased as the dependent PRs land.

The tracer-aware schema selection in rpc-compat is removed from this branch. It stays in #1588 as a separate change.

## What it contains

- Amsterdam fork scheduling and hivechain block generation, built on geth `glamsterdam-devnet-8`.
- Engine API output for Amsterdam: `engine_newPayloadV5` with block access lists and slot numbers, `engine_forkchoiceUpdatedV4` with three parameters, hex-encoded execution requests, and EIP-8282 builder deposit and exit transactions.
- A `bal` output that dumps each Amsterdam block's access list. Each dump carries the header hash and the hash of the assembled list, and the dump also runs when import fails. That made the fork-boundary mismatch below diagnosable.
- A mapper fix in all six clients: declare BPO blob schedules only for scheduled forks. The mappers emitted `bpo1` to `bpo5` blob schedules unconditionally, with hardcoded fallback values. On a chain that ends at Amsterdam, geth resolved the active blob schedule to the phantom `bpo5` entry, so `eth_config` reported blob parameters that did not match the chain.
- The EIP-7997 deterministic factory contract in genesis. Devnet-7 geth inserted the contract at the start of the first Amsterdam block when it was absent from state. No other client does this, so the activation block's access list and state root diverged. Seeding the contract in genesis matches how the devnets place it. Thanks to @marioevz for the diagnosis and the fix.

## Dependencies

`go.mod` pins go-ethereum at `aa1f2fcf`, the tip of `glamsterdam-devnet-8`. The published `ethpandaops/geth:glamsterdam-devnet-8` image is built from the same commit, so the fixture generator and the client under test agree. The branch includes current `ethereum/hive` master.

## Testing (2026-08-20)

- The hive Go test suite passes. It now covers Amsterdam chain generation, the fork-boundary block, and the Engine API outputs.
- The Amsterdam chain regenerates byte-identical after the geth bump, so the existing fixtures stay valid.
- rpc-compat against 252 fixtures generated from this branch's hivechain (`-lastfork amsterdam -length 60`, Amsterdam at block 57), with all six clients on their `glamsterdam-devnet-8` images:

| Client | Passed |
|---|---|
| go-ethereum | 252/252 |
| Nethermind | 187/252 |
| Erigon | 184/252 |
| Besu | 176/252 |
| reth | 170/252 |
| ethrex | 46/252 |

## Known issues

- Most divergences, 56 to 64 per running client, are in the block-identity fields of `eth_simulateV1` responses: `hash`, `stateRoot`, `parentHash`, `blockAccessListHash`, `size`. The running clients produce different `blockAccessListHash` values from each other and from geth for the same simulated block. Nothing pins down the access list of a simulated block. This is an execution-apis question, not a defect in one client.
- ethrex import stops at block 1 with `Base fee per gas is incorrect`. This is its known pre-merge chain limitation, separate from the fixed V4 parameter handling in lambdaclass/ethrex#7074.
- speccheck reports one pre-existing failure: `eth_simulateV1/ethSimulate-two-blocks-with-complete-eth-sends` carries `{"code": -32015, "message": "out of gas"}`, and the spec requires a `-32015` message to match `^vm execution error.*`. This is a spec-versus-geth question to raise in execution-apis.


simulators/ethereum/rpc-compat/schema.go
```diff
@@ -4,32 +4,76 @@ import (
...
+	tracer map[string]map[string]*jsonschema.Schema
+}
+
+// forRequest returns the result schema for a speconly response. For debug
+// tracing methods, the anyOf branch matching the requested tracer is
+// selected — validating against the whole anyOf is vacuous because the
+// unconstrained named-tracer branch accepts anything. A known tracer with no
+// matching branch is an error, not a fallback, so a spec title reword cannot
+// silently disable validation.
+func (s methodSchemas) forRequest(method, request string) (*jsonschema.Schema, error) {
```
lightclient: is this required for glam support?

MysticRyuujin: It is not explicitly required, but it is a dependency on the execution-apis tests for the tracer named configs, it's basically just https://github.com/ethereum/hive/pull/1588

This PR was a meta PR for testing end to end and I included that so I could target one branch for testing.

I can remove it if that is your only objection, but then I'm still stuck with the same problem discussed over there.

lightclient: Yeah I would rather we focus on getting the glam tests for the basic RPCs in then we can focus on the tracers tests, is that okay?

MysticRyuujin: Yes, that's fine, I'll remove it now and re-push and we can deal with the named tracer stuff at a later date, not a problem.

MysticRyuujin:
Someone from geth maybe chime in but I think this can be updated to point back to main?

lightclient (approved):
LGTM!

fjl:
The description of this PR says the tracer-schema related changes are included in this, but they actually are: https://github.com/ethereum/hive/pull/1589/changes#diff-e0a6b2c60f85be691cd18bc590404e13214dde4e8f8a7819a4bb404f1b20cfcaR14

PR #1590, opened 2026-08-02
simulators/lean: Added 3 new test cases in validation and subsequent infrastructure for them 

groverInnovate:
## Summary

This PR adds three validation scenarios for Lean clients, makes the mock gossip node compatible with devnet5 blocks, and updates the Ream image to run the devnet5 binary with a compatible runtime.

### `clients/ream/Dockerfile`

- Uses a multi-stage build with separate devnet4 and devnet5 source images.
- Runs both binaries from an `ubuntu:24.04` runtime image, whose glibc version is compatible with the devnet5 binary.
- Installs runtime CA certificates and copies both Ream binaries explicitly from their source stages.
- Records version information for both binaries during the image build.

The previous image used the devnet4 image as the final runtime base. That base could not run the devnet5 binary because of its older system libraries.

### `simulators/lean/src/utils/libp2p_mock.rs`

- Configures the mock gossipsub node with anonymous validation which matches Lean clients.
- Preserves gossip topic hashes and captures the original payload bytes so a client-produced signed block can be decoded and replayed exactly.
- Adds support for decoding framed-Snappy, raw-Snappy, and raw SSZ block payloads.
- Adds Lean block tree-hash roots for matching gossip blocks with fork-choice entries.
- Raises the mock gossip transmit limit to 16 MiB. Ream permits large Lean gossip messages, while libp2p's default mock limit is 64 KiB; without this, valid devnet5 blocks were dropped before the test received them.

### `simulators/lean/src/scenarios/validation.rs`

- Adds a configurable genesis delay so tests can exercise clients before the relevant slot without racing client startup.
- Adds helpers to wait for client-generated blocks, match them against fork-choice state, and assert that rejected blocks do not enter the store.
- Adds three tests:
  1. **Future-slot horizon rejection**:  a known-parent block beyond the client's current time must not be imported or move fork choice.
  2. **Recovery after invalid gossip**:  invalid gossip must be rejected while the client continues to produce and import a later valid block.
  3. **Duplicate valid block idempotence**:  replaying the exact same valid signed block must leave exactly one fork-choice entry.

These tests cover client-facing validation and recovery behavior that was not previously exercised by the Lean Hive validation suite.

## Verification

- `cargo fmt --manifest-path simulators/lean/Cargo.toml -- --check`
- `cargo check --manifest-path simulators/lean/Cargo.toml`
- Full devnet5 validation suite: **7/7 tests passed** with Ream.

cc: @Dsorken @KolbyML

### DIFF
--- a/clients/ream/Dockerfile
+++ b/clients/ream/Dockerfile
@@ -2,22 +2,27 @@ ARG devnet4_baseimage=ghcr.io/reamlabs/ream
 ARG devnet4_tag=latest
 ARG devnet5_baseimage=ghcr.io/reamlabs/ream
 ARG devnet5_tag=latest-devnet5
+ARG runtime_baseimage=ubuntu:24.04
 
+FROM $devnet4_baseimage:$devnet4_tag AS devnet4
 FROM $devnet5_baseimage:$devnet5_tag AS devnet5
-FROM $devnet4_baseimage:$devnet4_tag
+FROM $runtime_baseimage AS runtime
+
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends ca-certificates \
+    && rm -rf /var/lib/apt/lists/*
 
 COPY validators.yaml /assets/lean/validators.yaml
-RUN cp /usr/local/bin/ream /usr/local/bin/ream-devnet4
+COPY --from=devnet4 /usr/local/bin/ream /usr/local/bin/ream-devnet4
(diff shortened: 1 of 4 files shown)


KolbyML:
@groverInnovate please remove devnet4 support from the ream docker image, as devnet4 is no longer supported by hive

groverInnovate:
Done @KolbyML

KolbyML (approved):
:shipit: looks good

PR #1592, opened 2026-08-06
Preserve Nethermind genesis slot number

flcl42:
## What

Pass the `slotNumber` genesis field through Hive's Nethermind mapper.

## Why

The mapper converted Hive's Geth-style genesis into a Nethermind chain spec but dropped `slotNumber`. EIP-7843 fixtures therefore produced a different genesis block hash in Nethermind, and the first forkchoice update remained syncing.

## Impact

Nethermind now receives the fixture's configured genesis slot number. Existing genesis files without this field retain the current behavior because the mapped value is null and Nethermind treats it as unset.

## Checks

- The previously failing EIP-7843 SLOTNUM consume-engine case passes.
- Hive Go packages pass, apart from the existing `cmd/hiveview` test whose `/` path assumption is incompatible with Windows.

Or we can just https://github.com/ethereum/hive/pull/1593

### DIFF
--- a/clients/nethermind/mapper.jq
+++ b/clients/nethermind/mapper.jq
@@ -262,6 +262,7 @@ def clique_engine:
     "blobGasUsed": .blobGasUsed,
     "excessBlobGas": .excessBlobGas,
     "parentBeaconBlockRoot": .parentBeaconBlockRoot,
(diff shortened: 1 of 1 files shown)


spencer-tb:
CIFO #1593

PR #1593, opened 2026-08-06
Use Geth-style genesis for Nethermind

flcl42:
## What

Replace Hive's Nethermind-specific Parity chain-spec conversion with the Geth-style genesis mapper already used by the Geth client, then point Nethermind at the mapped `/genesis.json`.

The mapper keeps `config` first for Nethermind's format detection and adds `0x` prefixes to allocation keys for Nethermind's strict address parser.

## Why

Maintaining a separate Parity-style conversion duplicates the fork schedule and must explicitly copy every genesis header field. It dropped the EIP-7843 `slotNumber` field, causing Nethermind to derive the wrong genesis hash. Nethermind can now load Geth-style genesis files directly, so retaining the input schema avoids this class of field-loss bug.

## Impact

Nethermind receives the same environment-derived fork configuration as Geth while preserving fixture-level genesis fields. The old generated `/chainspec/test.json` is no longer used.

This is an independent broader alternative to the one-field mapper fix in #1592.

## Checks

- The previously failing EIP-7843 SLOTNUM consume-engine case passes.
- The Glamsterdam filtered sweep reached 2,313 passed cases with zero failures or errors before it was intentionally stopped.
- Hive's core Go packages pass; the full suite still has the existing `cmd/hiveview` test whose `/` path assumption is incompatible with Windows.

### DIFF
--- a/clients/nethermind/mapper.jq
+++ b/clients/nethermind/mapper.jq
@@ -16,313 +16,108 @@ def remove_empty:
   )
 ;
 
-# Converts number to hex, from https://rosettacode.org/wiki/Non-decimal_radices/Convert#jq
-def int_to_hex:
-  def stream:
-    recurse(if . > 0 then ./16|floor else empty end) | . % 16 ;
-  if . == 0 then "0x0"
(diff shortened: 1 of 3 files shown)


spencer-tb (approved):
Wow nice! LGTM :)

PR #1594, opened 2026-08-06
simulators/lean: fix gossip mock duplicate-publish and gossip snappy framing

MegaRedHand:
Two gossip tests currently fail on **every** lean client on [hive.leanroadmap.org](https://hive.leanroadmap.org/):

```
gossip: caches orphan block              0/6
gossip: deduplicates duplicate messages  0/6
(the other three gossip tests)           6/6
```

Both fail on the same line, before anything reaches the client under test:

```
should publish parent block: "Failed to publish message: Duplicate"
should publish second copy of invalid block: "Failed to publish message: Duplicate"
```

This is a mock bug, not a client bug. The regression starts at d764a937 (#1590), which is also where the run went from 5 failures to 12.

## 1. Message ids collide for anonymous publishers

#1590 switched the mock to `MessageAuthenticity::Anonymous`. That part is right: lean clients use anonymous gossipsub validation, so a signed envelope gets rejected before block validation. But the default `message_id_fn` was left in place, and it is:

```rust
// libp2p-gossipsub 0.49.4, src/config.rs:518
let mut source_string = if let Some(peer_id) = message.source.as_ref() {
    peer_id.to_base58()
} else {
    PeerId::from_bytes(&[0, 1, 0]).expect("Valid peer id").to_base58()
};
source_string.push_str(&message.sequence_number.unwrap_or_default().to_string());
```

An anonymous message carries neither a source nor a sequence number, so **every message the mock publishes hashes to the same id**. `Behaviour::publish` checks `duplicate_cache.contains(&msg_id)` and returns `PublishError::Duplicate` for the second publish regardless of payload. The three tests that publish once are unaffected; the two that publish twice cannot pass.

`duplicate_cache_time` is not a workaround here: `DuplicateCache::contains` does not sweep expired entries (only `insert` does, via `TimeCache::entry`), and the sweep happens after the check.

The same collision has a quieter effect on the receive path — `handle_received_message` drops everything after the first message for the same reason, which silently caps `take_gossip_messages()` that #1590's own new validation tests rely on.

**Fix:** derive the id from the payload using the lean networking spec's scheme (`sha256(domain ‖ topic_len ‖ topic ‖ data)[..20]`), so the mock agrees with clients on what counts as a duplicate.

Re-sending byte-identical bytes still has to get past the publisher-side cache, so `MockNode::publish_duplicate` bumps a mock-local salt mixed into the id. The salt never travels on the wire, so the client still receives a genuine duplicate and its own deduplication is what the test exercises.

## 2. Gossip blocks were framed snappy

While verifying the above: `encode_gossip_block` compressed with `FrameEncoder`, but gossip payloads carry **raw** snappy — leanSpec decompresses gossip with the raw decoder only (`node/networking/gossipsub/behavior.py`), and framing belongs to req/resp.

Nothing failed as a result, which is the problem. Both of these tests only assert that the client did not import the block and stayed responsive, and both hold trivially when the client cannot decompress the payload. A local run against ethlambda logged, for every block the mock published:

```
ERROR Failed to decompress gossipped block err=snappy: corrupt input
```

So fixing only the first issue would turn the tests green without them testing anything.

`decode_gossip_block` had the mirror image of this: it took whichever of raw snappy, framed snappy or bare SSZ happened to parse. Its input is bytes a client published on a gossip topic, so raw snappy is the only conformant encoding, and accepting the other two would hide a client's framing defect and let it pass the validation tests built on that decode. It is now strict, and reports which of the two steps failed.

## Verification

Local hive run against `ethlambda_devnet5`:

| suite | before | after |
|---|---|---|
| gossip | 4/6 | **6/6** |
| validation (exercises `decode_gossip_block`) | 7/7 | **7/7** |

The client logs show the gossip tests became meaningful rather than merely green. Before, every published block was a decompress error. After:

```
test 5  Received block from gossip slot=2 proposer=0 parent_root=cacacaca   <- orphan child
        Received block from gossip slot=1 proposer=0 parent_root=00000000   <- parent
test 6  Received block from gossip slot=1 proposer=9999                     <- once; client dropped the duplicate
test 4  Failed to decompress ... (invalid header)                           <- the deliberate garbage, still correct
```

Test 5 receiving two payloads where it previously received one is the direct evidence that the second publish now reaches the wire. Test 6 receiving exactly one is the direct evidence that the duplicate went out and the client deduplicated it. The validation suite staying at 7/7 shows a real client envelope decodes under the strict decoder.

Two things worth flagging for review:

- The strict decoder can newly fail a client that publishes framed snappy on a gossip topic. That is the intent, but it is a behaviour change beyond the reported failure.
- On a framing deviation, `wait_for_client_generated_valid_block` still panics with "client did not gossip and import a valid block within N seconds"; the framing error is appended as `last_decoding_error` rather than leading. That message predates this PR, so I left it alone, but it is worth sharpening.

### DIFF
--- a/simulators/lean/src/scenarios/gossip.rs
+++ b/simulators/lean/src/scenarios/gossip.rs
@@ -328,7 +328,9 @@ dyn_async! {
         mock.publish(block_topic.clone(), block_bytes.clone())
             .expect("should publish first copy of invalid block");
         sleep(Duration::from_secs(1)).await;
-        mock.publish(block_topic, block_bytes)
+        // Re-send the exact same bytes, so the client sees a real duplicate.
+        // The mock has to bypass its own publisher-side duplicate cache to do it.
+        mock.publish_duplicate(block_topic, block_bytes)
             .expect("should publish second copy of invalid block");
 
         mock.process_events_for(Duration::from_secs(5)).await;
--- a/simulators/lean/src/utils/libp2p_mock.rs
+++ b/simulators/lean/src/utils/libp2p_mock.rs
@@ -1,5 +1,7 @@
 use std::collections::{HashMap, HashSet};
 use std::io::{self, Cursor, Read, Write};
+use std::sync::atomic::{AtomicU64, Ordering};
+use std::sync::Arc;
 use std::time::Duration;
 
 use alloy_primitives::B256;
@@ -9,7 +11,8 @@ use futures::prelude::*;
 use libp2p::{
     gossipsub::{
         Behaviour as GossipsubBehaviour, ConfigBuilder as GossipsubConfigBuilder,
-        Event as GossipsubEvent, IdentTopic, MessageAuthenticity, TopicHash, ValidationMode,
+        Event as GossipsubEvent, IdentTopic, Message as GossipsubMessage, MessageAuthenticity,
+        MessageId, TopicHash, ValidationMode,
     },
     request_response::{
         self, Behaviour as RequestResponseBehaviour, Codec, Event as ReqRespEvent,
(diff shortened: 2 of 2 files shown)


KolbyML (approved):
:shipit: looks good

PR #1595, opened 2026-08-11
clients/nethermind: only mark post-merge genesis when TTD is set

flcl42:
## Summary

- Emit `terminalTotalDifficulty` only when `HIVE_TERMINAL_TOTAL_DIFFICULTY` is configured.
- Do not emit Geth-specific `terminalTotalDifficultyPassed` from Nethermind's mapper.
- Preserve the existing Engine API configuration for explicit-TTD runs.

## Root cause

The Geth-style Nethermind mapper defaulted TTD to the maximum value when the Hive environment did not provide a terminal difficulty. The matching RPC configuration did not expose an Engine API URL in that case. Nethermind consequently exited during merge initialization with `Engine module wasn't configured on any port`, before the devp2p discovery tests ran.

## Validation

- JQ matrix passed for unset, zero, and explicit TTD configurations before the final unused-field removal.
- Focused Hive `discv4` run passed 16/16 tests with the patched client image.
- Focused Hive Go packages passed.
- `git diff --check` passed after the final removal.
- Full `go test ./...` remains limited by the unrelated Windows `cmd/hiveview/TestBuildAllBundles` esbuild error: `The working directory "/" is not an absolute path`.

### DIFF
--- a/clients/nethermind/mapper.jq
+++ b/clients/nethermind/mapper.jq
@@ -28,18 +28,6 @@ def to_bool:
   end
 ;
 
-# Only mark the chain as post-merge when Hive supplied a terminal total difficulty.
-def merge_config:
-  if env.HIVE_TERMINAL_TOTAL_DIFFICULTY == null then
-    {}
-  else
-    {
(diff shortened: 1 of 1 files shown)


spencer-tb (approved):
LGTM!

PR #1602, opened 2026-08-25
fix(simulator): Correctly enforce fork-choice field names

### LINKED ISSUE #1599: Fork-Choice Assertion Branch Mismatch
## Description

6/8 of the assertion branches in the `assert_fork_choice_checks` function in the Lean simulator do not match the fixture field names. The `fork-choice` suite silently drops the justification, finalization, head-root and safe-target expectation the fixture pins.

| Branch reads | Occurrences | Fixtures actually pin (frequency) |
|---|---|---|
| `headSlot` | 440 | Same key. Works |
| `time` | 17 | Same key. Works  |
| `headRoot` | 0 | `headRootLabel` (210) |
| `justifiedCheckpoint/slot` | 0 | `latestJustifiedSlot` (71) |
| `justifiedCheckpoint/root` | 0 | `latestJustifiedRootLabel` (29) |
| `finalizedCheckpoint/slot` | 0 | `latestFinalizedSlot` (59) |
| `finalizedCheckpoint/root` | 0 | `latestFinalizedRootLabel` (17) |
| `safeTarget` | 0 | `safeTargetSlot` (15), `safeTargetRootLabel` (11) |

## Proposed Solution [WIP]
Fix naming and root label resolution and assert the right values from `steps.checks` and `steps.storeSnapshot` against `DriverSnapshot`

richardgreg:
The PR corrects the names and key lookup in the `assert_fork_choice_checks` function. It correctly compares the appropriate values, i.e., between the fixture pins and the client's store `DriverSnapshot`. Unit tests included.

It effectively adds 397 new check instances

Fixes #1599

### DIFF
--- a/simulators/lean/src/scenarios/spec_assets.rs
+++ b/simulators/lean/src/scenarios/spec_assets.rs
@@ -296,6 +296,15 @@ async fn post_json_raw(client: &Client, path: &str, payload: &Value) -> reqwest:
         .unwrap_or_else(|err| panic!("failed to POST {url}: {err}"))
 }
 
+fn driver_step_request(step: &Value) -> Value {
+    let mut request = step.clone();
+    if let Some(object) = request.as_object_mut() {
+        object.remove("checks");
+        object.remove("storeSnapshot");
+    }
+    request
+}
+
 async fn run_fork_choice_fixture(client: &Client, fixture: &SpecFixtureCase) {
     let steps = fixture
         .case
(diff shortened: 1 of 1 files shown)


richardgreg:
Hi @Dsorken, please take a look

cc @KolbyML

richardgreg:
converted to draft. runs and reports against each client are still needed for a seamless merge

richardgreg:
In my initial implementation, I added a step in the assertion function that stripped `steps[].checks` and `steps{}.storeSnapshot` from the client's request body. My reasoning was that they contained the correct expected values, and they shouldn't be sent to the client so it doesn't simply echo them back. But gean and qlean's driver depends on receiving `checks` to compute the safe target root correctly. I will treat it as a separate issue if we believe it is a necessary step.

Below are the results from the A/B test live runs: feature branch vs latest commit on upstream/master

| Client | Branch pass/total | Baseline pass/total | Newly failing | Newly passing |
|---|---|---|---|---|
| ream | 91/123 | 91/123 | 0 | 0 |
| ethlambda | 123/123 | 123/123 | 0 | 0 |
| gean | 123/123  | 123/123 | 0 | 0 |
| grandine_lean | 71/123 | 71/123 | 0 | 0 |
| lantern | 121/123 | 121/123 | 0 | 0 |
| zeam | 122/123 | 122/123 | 0 | 0 |
| qlean | 56/123  | 56/123 | 0 | 0 |

*Newly failing means it fails on the feature branch and passes on the baseline. Newly passing means it passes on the feature branch and fails on the baseline.

PR #1604, opened 2026-08-27
simulators/ethereum/engine: apply withdrawals CL mock config

taratorio:
The `WithdrawalsBaseSpec` wasn't setting `SlotsToSafe` and `SlotsToFinalized` to 32 and 64 respectively as in the `BaseSpec`. This caused some tests in the `withdrawals` suite to change the `FinalisedHash` backwards or on a different fork. This should not be allowed as the `FinalisedHash` is final and is only allowed to move forward on the same chain. This issue got surfaced in Erigon while adding some new logic for support for longer reorgs during non-finality and adding validation for this as part of it. With `SlotsToFinalized` being set to 64 the tests pass because the `FinalisedHash` correctly sits further behind than the re-org point the tests are exercising.

### DIFF
--- a/simulators/ethereum/engine/suites/withdrawals/spec_unit_test.go
+++ b/simulators/ethereum/engine/suites/withdrawals/spec_unit_test.go
@@ -1,9 +1,12 @@
 package suite_withdrawals
 
 import (
+	"math/big"
 	"testing"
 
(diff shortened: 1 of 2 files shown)


spencer-tb (approved):
LGTM!

PR #1606, opened 2026-09-03
internal/libhive,simulators/ethereum/eels: support optional authenticated GitHub clones, redact token from recorded commands

danceratopz:
Adds an optional `github_token` build-arg to the five eels simulator Dockerfiles (`consume-engine`, `consume-enginex`, `consume-rlp`, `consume-sync`, `execute-blobs`). When set, the `git clone` of execution-specs and the subsequent branch fetch authenticate against GitHub; when unset (the default), the build behaves exactly as today. hive also redacts the token from the command line it records in the results and serves to simulators.

## Motivation

The simulator image builds clone execution-specs anonymously, and GitHub rate-limits anonymous git-over-HTTPS per source IP. On shared CI runners this intermittently kills the build before any test runs:

```
Step 6/13 : RUN git clone --depth 1 https://github.com/ethereum/execution-specs.git && ...
fatal: could not read Username for 'https://github.com': No such device or address
fatal: expected flush after ref listing
```

GitHub answers the anonymous `info/refs` request with an HTTP 401 auth challenge when the IP has exceeded its unauthenticated quota. Git then tries to prompt for credentials, and there is no terminal inside a Docker build, so the clone dies and hive reports `image build failed`. Git's only code path to a username prompt is a 401 response (`http.c` handles 429 separately and never prompts), so the failure mode identifies the cause even though the HTTP exchange is not visible in build logs.

We have observed this on ethpandaops runners across ethpandaops/hive-tests dashboard runs (e.g. [consume-engine, nethermind](https://github.com/ethpandaops/hive-tests/actions/runs/33615372383/job/100199901867) and [consume-engine, reth](https://github.com/ethpandaops/hive-tests/actions/runs/33615372383/job/100199901866)), ethereum/execution-specs CI (e.g. [this Engine job](https://github.com/ethereum/execution-specs/actions/runs/33661674686/job/100355625044)), and other EF repo infra. GitHub [tightened unauthenticated rate limits in May 2025](https://github.blog/changelog/2025-05-08-updated-rate-limits-for-unauthenticated-requests/), explicitly covering HTTPS clones, and recommends authenticated requests; authenticating the clone is the fix GitHub documents rather than a workaround.

## Mechanism

When `github_token` is non-empty, the RUN step exports `GIT_CONFIG_*` environment variables that add an `http.https://github.com/.extraheader` authorization header, the same mechanism `actions/checkout` uses. The header applies to both the clone and the optional branch fetch, and the token is never written to the image filesystem or to git config on disk.

One caveat is documented in the Dockerfile comment: hive uses the classic Docker builder, which records build-arg values in the image history, so callers should pass short-lived tokens only. The natural choice on CI is the Actions job token (`github.token`), which is repo-scoped, works for cloning public repos, and expires when the job ends.

Hive records its command line in every results JSON and serves it via the `/hive` endpoint, so the token passed through `--sim.buildarg` would otherwise be published with the results. `NewTestManager` now redacts the values of sensitive build arguments from the recorded command, and the build-arg key match is case-insensitive.

## Usage

```
./hive --sim ethereum/eels/consume-engine --client reth \
    --sim.buildarg github_token=$GITHUB_TOKEN ...
```

A companion PR wires this up for the ethpandaops/hive-tests dashboard workflows.

## Validation

The RUN shell logic was tested standalone under dash (the `/bin/sh` of the bookworm-slim base images): with an empty token the request is anonymous and succeeds as before; with an invalid token GitHub rejects the operation (`remote: Invalid username or token`), confirming the header is actually sent; with a valid token the authenticated clone succeeds. Unit tests in `internal/libhive/redact_test.go` cover the command-line redaction for both `--sim.buildarg KEY=VALUE` and `--sim.buildarg=KEY=VALUE`, and the case-insensitive client-file filter.

### DIFF
--- a/internal/libhive/redact.go
+++ b/internal/libhive/redact.go
@@ -1,0 +1,54 @@
+package libhive
+
+import "strings"
+
+// redactedValue replaces the value of a sensitive build argument wherever hive
+// records or serves its own invocation: the run metadata in the results JSON
+// and the /hive endpoint of the simulation API.
+const redactedValue = "<redacted>"
+
+// isSensitiveBuildArg reports whether the value of a build argument must not be
+// recorded. The comparison is case-insensitive because client files and
+// --sim.buildarg conventionally use lowercase keys (github_token) while
+// excludedBuildArgs lists them in uppercase.
+func isSensitiveBuildArg(key string) bool {
+	return excludedBuildArgs[strings.ToUpper(key)]
+}
+
+// redactCommandArgs returns a copy of a hive command line in which the values of
+// sensitive build arguments are replaced by redactedValue. It recognises both the
+// "--flag KEY=VALUE" and the "--flag=KEY=VALUE" spelling of any flag whose name
+// ends in ".buildarg" (currently --sim.buildarg). All other arguments are copied
+// unchanged, and the input slice is not modified.
+func redactCommandArgs(args []string) []string {
+	out := make([]string, len(args))
+	copy(out, args)
+	for i := 0; i < len(out); i++ {
+		flag := strings.TrimLeft(out[i], "-")
+		if flag == out[i] {
+			continue // not a flag
+		}
+		name, inline, hasInline := strings.Cut(flag, "=")
+		if !strings.HasSuffix(name, ".buildarg") {
+			continue
+		}
+		if hasInline {
+			// inline is a suffix of out[i]: keep the "--name=" prefix as written.
+			out[i] = out[i][:len(out[i])-len(inline)] + redactBuildArg(inline)
+		} else if i+1 < len(out) {
+			i++
+			out[i] = redactBuildArg(out[i])
+		}
+	}
+	return out
+}
+
+// redactBuildArg redacts the value of a KEY=VALUE build argument when KEY is
+// sensitive. Anything else is returned as is.
+func redactBuildArg(kv string) string {
+	key, _, ok := strings.Cut(kv, "=")
+	if !ok || !isSensitiveBuildArg(key) {
+		return kv
+	}
+	return key + "=" + redactedValue
+}
--- a/internal/libhive/redact_test.go
+++ b/internal/libhive/redact_test.go
@@ -1,0 +1,90 @@
+package libhive
+
+import (
+	"reflect"
+	"slices"
+	"testing"
+)
+
+func TestRedactCommandArgs(t *testing.T) {
+	tests := []struct {
+		name string
+		in   []string
+		want []string
+	}{
+		{
+			name: "separate value argument",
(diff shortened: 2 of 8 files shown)


internal/libhive/redact.go
```diff
@@ -0,0 +1,54 @@
...
+// redactedValue replaces the value of a sensitive build argument wherever hive
+// records or serves its own invocation: the run metadata in the results JSON
+// and the /hive endpoint of the simulation API.
+const redactedValue = "<redacted>"
+
+// isSensitiveBuildArg reports whether the value of a build argument must not be
+// recorded. The comparison is case-insensitive because client files and
+// --sim.buildarg conventionally use lowercase keys (github_token) while
+// excludedBuildArgs lists them in uppercase.
+func isSensitiveBuildArg(key string) bool {
```
spencer-tb, line 14: internal/libdocker/builder.go` appends every `arg.Value` to `logctx` before emitting the structured `building image` log, so `github_token` is still logged verbatim :)

spencer-tb (commented):
I think credentials should stay out of Docker build arguments entirely. Docker explicitly warns that ARG and ENV are inappropriate for secrets because their values may persist in image metadata and history: https://docs.docker.com/build/building/variables/#build-arguments

I (codex) drafted an alternative that keeps the token out of Docker build arguments entirely: https://github.com/danceratopz/hive/pull/3

It adds an environment-backed `--sim.buildsecret` interface and passes the token through an optional BuildKit secret mount. The token therefore stays out of Hive's argv, structured logs, result metadata, and Docker image history, while builds without a token continue cloning anonymously.

Let me know what you think! :)

danceratopz:
> I think credentials should stay out of Docker build arguments entirely. Docker explicitly warns that ARG and ENV are inappropriate for secrets because their values may persist in image metadata and history: https://docs.docker.com/build/building/variables/#build-arguments
> 
> I (codex) drafted an alternative that keeps the token out of Docker build arguments entirely: [danceratopz#3](https://github.com/danceratopz/hive/pull/3)
> 
> It adds an environment-backed `--sim.buildsecret` interface and passes the token through an optional BuildKit secret mount. The token therefore stays out of Hive's argv, structured logs, result metadata, and Docker image history, while builds without a token continue cloning anonymously.
> 
> Let me know what you think! :)

Thanks! I took another look and came up with another suggestion on top of your suggestion! At this point, I think we should tackle the problem head on with a better solution that we've discussed at times: Publishing images with pre-loaded fixtures for the `eels/` simulators. This would greatly improve devex as switching between different test fixture releases and even simulator versions will be blazing quick instead of the clone+fixture-download for every run we have now. This would avoid the clone throttling up front.

PR #1607, opened 2026-09-09
internal/libhive: support simulator build configurations via --config

danceratopz:
Adds `--config` to load client and simulator build configurations from a single YAML file. The file is a flat list in which each entry is either a client (`client` key) or a simulator (`simulator` key) with an optional Dockerfile extension and build arguments. Client entries use the existing `--client-file` format, so `--client-file` becomes an alias of `--config`: existing client files keep working unchanged and can gain simulator entries in place. This enables per-simulator configuration and selection of alternate Dockerfiles when available. The second commit replaces the initial `--sim.file` approach following review; `--sim.file` and `--sim-file` are removed.

```yaml
- client: go-ethereum
  build_args:
    baseimage: docker.io/ethereum/client-go
    tag: latest
- simulator: ethereum/eels/consume-engine
  build_args:
    fixtures: stable@latest
    branch: ""
    disable_strict_exception_matching: nimbus-el
```

Run with `./hive --config devnet.yaml`. Entries run in file order; `--client` filters the client entries when set explicitly, `--sim` filters the simulator entries using its existing regexp semantics, and `--sim.buildarg` overrides matching file arguments. A file without simulator entries matches `--sim` against the inventory as before, and a file without client entries uses the `--client` list, so client-only and simulator-only files are both valid. Unknown fields, entries with both or neither key, `nametag` on a simulator, unknown or duplicate simulators, missing alternate Dockerfiles, and empty files are rejected. Passing both `--config` and `--client-file` is an error. The file is optional. Without it, `--client`, `--sim`, `--sim.buildarg` and the default `Dockerfile` behave exactly as before, including `hive_context.txt` build contexts.

[`docs/commandline.md`](https://github.com/danceratopz/hive/blob/add-sim-file/docs/commandline.md#simulator-build-parameters) documents the file format and currently supported EELS examples. The documentation is intentionally minimal: no simulator ships an alternate Dockerfile yet, so the `dockerfile` field is described generically and the examples use only build arguments the current EELS simulators accept. The change adds configuration plumbing; simulator Dockerfiles and image publishing remain separate work.

Validation: `gofmt`, `go test -race . ./internal/... ./hivesim/...`, `go vet . ./internal/... ./hivesim/...`, `go build`, and `git diff --check` passed on the branch head. Regression tests cover parsing of client-only, simulator-only and mixed files, selection, argument precedence and isolation, and Docker build requests for default and alternate Dockerfiles with overridden contexts. CLI smoke checks verified both flag spellings, the alias conflict error, and error reporting for missing or invalid files. No live Docker simulation was run for this change.

### Motivation

We'd like to start publishing images for EELS simulators and this mechanism is required to maintain support of building an arbitrary branch/ref from source (each eels `Dockerfile` would move to `Dockerfile.git` and be replaced with a minimal Dockerfile that pulls in a pre-built image).

### DIFF
--- a/docs/commandline.md
+++ b/docs/commandline.md
@@ -74,6 +74,45 @@ arguments are:
 - `github`: For client Dockerfiles building from git, this setting can be used to change
    the source code repository (fork) on GitHub. Example: `ethereum/go-ethereum`.
 
+### Simulator Build Parameters
+
+Use `--sim.file` (or its alias `--sim-file`) to supply simulator build configurations
+in a YAML file:
+
+    ./hive --sim.file simulators.yaml --client go-ethereum
+
+For example, the EELS consume simulators currently build from source. A configuration
+using supported arguments is:
+
+    - simulator: ethereum/eels/consume-engine
+      build_args:
+        fixtures: stable@latest
+        branch: ""
+        disable_strict_exception_matching: nimbus-el
+
+Check each simulator's Dockerfile for its supported build arguments and defaults.
+The EELS `consume-*` simulators accept `fixtures` (the fixture input, default
+`stable@latest`) and `branch` (the execution-specs Git ref; empty uses the repository's
+default branch). `consume-engine` and `consume-enginex` also accept
+`disable_strict_exception_matching`, which defaults to `nimbus-el` and is passed to
+consume's corresponding option.
+
+Each entry supports:
+
+- `simulator`: A known simulator directory under simulators/. Each simulator may appear once.
+- `dockerfile`: An optional extension, such as `git` for `Dockerfile.git`. The selected
+  file must exist in the simulator directory. If omitted, uses `Dockerfile`.
+- `build_args`: Arguments passed to the simulator's Dockerfile.
+
+All entries run in file order unless `--sim` is supplied to filter them using its usual
+regular expression matching. Repeated `--sim.buildarg NAME=VALUE` options apply to every
+selected simulator and override matching arguments from the file. The simulator's
+`hive_context.txt`, if present, still determines its build context. In `--dev` mode,
+simulators are not built or run.
+
+Without `--sim.file`, simulator selection and builds behave as before: `--sim` selects
+simulators, their default `Dockerfile` is used, and `--sim.buildarg` supplies build arguments.
+
 ### Docker Options
 
 `--docker.pull`: Setting this option makes hive re-pull the base images of all built
--- a/docs/simulators.md
+++ b/docs/simulators.md
@@ -11,7 +11,9 @@ docker.
 Simulators live in the `simulators/` directory of the hive repository. There is a
 dedicated sub-directory for every simulator. When hive runs a simulation, it first builds
 an image using `docker build` in the simulator directory, using the Dockerfile. The image
-must contain all resources needed for testing.
+must contain all resources needed for testing. A simulator directory may also provide
+alternate `Dockerfile.<ext>` files, which are selected through the `--sim.file`
+configuration described in [Hive Commands].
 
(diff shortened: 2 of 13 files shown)


spencer-tb (commented):
Thanks for adding this support! 2 potential paths before approving on my side :)

Let me know what you prefer, I'm happy with both!

**1) Keep the current approach, and align the spelling.**

Drop the `--sim-file` alias (`hive.go:95`) so `--sim.file` is the only form, and add `--client.file`:
```bash
./hive --sim.file simulators.yaml --client.file clients.yaml
```

**2) Or switch to one file and one flag for both.** Both parsers exist, so this is mostly wiring, i.e just use a single `--configs` flag:
```yaml
# devnet8.yaml
clients:
  - client: go-ethereum
    build_args: { baseimage: docker.io/ethereum/client-go, tag: latest }
simulators:
  - simulator: ethereum/eels/consume-engine
    build_args: { tag: glamsterdam-devnet-8 }
  - simulator: ethereum/eels/consume-rlp
    dockerfile: git
    build_args: { branch: devnets/glamsterdam/8, fixtures: tests-glamsterdam-devnet@v8.1.4 }
```

```bash
./hive --config devnet8.yaml
```

`--client` and `--sim` stay as optional filters over the file, `--client-file` stays or is removed for `--config`.

danceratopz:
> Thanks for adding this support! 2 potential paths before approving on my side :)
> 
> Let me know what you prefer, I'm happy with both!

Thanks, good suggestions! I think it's cleaner to only require specifying a single file. `--sim.file` is gone!

> **2) Or switch to one file and one flag for both.** Both parsers exist, so this is mostly wiring, i.e just use a single `--configs` flag:
> 
> ```yaml
> # devnet8.yaml
> clients:
>   - client: go-ethereum
>     build_args: { baseimage: docker.io/ethereum/client-go, tag: latest }
> simulators:
>   - simulator: ethereum/eels/consume-engine
>     build_args: { tag: glamsterdam-devnet-8 }
>   - simulator: ethereum/eels/consume-rlp
>     dockerfile: git
>     build_args: { branch: devnets/glamsterdam/8, fixtures: tests-glamsterdam-devnet@v8.1.4 }
> ```
> 
> ```shell
> ./hive --config devnet8.yaml
> ```
> 
> `--client` and `--sim` stay as optional filters over the file, `--client-file` stays or is removed for `--config`.

There's a small tweak to the file format, so that we can alias `--client-file` to `--config` and keep it backwards compatible. Instead of top-level `clients:` and `simulators:` sections, the file stays a flat list and each entry is identified by its `client` or `simulator` key:

```yaml
- client: go-ethereum
  build_args:
    baseimage: docker.io/ethereum/client-go
    tag: latest
- simulator: ethereum/eels/consume-engine
  build_args:
    fixtures: stable@latest
    branch: ""
- simulator: ethereum/eels/consume-rlp
  build_args:
    fixtures: stable@latest
    branch: ""
```

Then:

```console
./hive --config devnet.yaml
```

Selection works as before: `--client` filters the client entries when set explicitly and `--sim` filters the simulator entries. A file without simulator entries falls back to matching `--sim` against the inventory, and a file without client entries falls back to the `--client` list, so a simulators-only file is valid too. E.g.,

```console
./hive --config devnet.yaml --client go-ethereum --sim consume-engine
```

The only potential disadvantage is that we insist on a single file, so you can't mix and match client and simulator config files. But I don't think there's a real use case for mixing them, since our images are tied to mainnet or specific devnet configuration anyway.

I just pushed this change, but don't mind doing another round to get this right if need be!

spencer-tb (approved):
Awesome! I like that `--client-file` still works :)

Approved from my side! :)

PR #1608, opened 2026-09-10
simulators/ethereum/eels: add consume-reorg simulator (blockchain_test_engine_reorg fixtures)

CPerezz:
Adds `simulators/ethereum/eels/consume-reorg`, a hive simulator that runs EELS `consume reorg` against `blockchain_test_engine_reorg` fixtures (ethereum/execution-specs#3556): a DAG of engine payloads plus an ordered list of Engine API / JSON-RPC steps, each with its set of spec-legal outcomes. One fresh client per fixture, pass/fail per test, no runtime knobs.

Removes `simulators/ethereum/engine/suites/engine/reorg.go`. Its scenarios — `SidechainReOrgTest`, `TransactionReOrgTest` (ReOrgOut / ReOrgDifferentBlock / NewPayloadOnRevert / ReOrgBackIn), `ReOrgBackToCanonicalTest` (depth 5 and 10), `ReOrgBackFromSyncingTest`, `ReOrgPrevValidatedPayloadOnSideChainTest`, `SafeReOrgToSideChainTest` — are re-expressed as declarative fixtures in execution-specs `tests/reorg/` and executed by the new simulator, so the CLMock-based versions are no longer needed.

Client wrappers: `HIVE_ENGINE_MAX_REORG_DEPTH` sets the Engine API reorg depth cap for go-ethereum (`--engine.maxreorgdepth`) and erigon (`MAX_REORG_DEPTH`); `HIVE_EXPECT_DEEP_REORGS=1` keeps its previous meaning (512).

The Dockerfile follows the other EELS simulators: `fixtures` takes a release tag, a URL or a staged `fixtures.tar.gz`, and `branch` selects a branch of the source repo. It adds one build arg, `repo`, so the simulator can be built from a fork while the fixture format is still under review — until execution-specs#3556 merges, `ethereum/execution-specs` has no `blockchain_test_engine_reorg`. Both source paths are exercised:

```sh
# fixtures from a preview release, sources from the PR branch (no checkout needed)
./hive --sim ethereum/eels/consume-reorg \
  --sim.buildarg repo=https://github.com/CPerezz/execution-specs \
  --sim.buildarg branch=reorg-suite-pr \
  --sim.buildarg fixtures=https://github.com/CPerezz/execution-specs/releases/download/reorg-preview%40v1/fixtures_reorg.tar.gz \
  --client go-ethereum,reth,nethermind,besu,erigon

# or fill locally and stage them
simulators/ethereum/eels/consume-reorg/stage.sh ../execution-specs
./hive --sim ethereum/eels/consume-reorg \
  --sim.buildarg repo=https://github.com/CPerezz/execution-specs \
  --sim.buildarg branch=reorg-suite-pr \
  --sim.buildarg fixtures=/fixtures --client go-ethereum
```

Once the format is upstream, `repo` defaults to `ethereum/execution-specs` and `stage.sh` can go away.

Depends on ethereum/execution-specs#3556. Findings from running the suite across five clients are in the comments below.


CPerezz:
**Atomicity of `forkchoiceUpdated` when it returns `-38002`**

One behavioural difference we found while running the new reorg fixtures across clients, recorded here because it's a spec `MUST` that the suite currently tolerates.

**Case.** Canonical `a1 ← a2 ← a3` (head = `a3`); a sibling branch `b1 ← b2` delivered by `newPayload` only (both `VALID`). Then:

```
engine_forkchoiceUpdated(head = b2, safe = a2, finalized = a1)
```

`a2` and `a1` are on the `a` chain, not on `b2`'s chain. The [Engine API spec for `engine_forkchoiceUpdated`](https://github.com/ethereum/execution-apis/blob/main/src/engine/paris.md#engine_forkchoiceupdatedv1) requires an error here:

> Client software **MUST** return `-38002: Invalid forkchoice state` error if the payload referenced by `forkchoiceState.headBlockHash` is `VALID` and a payload referenced by either `forkchoiceState.finalizedBlockHash` or `forkchoiceState.safeBlockHash` does not belong to the chain defined by `forkchoiceState.headBlockHash`.

All five clients do return it. The question is what the node's state is afterwards, and the same spec section says:

> All updates to the forkchoice state resulting from this call **MUST** be made atomically.

i.e. a rejected call must change nothing.


| client | after the `-38002` |
|---|---|
| go-ethereum `v1.17.6-unstable-02872e9e` | **`latest` = `b2`, heights 1–2 = `b1`/`b2`** — reorg applied |
| reth `2.5.2` | **`latest` = `b2`, heights 1–2 = `b1`/`b2`** — reorg applied |
| nethermind `1.40.0-unstable` | unchanged |
| besu `v26.9-develop` | unchanged |
| erigon `3.7.0-dev` | unchanged |

Seems that Geth and Reth switch the head first and validates after.

Atomic" here just means all-or-nothing as seen from outside: after the call either all three of head/safe/finalized moved, or none did. When that's violated, the response says the call was rejected while the node has actually reorged, so for one slot `eth_getBlockByNumber("latest")`, `eth_call` at `latest` and `newHeads` serve a chain no consensus client accepted.

The suite currently treats the forkchoice state as unspecified after `-38002`, so all clients pass. We should likely make it more strict.

CPerezz:
Next finding: `engine_newPayload` returns a JSON-RPC error instead of a payload status for a side chain rooted at genesis (erigon)

Second cross-client difference from the new reorg fixtures. This one is the single biggest failure-generation one(9 of erigon's 15 failing fixtures).

**Case.** Canonical `a1 ← a2 ← a3`, each delivered with `newPayload` + `forkchoiceUpdated` (head = `a3`). Then a competing chain rooted at **genesis**, delivered with `newPayload` only:

```
newPayload(b1)   parent = genesis     → VALID
newPayload(b2)   parent = b1          → VALID
newPayload(b3)   parent = b2          → -32000  validateAndStorePayload: append with gap
                                                blockNum=2, but current height=3
                                        stack: [txnum.go:255 accessors_chain.go:802 stageloop.go:297
                                                stageloop.go:351 executor.go:343 fork_validator.go:304
                                                fork_validator.go:266 exec_module.go:632 ...]
```

geth, reth, nethermind and besu all answer `VALID` for the same `b3` bytes.

The [`engine_newPayload` spec](https://github.com/ethereum/execution-apis/blob/main/src/engine/paris.md#engine_newpayloadv1) reserves the error object for failures unrelated to normal processing - a well-formed payload on a known side chain should get `VALID`, `ACCEPTED` or `SYNCING`:

> If any of the above fails due to errors unrelated to the normal processing flow of the method, client software **MUST** respond with an error object.

**Boundary** (11-variant probe, erigon `3.7.0-dev-ab8e9fde` vs geth `v1.17.6-unstable-02872e9e`; every payload asserted only to return *some* status):

| canonical height | fork point | side length | erigon | geth |
|---|---|---|---|---|
| 3 | genesis | 1 | pass | pass |
| 3 | genesis | 2 | pass | pass |
| 3 | genesis | 3 | **fails at `b3`** | pass |
| 3 | genesis | 4 | **fails at `b3`** | pass |
| 3 | `a1` | 3 | pass | pass |
| 6 | genesis | 3 | **fails at `b3`** | pass |
| 6 | genesis | 6 | **fails at `b3`** | pass |
| 6 | `a1` / `a2` / `a3` / `a5` | 3 | pass | pass |

So: rooted at genesis **and** >= 3 blocks → fails on the third payload, independent of canonical height and of how many more side blocks follow. Any fork point >= 1 is fine at any length.

CPerezz:
One more: `latestValidHash` becomes `0x00..0` when the same invalid block is mentioned again (besu)

This issue compresses all the 16 failures Besu has.

The [payload validation rules](https://github.com/ethereum/execution-apis/blob/main/src/engine/paris.md#payload-validation) allow exactly three things in it:

> - The block hash of the ancestor of the invalid payload satisfying the following two conditions: it is fully validated and deemed `VALID`; any other ancestor of the invalid payload with a higher `blockNumber` is `INVALID`
> - `0x0000000000000000000000000000000000000000000000000000000000000000` if the above conditions are satisfied by a PoW block.
> - `null` if client software cannot determine the ancestor of the invalid payload satisfying the above conditions.

So post-merge a hash means "the fork point is here", `0x00..0` means "the last good ancestor is a pre-merge PoW block", and `null` means "I cannot tell".

--- 

Now the issue, is that a canonical `a1 <- a2 <- a3` (head = `a3`); `bad` is a child of `a3` with a corrupted state root. The same payload bytes are then sent twice, and the forkchoice is pointed at it:

```
newPayload(bad)             → INVALID, latestValidHash = a3        (correct)
newPayload(bad)  [same]     → INVALID, latestValidHash = 0x00..0
                              validationError = "Block already present in bad block manager."
forkchoiceUpdated(head=bad) → INVALID, latestValidHash = 0x00..0
                              validationError = "<hash> is an invalid block"
```

| client | first `newPayload(bad)` | repeat `newPayload(bad)` | `forkchoiceUpdated(head=bad)` |
|---|---|---|---|
| besu `v26.9-develop-b330564` | `a3` | **`0x00..0`** | **`0x00..0`** |
| go-ethereum `v1.17.6-unstable-02872e9e` | `a3` | `a3` | `a3` |
| reth `2.5.2` | `a3` | `a3` | `a3` |
| nethermind `1.40.0-unstable` | `a3` | `a3` | `a3` |
| erigon `3.7.0-dev-ab8e9fde` | `a3` | `a3` | `a3` |

The genesis here is post-merge, so there is no PoW block anywhere in the chain for `0x00..0` to refer to.

CPerezz:
Another one: A forkchoice update that does not move the head discards `safeBlockHash` and `finalizedBlockHash` (erigon)

Canonical chain `a1 ... a6`, each delivered with `newPayload` + `forkchoiceUpdated(head=aN, safe=0x0, finalized=0x0)`, so both markers start unset and the head is `a6`. Then one call that only advances the markers:
```
forkchoiceUpdated(head = a6, safe = a4, finalized = a2)   → VALID
eth_getBlockByNumber("safe")       → -39001 block "safe" not available (head block: 6)
eth_getBlockByNumber("finalized")  → -39001
```

The same call persists both markers when it also advances the head.

| case | call under test | geth | reth | nethermind | besu | erigon |
|---|---|---|---|---|---|---|
| head advances | `newPayload(a6)`, `FCU(head=a6, safe=a4, fin=a2)` | pass | pass | pass | pass | pass |
| same head, both markers | head already `a6`, `FCU(head=a6, safe=a4, fin=a2)` | pass | pass | pass | pass | **`safe` null** |
| same head, safe only | `FCU(head=a6, safe=a4, fin=0x0)` | pass | pass | pass | pass | **`safe` null** |
| same head, finalized only | `FCU(head=a6, safe=0x0, fin=a2)` | pass | pass | pass | pass | **`finalized` null** |
| rewind above finalized | markers set while advancing, then `FCU(head=a4, safe=a4, fin=a2)` | pass | pass | pass | pass | pass |

So the trigger is exactly "this call does not change the head"; both markers are affected, and the rewind path (which does change the head) is fine.
The trigger is realistic since a  consensus client re-sends the current head with advanced `safe`/`finalized` after a missed slot - and until the next head change `eth_getBlockByNumber("safe")` / `("finalized")` error out or lag, which affects anything reading those tags. Nothing diverges about the canonical chain itself tho.

CPerezz:
Final one: Rewinding the head to a canonical ancestor: one client bug and one spec conflict

Canonical `a1 ... a8` (head `a8`), then a single `forkchoiceUpdated(head = a5, safe = a5, finalized = X)`, varying `X`, then checking whether the head actually moved.

The relevant rules in [`engine_forkchoiceUpdated`](https://github.com/ethereum/execution-apis/blob/main/src/engine/paris.md#engine_forkchoiceupdatedv1):

> Client software **MAY** skip an update of the forkchoice state and **MUST NOT** begin a payload build process if there is a known `finalizedBlockHash` and `forkchoiceState.headBlockHash` references a `VALID` ancestor of the latest known finalized block [...] client software **MUST** return `{payloadStatus: {status: VALID, latestValidHash: forkchoiceState.headBlockHash, validationError: null}, payloadId: null}`.

> Client software **MUST** return `-38002: Invalid forkchoice state` error if the payload referenced by `forkchoiceState.headBlockHash` is `VALID` and a payload referenced by either `forkchoiceState.finalizedBlockHash` or `forkchoiceState.safeBlockHash` does not belong to the chain defined by `forkchoiceState.headBlockHash`.

**`applied` = head moved to `a5`; `noop` = `VALID` returned with the head left at `a8`.

| finalized at the time | does the skip clause apply? | geth `v1.17.6-unstable` | reth `2.5.2` | nethermind `1.40.0-unstable` | besu `v26.9-develop` | erigon `3.7.0-dev` |
|---|---|---|---|---|---|---|
| `0x0` (none known) | **no** - the clause requires a known finalized hash | applied | **noop** | applied | applied | applied |
| `a3` (head above it) | **no** - head is not an ancestor of finalized | applied | **noop** | applied | applied | applied |
| `a5` (head *is* finalized) | unclear - "ancestor of" vs "is" | applied | **noop** | applied | applied | applied |
| `a6` (head below it) | yes | noop | refused `-38006` | `-38002` | `-38002` | applied |

1. reth: the skip is unconditional. 
reth returns `VALID` with an unchanged head for *every* rewind to a canonical ancestor - including the first row, where no finalized block is known at all and no reading of the clause permits a skip. 

2. All clients vs. spec: the last row is a spec conflict, not a client bug. 
If the head is an ancestor of finalized, then finalized is a *descendant* of the head, so it does not belong to the chain defined by the head - the skip clause says `VALID`, the consistency rule says `-38002`, and nothing states which is evaluated first. 
The result? five clients, four distinct answers, none clearly wrong. geth takes the skip, nethermind and besu take the consistency rule, reth reports a depth limit, erigon performs the rewind (which neither rule suggests). 

We should probably file an issue in execution-apis to tackle this and resolve it. Then update the tooling to be strict.

PR #1613, opened 2026-09-17
simulators/ethereum/rpc-compat: add support for test validation scripts

fjl:
This is the counterpart to https://github.com/ethereum/execution-apis/pull/893


MysticRyuujin:
One question on the dependency.

The Dockerfile clones execution-apis at $branch for the fixtures, and go.mod pins execution-apis/tools at a fixed commit for the loader. The next time the .io format changes, the fixtures will update on rebuild but the loader will not until someone bumps the pin.

Maybe we should update the Dockerfile to use a replace?

```Dockerfile
  RUN go mod edit -replace github.com/ethereum/execution-apis/tools=/execution-apis/tools && go build -v .
```

That would also mean the pin does not need to chase the final [execution-apis/#893](https://github.com/ethereum/execution-apis/pull/893) commit after it squash-merges.

MysticRyuujin:
This no longer builds against the current execution-apis head, RunScript now takes a ScriptConfig:

```go
main.go:186:24: cannot use t (variable of type *hivesim.T) as iofile.ScriptConfig value in argument to test.RunScript
```

It also needs OpenRPCSchema set so the openrpc global exists in hive, otherwise scripts that pass speccheck throw a ReferenceError.

I assume still WIP

PR #1614, opened 2026-09-22
Add pbt-artifacts simulator for EIP-8347 conformance

CPerezz:
Conformance for the EIP-8347 offline migration artifacts: the preimage file and the PBT snapshot.

Until now each client tested its own artifact handling in its own tree, on branches that are not merged anywhere. There was no shared answer to the question "does this file mean the same thing to everyone". This simulator is that answer: one fixture set, one verdict per client, and a capability matrix showing who implements what.

Clients are driven through a per client shim that the simulator uploads into the container at `/hive-bin/pbt-artifacts.sh`, so no client integration file changes and adding a client is a single script. 
R

A client that lacks a feature reports it as unsupported instead of failing case by case, which is what turns a run into a matrix rather than a wall of red. 

Measured with the shipped `clients.yaml`, 52 cases (13 preimage, 38 snapshot, 1 unspecified):

| client | anchor state | consume preimages | consume snapshot | produce preimages | produce snapshot |
|---|---|---|---|---|---|
| go-ethereum | ok | 13/13, all attributed | 38/38, all attributed | byte identical | byte identical |
| nethermind | ok | 11/13 | 37/38 | unsupported | unsupported |
| erigon | ok | unsupported | unsupported | byte identical | unsupported |
| besu | ok | unsupported | unsupported | unsupported | unsupported |

Producer agreement: erigon and geth emit byte identical preimage files; the snapshot has one producer, so that agreement stays inconclusive until a second client can write one.

Nethermind's three misses are not rejections. On a truncated preimage file, a truncated snapshot or a trailing byte, its readers let an `EndOfStreamException` escape from `Stream.ReadExactly` and the node dies. The outcome for a user is the same, but an unhandled exception is not conformance.

Fixtures cover one embedding rule per account: code sizes around the 31 byte chunk, PUSH32 and PUSH2 straddling a chunk, PUSHDATA running past the end of the code, a zero chunk that must be absent and a trailing one, code spanning two code groups, shared bytecode, code that merely starts with `0xef0100`, delegations to a shared target, a distinct one, a missing account, an EOA and another delegation, storage either side of the header split, an account with only overflow storage, a zero balance contract, maximum nonce and balance. 

Depends on CPerezz/go-ethereum#41, which the geth pin names directly.


CPerezz:
The preimage file is derived from the allocation rather than taken from a producer, and the snapshot's leaf set is checked against an independent derivation of the embedding rules before anything is written, so the canonical bytes do not depend on any client being measured against them. 

Anyways, the next task will be to include the generations in the testing to make sure we don't only consume correctly.

